PR #103, opened 2026-02-26
feat: isolate sync branches across linked worktrees

jlevy:
Moves commit 1901a43 off main after restoring main to 1ad0234.

This keeps the change reviewable in a PR instead of directly on main.


jlevy:
Subsumed by https://github.com/jlevy/tbd/pull/121

PR #116, opened 2026-04-30
fix: handle invalid issue files safely

### LINKED ISSUE #115: tbd list/ready/stats crash silently when any issue file fails Zod validation; write path also lets invalid issues be created
## Summary

Three related defects combine into a high-impact failure mode where a single bad issue file makes most tbd read commands return zero results, with a confusing crash:

1. **Write path doesn't validate against `IssueSchema`** — `tbd create` will happily write a title that violates `z.string().min(1).max(500)` (and presumably any other constraint), poisoning the data store.
2. **Read path crashes ungracefully on parse failures** — `listIssues()` catches the parse error but then trips a Node `util.inspect` regression while logging it, throwing synchronously and aborting the command.
3. **Error formatting is opaque** — even when it doesn't crash, the warning passes a raw `ZodError` object to `console.warn`, which is unreadable and (per #2) actively dangerous.

Net effect: `tbd list`, `tbd ready`, `tbd stats`, `tbd search`, `tbd stale`, `tbd label list`, and `tbd status` all silently report 0 issues even though thousands of valid ones exist on disk.

## Symptoms

In a real repo with ~3,300 issues, after a planning agent created issues with very long titles:

```
$ tbd status
...
STATISTICS
  Ready:       0
  In progress: 0
  Blocked:     0
  Open:        0
  Total:       0 (3297 on remote - run 'tbd sync')
...
⚠ Clone status - Config has prefix 'ar' but no issues
    This suggests issues may have been created but not synced,
    or were lost due to sync issues on another machine.
```

The "0 / 3297 on remote" message is misleading — the issues are on disk, in sync with remote, just unreadable.

```
$ tbd --debug list
Error: Cannot read properties of undefined (reading 'value')

Stack trace:
TypeError: Cannot read properties of undefined (reading 'value')
    at formatProperty (node:internal/util/inspect:2280:12)
    at formatRaw (node:internal/util/inspect:1176:9)
    at formatValue (node:internal/util/inspect:932:10)
    at inspect (node:internal/util/inspect:409:10)
    at formatWithOptionsInternal (node:internal/util/inspect:2645:40)
    at formatWithOptions (node:internal/util/inspect:2507:10)
    at console.value (node:internal/console/constructor:377:14)
    at console.warn (node:internal/console/constructor:437:61)
    at listIssues (.../get-tbd/dist/bin.mjs:99778:13)
```

`tbd doctor` reports everything healthy because it checks file structure, not parseability.

### (1) Write path accepts invalid titles

`packages/tbd/src/cli/commands/create.ts` constructs an `Issue` object and calls `writeIssue()` without ever running it through `IssueSchema.parse()`. `writeIssue()` (storage.ts:38) just calls `serializeIssue()` and writes the bytes. So `tbd create "<2000 char title>"` succeeds and writes an unreadable file.

This bit us when an AI agent put a multi-paragraph description into the title field — five issues with titles between 658 and 1089 chars made it onto the `tbd-sync` branch and got pushed. They now break every clone of the repo.

### (2) Read path crashes synchronously on parse failure

`packages/tbd/src/file/storage.ts:88-91`:

```js
} catch (error) {
  // Skip invalid files with a warning
  console.warn(`Skipping invalid issue file: ${file}`, error);
}
```

The intent is graceful — log the bad file, keep going — but `console.warn(label, zodError)` throws on **Node v24.13.0 + zod 3.25.76**. The throw is synchronous and uncaught, so it propagates out of `listIssues` and aborts whatever command called it. Result: instead of "skipped 5 of 3313 issues", you get "0 issues" and a cryptic stack trace (only with `--debug`).

### (3) The Node × Zod inspect crash

Independent of tbd, this is reproducible in 6 lines:

```js
// Node v24.13.0, zod 3.25.76
const { z } = require('zod');
try {
  z.object({ id: z.string().regex(/^x-/) }).parse({ id: 'wrong' });
} catch (e) {
  console.warn('msg:', e);
}
// → TypeError: Cannot read properties of undefined (reading 'value')
//     at formatProperty (node:internal/util/inspect:2280:12)
```

Worth filing upstream too, but tbd should not depend on `console.warn` being able to inspect arbitrary error shapes. Even without this Node bug, dumping the raw `ZodError` is unfriendly.

## Real-world reproduction (in a tbd repo)

```bash
# 1. Create an issue with an oversize title — succeeds, no warning
LONG=$(python3 -c "print('x'*600)")
tbd create "$LONG" --type=task

# 2. Try to list issues — crashes
tbd list
# Error: Cannot read properties of undefined (reading 'value')

# 3. Status reports 0 issues despite the file being on disk
tbd status
#   Total: 0 ...

# 4. Doctor reports healthy
tbd doctor
#   ✓ Issue validity
```

## Why 0.1.26 doesn't fix it

`packages/tbd/src/file/storage.ts` and `packages/tbd/src/file/parser.ts` are byte-identical between v0.1.25 and v0.1.26 — `git log v0.1.25..v0.1.26 -- packages/tbd/src/file/{storage,parser}.ts` shows no commits.

### Fix 1 — Validate on write (prevention)

Make it impossible to persist an invalid issue. Either:

a. **In `writeIssue()`** (storage.ts:38) — `IssueSchema.parse(issue)` before serializing. This is the strongest guarantee because every write path goes through it. Cost: turns silent corruption into a clear early error at every callsite.

b. **In `tbd create`** (create.ts) — validate `title` (and any other user-controlled fields) against `IssueSchema.shape.title` before constructing the issue, with a friendly error like *"Title is too long (1089 chars, max 500). Move detail into the description body."*

Doing both is reasonable: (b) gives a clean user-facing error, (a) is the backstop for `update`, importers, and merge code paths.

### Fix 2 — Graceful read (don't kill the command)

storage.ts:90 — never pass the raw error object to `console.warn`. Format it to a string first:

```js
} catch (error) {
  const msg = error instanceof z.ZodError
    ? error.issues.map(i => `${i.path.join('.') || '<root>'}: ${i.message}`).join('; ')
    : error instanceof Error ? error.message
    : String(error);
  console.warn(`Skipping invalid issue file: ${file}: ${msg}`);
}
```

This makes the parse-skip actually graceful (which is what the comment says it should be) and removes the dependency on Node's ability to inspect arbitrary error shapes.

### Fix 3 — Surface skipped files clearly

The current behavior — print warnings to stderr per file, then continue — is fine as far as it goes, but most CLIs swallow stderr in practice (especially when piping). Consider:

- Have `listIssues` return both `issues` and `skipped: { file, reason }[]`, and let commands surface a one-line summary like `(5 issue files skipped — run 'tbd doctor' for details)`.
- Add an explicit health check in `tbd doctor` that re-parses every issue file and reports invalid ones with their schema errors. Today doctor reports `✓ Issue validity` even when nothing is parseable.

### Fix 4 — Self-heal command (optional)

`tbd doctor --fix` could offer to truncate oversize titles (move overflow into description) for known-recoverable cases, since the user otherwise has to hand-edit YAML in the worktree branch.

## Versions

- get-tbd: 0.1.25 and 0.1.26 (both affected — code is identical)
- Node: v24.13.0 (likely all 24.x)
- zod: 3.25.76 (the bundled version in get-tbd)

## Related

The Node `util.inspect` crash on ZodError is plausibly a Node 24 regression and/or a zod 3.25 inspect-customization issue — worth a separate upstream report, but tbd's fixes don't depend on it.

jlevy:
## Summary

- validate issues in `writeIssue` and reject overlong CLI titles before writing
- format Zod/read errors as strings so invalid issue files are skipped without crashing
- report parse-invalid issue files in `tbd doctor`
- add regression coverage for create, storage, list, and doctor behavior

Fixes #115

## Validation

- `pnpm run ci`
- pre-push hook: `pnpm test`

### DIFF
--- a/.changeset/valid-issue-files.md
+++ b/.changeset/valid-issue-files.md
@@ -1,0 +1,5 @@
+---
+"get-tbd": patch
+---
+Reject invalid issue titles before writing issue files, skip parse-invalid issue files
+without crashing, and report invalid issue files in `tbd doctor`.
--- a/packages/tbd/src/cli/commands/create.ts
+++ b/packages/tbd/src/cli/commands/create.ts
@@ -54,6 +54,7 @@ class CreateHandler extends BaseCommand {
         ? undefined
         : validateIssueTitle(title, {
             emptyMessage: 'Title is required. Use: tbd create "Issue title"',
+            rejectBlank: true,
           });
 
     // Parse and validate options
--- a/packages/tbd/src/cli/commands/create.ts
+++ b/packages/tbd/src/cli/commands/create.ts
@@ -25,6 +25,7 @@ import { resolveDataSyncDir } from '../../lib/paths.js';
 import { now } from '../../utils/time-utils.js';
 import { readConfig } from '../../file/config.js';
 import { resolveAndValidatePath, getPathErrorMessage } from '../../lib/project-paths.js';
+import { validateIssueTitle } from '../lib/issue-input-validation.js';
 
 interface CreateOptions {
   fromFile?: string;
@@ -48,6 +49,12 @@ class CreateHandler extends BaseCommand {
     if (!title && !options.fromFile) {
       throw new ValidationError('Title is required. Use: tbd create "Issue title"');
     }
+    const validatedTitle =
+      title === undefined
+        ? undefined
+        : validateIssueTitle(title, {
+            emptyMessage: 'Title is required. Use: tbd create "Issue title"',
+          });
 
     // Parse and validate options
     const kind = this.parseKind(options.type ?? 'task');
@@ -76,7 +83,13 @@ class CreateHandler extends BaseCommand {
     }
 
     if (
-      this.checkDryRun('Would create issue', { title, kind, priority, spec: specPath, ...options })
+      this.checkDryRun('Would create issue', {
+        title: validatedTitle,
+        kind,
+        priority,
+        spec: specPath,
+        ...options,
+      })
     ) {
       return;
     }
@@ -122,7 +135,7 @@ class CreateHandler extends BaseCommand {
         type: 'is',
         id,
         version: 1,
-        title: title!,
+        title: validatedTitle!,
         kind,
         status: 'open',
         priority,
@@ -163,8 +176,8 @@ class CreateHandler extends BaseCommand {
 
     // Output with display ID (prefix + short ID)
     const displayId = `${prefix!}-${shortId!}`;
-    this.output.data({ id: displayId, internalId: id, title }, () => {
-      this.output.success(`Created ${displayId}: ${title}`);
+    this.output.data({ id: displayId, internalId: id, title: validatedTitle }, () => {
+      this.output.success(`Created ${displayId}: ${validatedTitle}`);
     });
   }
 
--- a/packages/tbd/src/cli/commands/doctor.ts
+++ b/packages/tbd/src/cli/commands/doctor.ts
@@ -12,7 +12,7 @@ import { join } from 'node:path';
 
 import { BaseCommand } from '../lib/base-command.js';
 import { requireInit } from '../lib/errors.js';
-import { listIssues } from '../../file/storage.js';
+import { listIssues, type InvalidIssueFile } from '../../file/storage.js';
 import { readConfig } from '../../file/config.js';
 import type { Config, Issue, IssueStatusType } from '../../lib/types.js';
 import { resolveDataSyncDir, TBD_DIR, WORKTREE_DIR, DATA_SYNC_DIR } from '../../lib/paths.js';
@@ -59,6 +59,7 @@ class DoctorHandler extends BaseCommand {
   private cwd = '';
   private config: Config | null = null;
   private issues: Issue[] = [];
+  private invalidIssueFiles: InvalidIssueFile[] = [];
 
   async run(options: DoctorOptions): Promise<void> {
     const tbdRoot = await requireInit();
@@ -75,7 +76,11 @@ class DoctorHandler extends BaseCommand {
 
     // Load issues
     try {
-      this.issues = await listIssues(this.dataSyncDir);
+      this.invalidIssueFiles = [];
+      this.issues = await listIssues(this.dataSyncDir, {
+        warnOnInvalid: false,
+        onInvalidIssue: (invalidIssue) => this.invalidIssueFiles.push(invalidIssue),
+      });
     } catch {
       // May fail if no issues yet
     }
@@ -114,7 +119,7 @@ class DoctorHandler extends BaseCommand {
     healthChecks.push(await this.checkTempFiles(options.fix));
 
     // Check 8: Issue validity
-    healthChecks.push(this.checkIssueValidity(this.issues));
+    healthChecks.push(this.checkIssueValidity(this.issues, this.invalidIssueFiles));
 
     // Check 9: Worktree health (with fix support)
     // Run BEFORE ID mapping check — worktree repair and data migration can
@@ -130,7 +135,11 @@ class DoctorHandler extends BaseCommand {
     if (dataLocationResult.status === 'ok' && dataLocationResult.message?.includes('migrated')) {
       this.dataSyncDir = await resolveDataSyncDir(this.cwd);
       try {
-        this.issues = await listIssues(this.dataSyncDir);
+        this.invalidIssueFiles = [];
+        this.issues = await listIssues(this.dataSyncDir, {
+          warnOnInvalid: false,
+          onInvalidIssue: (invalidIssue) => this.invalidIssueFiles.push(invalidIssue),
+        });
       } catch {
         // Will be caught by other health checks
       }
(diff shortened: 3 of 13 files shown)


jlevy (commented):
## Senior Engineering Review

### Summary

This PR cleanly fixes the failure chain described in #115: `writeIssue` now validates against `IssueSchema` so invalid issues can't be written, `listIssues` reports skipped files via formatted strings (avoiding the `console.warn(label, ZodError)` Node `util.inspect` regression), and `tbd doctor` surfaces parse-invalid files as a real diagnostic instead of hiding them. Scope is right-sized for the bug and well-tested. Recommend merging with a couple of small follow-ups.

### Strengths

- **Defense in depth**: `validateIssueTitle` at the CLI layer gives users actionable errors ("501 chars, max 500. Move detail into the description body."), while `IssueSchema.parse` in `writeIssue` is the catch-all that prevents any caller (CLI, sync, import, attic, dep, label) from poisoning the data store. Both layers are needed; both are present.
- **`formatUnknownError` / `formatZodError`** are the right abstraction. They format errors *before* they hit `console.warn`, which is the actual root cause of the user-reported crash on Node v24.13 + Zod 3.25. Path-qualified output (e.g., `title: too big`) is much more useful than the prior `[object Object]`-style dump.
- **`listIssues` API**: `warnOnInvalid` defaulted to `true` keeps all 25+ existing callers behavior-compatible while letting `doctor` opt into structured reporting via `onInvalidIssue`. Backward-compat with no migration needed across the rest of the codebase.
- **Constants extraction**: `ISSUE_TITLE_MAX_LENGTH` / `ISSUE_BODY_MAX_LENGTH` / `IssueTitle` are now single-source-of-truth, and the test fixtures use them — no more drift risk between schema and validation messages.
- **Test coverage**: storage unit test confirms `writeIssue` rejects *and* doesn't leak a partial file ([packages/tbd/tests/storage.test.ts:149-158](packages/tbd/tests/storage.test.ts#L149-L158)). Integration test in `corrupted-data.test.ts` exercises the full read path through `tbd list` and `tbd doctor`, including the doctor health-check format and Windows-friendly path matching.
- **Doctor diagnostic**: invalid files are now reported as `error` status with file paths and reasons, where previously they crashed every read silently.

### Issues / Suggestions (all non-blocking)

1. **Whitespace-only title inconsistency between `create` and `update`** ([packages/tbd/src/cli/commands/create.ts:55-57](packages/tbd/src/cli/commands/create.ts#L55-L57) vs [packages/tbd/src/cli/commands/update.ts:233-236](packages/tbd/src/cli/commands/update.ts#L233-L236)).
   - `tbd update --title "   "` errors with "Title cannot be empty" because `update` passes `rejectBlank: true`.
   - `tbd create "   "` does not — it would create an issue with a 3-space title (passes `IssueTitle.min(1)`).
   - Pre-existing `create` didn't trim either, so this isn't a regression, but the new helper makes it trivial to fix. Suggest passing `rejectBlank: true` in `create` too for parity.

2. **No direct unit test for `validateIssueTitle`** ([packages/tbd/src/cli/lib/issue-input-validation.ts](packages/tbd/src/cli/lib/issue-input-validation.ts)). The coverage report shows it at 0% — the tryscript only exercises it via the compiled binary subprocess, which Vitest's source coverage can't see. The function has three distinct branches (empty, too-long, fall-through "Invalid title") and two flag combinations; a small unit test would lock the contract.

3. **`listInvalidIssueFiles` is exported but unused** ([packages/tbd/src/file/storage.ts:137](packages/tbd/src/file/storage.ts#L137)). `doctor` uses `listIssues(..., { onInvalidIssue })` directly. Either drop it (YAGNI) or migrate `doctor` to use it.

4. **Doctor diagnostic `path` field is approximate** ([packages/tbd/src/cli/commands/doctor.ts:656](packages/tbd/src/cli/commands/doctor.ts#L656)). Shown as `.tbd/issues`, but actual files live under `.tbd/data-sync-worktree/.tbd/data-sync/issues/`. Not new to this PR (other diagnostics use the same shorthand) but worth noting since users following the path won't find the file. Could pass `this.dataSyncDir`-relative path instead.

5. **`validateIssueTitle` returns `result.data` for an unchanged string**. Since `IssueTitle` doesn't transform, `result.data === title`. Returning it adds no value — could just `return title;`. Minor stylistic.

6. **Behavioral change worth documenting**: `IssueSchema.parse` in `writeIssue` runs in default `.strip()` mode, so any unknown frontmatter fields a caller might have attached to an `Issue` object will be silently dropped at write. In practice this matches `parseIssue` (which already strips on read), so it's not a regression for round-trips, but anyone hand-constructing an `Issue` with extension data should use the `extensions` namespace ([packages/tbd/src/lib/schemas.ts:76](packages/tbd/src/lib/schemas.ts#L76)).

### Documentation

- Changeset is accurate and concise. Marked `patch` — defensible since the CLI behavior change (rejecting too-long titles) is the bug fix described in #115; storage validation tightening is internal.
- No spec doc references the affected modules; nothing in `docs/project/specs/active/` looks out of sync.
- No CHANGELOG entry beyond the changeset, which is the project convention.

### Unresolved Comments

None — DeepSource shows grade A across all categories, no inline review comments outstanding.

### CI Status

All checks passing:
- Benchmark, Coverage & Lint, Test (macos/ubuntu/windows), DeepSource Secrets — all green.

### Recommendation

**Approve with optional follow-ups.** The fix is correct, scoped tightly to #115, and well-tested at both unit and integration levels. The suggestions above are polish, not blockers.

PR #117, opened 2026-05-07
process: Add docs config redesign plan spec

jlevy:
## Summary

Adds `docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md`, a planning spec for a redesign of the docs/config system (the area PR #87 attempted to address but never merged).

The spec:

- Reviews the current f03 system and the unmerged PR #87 f04 design, calling out the architectural smells (three coexisting layers — `sources` + `files` + `lookup_path` — that produced 12 bug-fix commits, repo sync wired only in config not pipeline, local-source type spec-only, doc types still a closed enum).
- Restates the 8 goals from the user request explicitly (G1–G8).
- Proposes 5 additional goals for review (G9 reproducibility, G10 provenance, G11 clean schema break, G12 atomic sync, G13 auth-ready).
- Outlines three candidate design approaches with pros/cons:
  - **A**: Finish PR #87 as-designed (smallest delta).
  - **B** (recommended): Clean f05 re-cut — one ordered source list, four types (`bundled`/`local`/`git`/`url`), no parallel `files` or `lookup_path`, doc types as data, override = priority, eject/diff/unfork commands for the roundtrip workflow.
  - **C**: Fully pluggable source-type providers (defer).
- Lists 9 open questions to resolve before any code is written.
- Sketches a two-phase implementation plan (schema + sync + migration; then override / promotion / roundtrip).

This is a planning spec at the design-options stage — no code changes.

## Test plan

- [ ] Spec read for clarity and completeness
- [ ] User confirms which of G9–G13 to keep
- [ ] User picks an approach (A/B/C) before follow-up implementation spec is written
- [ ] Open questions resolved before beads are created

https://claude.ai/code/session_01PhbYdWX7DUBpUBVuUesVuP

---
_Generated by [Claude Code](https://claude.ai/code/session_01PhbYdWX7DUBpUBVuUesVuP)_


jlevy:
Senior design review after reading the current PR, the existing f03/f04 context, and the related KDEX draft in `jlevy/kdex`.

## Summary

The direction is mostly right: replacing `files + lookup_path` with a manifest/lockfile/generated-map model is the right architectural move, and splitting `docref` from `docmap` is a good start toward extraction. I would not implement the remaining phases directly from this design yet, though. The biggest gaps are around resolver semantics for overrides, source-vs-bundle modularity, and explicit provenance for the add/remove/update/eject workflows.

My suggested high-level direction is to make this a three-layer system:

1. `docref`: pure address grammar and normalization. No fetching, no tbd policy.
2. `docgraph`: manifest normalization plus generated inventory/provenance graph. It contains all sources, bundles, documents, blobs, aliases, lock state, and override/fork edges. It never hides anything.
3. `docmap`: a resolved view over a `DocGraph` using a policy: source priority, type filters, alias rules, and shadowing behavior. tbd workflows (`source add/remove/update`, `eject`, `diff`, `upstream`, `unfork`) sit above this layer.

That distinction would make the design more reusable and gives tbd the workflow semantics it needs without forcing every future DocMap consumer to inherit tbd-specific override behavior.

## Strong parts

- The clean break from f03/f04 is correct. The old coexistence of `sources`, `files`, and `lookup_path` is exactly the kind of compatibility layer that tends to resurrect deprecated state.
- `sync` vs `update` follows the npm/uv mental model and is the right contract: `sync` installs the locked state; `update` intentionally moves it forward.
- `docref` as a very small parser with unknown schemes delegated to consumers is a good extraction boundary.
- KDEX's better ideas are visible here: a manifest, generated map, lockfile, offline resolution, status, whole-repo aggregate entries, and a thin CLI over a library. Keep leaning into that.

## High-impact issues to resolve

### 1. Override-by-priority currently contradicts the resolver implementation

The design says order is lookup order and first source wins for unqualified names (`packages/tbd/docs/design-docmap-format.md:139`, `packages/tbd/docs/design-docmap-format.md:463`, `docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:490`). That is also the core mechanism for local overrides (`docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:520`).

But the reference resolver and tests do the KDEX-style thing: basename collisions are ambiguous (`packages/tbd/src/docmap/resolve.ts:132`, `packages/tbd/tests/docmap-resolve.test.ts:102`). The spec itself also says multiple basename matches are ambiguous at `packages/tbd/docs/design-docmap-format.md:440`, so the doc contradicts itself.

This is not a small wording issue. If duplicate `<type>/<name>` entries are ambiguous, then the proposed shadcn-style override workflow does not work. If first match wins, then the current resolver and tests are wrong.

Recommended fix: make the raw graph and resolved view explicit.

- The `DocGraph` should retain every item, including shadowed items.
- The `DocMap` effective view should resolve by `(type, name)` plus source priority and return the winning item plus shadow metadata.
- Resolver APIs should accept context such as `{ type?: 'guideline', bundle?: 'coding', mode?: 'effective' | 'all' | 'strict' }`.
- Existing typed commands like `tbd guidelines typescript-rules` should pass a type constraint; otherwise `typescript` could collide across `guideline`, `shortcut`, and `reference`.
- Ambiguity should remain for cases priority cannot or should not resolve: duplicate aliases in the same effective scope, conflicting canonical keys, or explicit `mode: 'strict'` lookups.

That gives tbd override behavior without losing KDEX's useful “full map is lossless” property.

### 2. `bundle = source` is too tight for modularity

The current format defines bundle as attached to a source and says one source equals one bundle (`packages/tbd/docs/design-docmap-format.md:48`, `packages/tbd/docs/design-docmap-format.md:147`, `docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:218`). The plan even flags this as open (`docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:642`). I would resolve that open question the other way: split bundles from sources now.

Real use cases where a bundle should have multiple sources:

- An org docs bundle composed from a repo plus a few canonical web URLs.
- A `tbd` bundle with a small package-local core plus an external docs repo during migration.
- A product bundle where reference docs, examples, and code live in different repositories.
- Multiple single-file URL sources that should present as one user-visible bundle.

Suggested schema shape:

```yaml
docmap:
  schema: docmap/0.1
  bundles:
    - name: coding
      priority: 20
    - name: proj
      priority: 10
      local_root: ./docs/agent/
  sources:
    - id: coding-main
      bundle: coding
      docref: github:jlevy/coding-guidelines@main
      contents: [auto]
    - id: coding-readme
      bundle: coding
      docref: https://example.com/coding-style.md
      type: guideline
      as: coding-style
```

You can still keep the common CLI path simple (`tbd source add ...` auto-creates a bundle), but the format should not bake in 1:1 cardinality if we want DocMap reusable outside tbd.

### 3. Lockfile identity is under-specified

The lockfile examples and schema record `docref`, hash, materialization, and timestamp, but not a stable source id, bundle, or source-config fingerprint (`packages/tbd/docs/design-docmap-format.md:257`, `packages/tbd/src/docmap/schemas.ts:104`). That is fragile once there are multiple sources with the same docref, a source moves bundles, filters change, or `contents` remaps the same upstream tree into different landed docs.

Recommended lock entry fields:

- `source_id`: stable manifest source id.
- `docref`: normalized source address.
- `source_config_hash`: hash of the materialization-affecting source config (`docref`, `contents`, `glob`, `ignore`, `depth`, transforms).
- `revision`: resolved immutable upstream revision where applicable.
- `content_hash`: hash of the selected/materialized content, not just the repo checkout.
- `materialization`: fetch/cache strategy.

Then `sync`, `update`, `remove`, and orphan cleanup operate on source ids rather than trying to infer identity from `docref` or bundle name.

### 4. The override/fork relationship needs explicit provenance

The state model says tracked override vs pure local is computed by checking whether another bundle contributes the same `<type>/<name>` (`docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:671`). W8 says eject may record the original revision in frontmatter or a sidecar (`docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:926`), but this needs to be a first-class part of the workflow design.

Without explicit provenance, these cases get muddy:

- A pure local doc becomes a supposed override just because a later-added upstream bundle happens to have the same name.
- Removing an upstream bundle turns an override into pure local and loses the data needed for `diff` or `unfork`.
- Upstream renames or deletes a doc; the local file still exists but the relationship is no longer discoverable by name.
- `source upstream` needs to know the exact upstream source/path/revision to patch.

Recommendation: record an override edge, not just infer one. For example, in a sidecar or tbd overlay:

```yaml
overrides:
  - local_key: proj:guideline/typescript
    upstream_key: coding:guideline/typescript
    upstream_source_id: coding-main
    upstream_docref: github:jlevy/coding-guidelines@main//guidelines/typescript.md
    ejected_revision: <sha>
    ejected_content_hash: sha256:...
```

The effective DocMap can still compute shadowing by priority, but roundtrip workflows should use explicit override provenance.

### 5. The Zod schemas are not enforcing the clean-break contract

G11 says the f05 validator rejects unknown fields to prevent zombie schema fields (`docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md:131`). The new schemas use plain `z.object(...)`, which strips unknown keys by default in Zod (`packages/tbd/src/docmap/schemas.ts:63`, `packages/tbd/src/docmap/schemas.ts:78`). That means old fields can silently survive parse/load paths unless every caller has a separate unknown-field detector.

Also missing or under-specified cross-field validation:

- Remote sources can omit `bundle` even though the spec says bundle is required for remote sources (`packages/tbd/src/docmap/schemas.ts:65`).
- `contents[].type` and `source.type` are not checked against declared `doc_types`.
- `as` does not require `type`, and the spec is unclear whether `as` means “aggregate source name”, “single-file rename”, or KDEX-style `as: repo` (`packages/tbd/docs/design-docmap-format.md:163`).
- Bundle uniqueness, reserved names, duplicate source ids, and canonical-key collision checks are not represented at the schema/normalization boundary.

Use `.strict()` on the format schemas, then add a normalization/validation pass with `superRefine` or a separate `validateManifest()` for cross-object rules. If extension fields are needed for future consumers, make that explicit with an `extensions:` bag rather than accepting arbitrary top-level keys.

### 6. `as` should be made less overloaded before implementation

KDEX used `as: repo` for whole-repo aggregate treatment. This PR changes `as` into a string name, while also saying the aggregate key becomes the bundle name (`packages/tbd/docs/design-docmap-format.md:163`, `packages/tbd/docs/design-docmap-format.md:364`). That is confusing: if `as` is not the key, what does the value name?

I would split this into a discriminated mode:

```yaml
mode: files        # default: index files selected by contents/glob
mode: file         # one source becomes one doc; requires type/name
mode: repo         # one source becomes aggregate repo entry
```

Or keep KDEX's `as: repo` and use `contents: [{ path: README.md, type: reference, as: writing-overview }]` for single-file renames. Either is clearer than `as` meaning both “aggregate mode” and “rename”.

## Workflows

### Add source

The proposed preview-first add flow is sensible. I would make the dry-run inventory step the core operation:

1. Parse and normalize docref.
2. Choose or create a bundle.
3. Create a source id.
4. Dry-run fetch/list enough to show landed docs and conflicts.
5. Persist manifest only after preview confirmation.
6. Run `sync` and `build`, or print the exact next command if non-interactive mode disables side effects.

For CI/non-interactive use, support `--yes` and `--no-sync`.

### Remove source or bundle

Default-safe behavior is right: refuse removal if explicit override edges point at that source unless `--force-orphan` is passed. If bundles and sources split, expose both:

- `tbd source remove <source-id>` removes one fetch unit.
- `tbd bundle remove <bundle>` removes all sources in a bundle, with the same override safety checks.

Cache directory and lockfile entry deletion should be source-id based.

### Update source

Keep `sync` and `update` sharply separated. `sync` must reproduce the lockfile. `source update [source|bundle]` advances revisions and rewrites lock entries. Status should distinguish:

- missing cache
- cache hash mismatch
- locked and present
- upstream has newer revision
- local override diverges from current upstream
- explicit override is orphaned

### Eject/diff/upstream/unfork

The workflow is directionally good, but it needs explicit override provenance as above. Also, two-way diff is okay for v1 only if we record enough data to add three-way later without changing the sidecar format. Store eject-time revision/hash now even if v1 only displays local-vs-current-upstream.

## KDEX comparison

Useful KDEX ideas to borrow:

- A library API boundary: `load`, `sync`, `build`, `resolve`, `read`, `status` (`attic/kdex/docs/project/design/kdex-spec.md:198`). This PR has schemas and a resolver, but not yet the core API surface that future extraction would want.
- `ResolvedItem` includes both source address and local path (`attic/kdex/docs/project/design/kdex-spec.md:134`). DocMap entries should carry enough provenance to be useful without re-reading the manifest.
- Prompt-friendly map rendering with a budget (`attic/kdex/docs/project/design/kdex-spec.md:802`) is worth keeping in mind for tbd prime / agent context injection.
- Whole-repo aggregate sources with subpath reads are valuable and should remain in scope, but the mode should be explicit.

Do not copy KDEX's ambiguity semantics wholesale. KDEX is a neutral knowledge index; tbd specifically needs override priority and typed command resolution. That difference is why I think `DocGraph` plus policy-driven `DocMap` is the better reusable abstraction.

## Smaller code/package notes

- If `docref` and `docmap` are intended to be externally consumable soon, consider exporting subpaths like `get-tbd/docref` and `get-tbd/docmap` or documenting that they are internal until extraction. Right now they are not exposed by `packages/tbd/package.json`.
- The module-level `index.ts` barrel files are okay if these are future package roots, but the repository's TypeScript guideline discourages duplicate/internal `index.ts` barrels. If these remain internal to tbd for a while, consider more descriptive filenames or make the package-boundary intent explicit.
- The tests are useful spec mirrors, but they currently encode the resolver contradiction. Add tests for the winning override case before implementing source walking.

## Bottom line

I would keep the PR's general direction, but I would change the design before building Phase 1/2 around it:

1. Introduce `DocGraph` as the lossless, provenance-rich inventory.
2. Make `DocMap` an effective policy view over that graph.
3. Split bundles from sources and add stable source ids.
4. Strengthen lockfile identity and schema strictness.
5. Make override provenance explicit.
6. Resolve the priority-vs-ambiguity contradiction in the spec and tests.

Those changes should make add/remove/update/eject workflows much more predictable, and they also make future extraction into one or two reusable repos feel natural instead of retrofitted.

jlevy:
Thanks for the thorough design review — the framing of `DocGraph` as the lossless inventory and `DocMap` as a policy view is exactly the right architectural lens, and the call-out on the resolver/spec contradiction is correct. Pushed `87a3d9a` with the small/clear fixes; the larger architectural items are now Q15–Q19 in the plan-spec for joint design review (linked below) so we can settle them together before Phase 2.

## Fixes applied in this commit

**1. Resolver semantics now match the spec's stated intent** (matches your point 1 partially)

`resolveLookupKey` previously threw `Ambiguous` whenever a basename matched in two bundles, contradicting the override-by-priority design. It now distinguishes:

- **Same `(type, name)` across bundles** → priority wins (first source listed beats later ones). This is the local-override mechanism the design promised.
- **Same basename across different `(type, name)` buckets** → genuine ambiguity (priority alone can't disambiguate a typed lookup).

A new helper `resolveByPriorityOrAmbiguous()` encapsulates the rule. Same priority semantics now apply to alias matches. Documents array is documented as required to be in source priority order.

Tests updated:
- "throws Ambiguous on a basename that matches multiple bundles" → split into a "priority wins when same `(type, name)`" test and a separate "throws Ambiguous when basename matches across different types" test (with a fresh `coding:shortcut/typescript` fixture entry to make the cross-type case testable).
- Same split for alias resolution.
- "resolves a bundle-scoped basename" → split into the unique-within-bundle case (clean resolution) and the cross-type case (ambiguous).

**2. Spec §4.4 / §4.5 rewritten to remove the contradiction**

The new §4.4 "Collisions and priority" spells out four explicit cases (canonical-key collisions, same `(type, name)` across bundles, same basename across types, status reporting) and forward-references Q15 for the `mode=` parameter. Old §4.5 folded into §4.4. Stale `§4.5` reference in `resolve.ts` updated.

**3. Zod schemas now `.strict()` everywhere** (your point 5)

Every object schema is `.strict()`. A stray `lookup_path:` from f04 now throws `unrecognized` at parse time rather than being silently dropped. New tests exercise this for top-level and per-source cases.

**4. Cross-field validation: bundle required for non-local docrefs** (also your point 5)

`SourceSchema.superRefine` checks: if `docref.kind !== 'path'` (i.e. `https:`/`github:`/`gitlab:`/`git:`/etc.), `bundle` is required. Local docrefs may omit it (consumer applies the `local` default). New tests for both sides.

## Deferred to joint review as Q15–Q19 in the plan-spec

The bigger structural items live in [`plan-2026-05-07-docs-config-redesign.md` § "Architectural questions surfaced in design review"](docs/project/specs/active/plan-2026-05-07-docs-config-redesign.md). Each lists 2–4 design options with tradeoffs; nothing tentative yet, just the option space:

- **Q15. Resolver semantics** (your point 1, full version) — keep current shape vs add `mode` parameter (`'effective' | 'all' | 'strict'`) vs full `DocGraph` + `DocMap` policy-view split.
- **Q16. Bundle ↔ source cardinality** (your point 2) — keep 1:1 vs split bundles from sources with stable source ids vs optional grouping. Lists the four real use cases you cited.
- **Q17. Lockfile identity** (your point 3) — docref-only (current) vs add `source_id` only vs full `source_config_hash` + `content_hash`. Q17 depends on Q16's outcome (source ids).
- **Q18. Override provenance** (your point 4) — computed-by-name (current) vs frontmatter pointer in the override doc vs sidecar edge vs tbd-internal `.tbd/` overlay file. Q18 depends on Q16 if we want stable source ids in the edge.
- **Q19. The overloaded `as` field** (your point 6) — keep + document vs split into a `mode:` discriminator on sources (`files`/`file`/`repo`) with `as` only on contents rules vs KDEX-aligned literals.

The spec notes that Q15–Q19 are linked: Q17/Q18 substantially benefit from whatever Q16 lands on; Q15's full DocGraph/DocMap shape is the natural home for Q18's override edges.

## Smaller notes

- **Module barrels / package boundaries.** Agreed they're not yet exposed via `package.json` exports. Will keep them internal-but-cleanly-separable until the architectural questions settle (Q15 in particular informs whether docmap stays one module or splits into DocGraph + DocMap layers). Once we pick a shape, exposing them as `get-tbd/docref` and `get-tbd/docmap` (or as their own packages) is a localized change.
- **Tests-as-spec-mirror discipline.** The new tests for priority-wins, cross-type ambiguity, schema strictness, and the bundle/docref cross-field rule are intentionally written to fail if the spec changes — they're the synchrony enforcement.

Let me know which of Q15–Q19 you want to dig into first; the resolver shape (Q15) and the bundle/source cardinality (Q16) seem like the foundational ones since Q17 and Q18 depend on them.

---

jlevy:
Closing without merging.

The docs config redesign was not merged as a whole. Its first slice shipped as forkable docs (f05) in #169; see `docs/project/specs/done/plan-2026-06-11-forkable-docs.md`, which maps this spec's concepts onto what shipped and explains what it deliberately left out. The docref format was adopted (`packages/tbd/src/docref/`). The rest of this spec (external bundles, lockfiles, priority overrides, a separate docs repo, and open questions Q15-Q19) stays future work. Main has moved to f08 and this branch now conflicts with it.

The branch is kept (head f113cef9). The spec file exists only there, and epic tbd-up8l and its beads still reference it; that follow-up is recorded on tbd-up8l. The doctor summary fix in 0493fa90 never reached main and is noted there too.

PR #118, opened 2026-05-11
fix: validate issues on write and gracefully skip on read (#115)

### LINKED ISSUE #115: tbd list/ready/stats crash silently when any issue file fails Zod validation; write path also lets invalid issues be created

jlevy:
## Summary

Fixes #115. A single bad issue file (e.g. one with an oversize title) was making `tbd list`, `ready`, `stats`, `status`, `search`, and `stale` silently report zero issues. This patch closes the loop on all three defects from the issue: validate before writing, format errors safely on read, and surface skipped files via `tbd doctor`.

## Changes

- **`writeIssue` validates against `IssueSchema`** (`packages/tbd/src/file/storage.ts`) — runs `IssueSchema.parse(issue)` before serializing, so corrupt issues can't reach disk via `create`, `update`, `import`, or `sync`. Throws `ZodError` early at every callsite instead of silently corrupting the data store.
- **`tbd create` validates titles up-front** (`packages/tbd/src/cli/commands/create.ts`) — checks `IssueSchema.shape.title` before constructing the issue and throws a friendly `Invalid title (1089 chars): String must contain at most 500 character(s). Move detail into the description (-d / -f).` instead of a raw schema error.
- **`listIssues` no longer crashes on parse failure** (`packages/tbd/src/file/storage.ts`) — added `formatIssueParseError()` that converts `ZodError`/`Error`/anything to a plain string before logging. This eliminates the Node v24 `util.inspect` crash on `console.warn(label, zodError)` that was aborting whole commands. Closes with a one-line summary: `(N issue file(s) skipped — run 'tbd doctor' to see details)`.
- **New `listIssuesDetailed`** returning `{ issues, skipped: { file, reason }[] }` so callers (`tbd doctor`) can surface the skip list cleanly without piping warnings through stderr.
- **`tbd doctor` Issue validity check** (`packages/tbd/src/cli/commands/doctor.ts`) — replaced a no-op check (which inspected already-parsed `Issue[]` and so could never see schema violations) with one that re-parses every file from disk via `listIssuesDetailed` and reports each invalid file with its schema error. Doctor's own `listIssues` calls switched to `listIssuesDetailed` to avoid double-warning.
- **`tbd doctor` Temp files check** — was reporting and scanning `.tbd/issues` (which doesn't exist in current installs) instead of the real worktree path; now reports `.tbd/data-sync/issues` and also catches `atomically`'s leftover `<name>.md.tmp-NNNN` intermediates, not just plain `.tmp`.
- **Tests**: added 11 cases to `packages/tbd/tests/storage.test.ts` covering write-time rejection of oversize/empty/max-length titles, the `listIssues` skip path, `listIssuesDetailed` return shape, the contract that warnings are always strings (never raw error objects), and `formatIssueParseError` over `ZodError`/`Error`/unknown values.

## Test Plan

- [x] Unit tests pass — `pnpm --filter get-tbd test` → 1054 passed (was 1043)
- [x] Typecheck passes — `pnpm typecheck`
- [x] Lint passes — `pnpm lint` (eslint --max-warnings 0)
- [x] Build succeeds — `pnpm build` (run by pre-push hook)
- [ ] CI green on this PR
- [x] Manual scenarios validated by tests:
  - [x] `writeIssue` rejects an issue with a 600-char title (poison-prevention)
  - [x] `writeIssue` rejects an empty title
  - [x] `writeIssue` accepts a title at the 500-char schema max
  - [x] `listIssues` returns valid issues and skips invalid files instead of throwing (the original `tbd list → 0 issues` regression)
  - [x] `console.warn` only ever receives strings, even when the underlying parse error is a `ZodError` (the Node v24 `util.inspect` crash)
  - [x] `listIssuesDetailed` returns the skipped file list with non-empty `reason` strings
  - [x] `formatIssueParseError` handles `ZodError`, `Error`, strings, and numbers
- [ ] Edge cases worth a manual smoke before merge:
  - [ ] In a real repo with a corrupt `.md` file, `tbd list`/`status`/`stats` now succeed and show a one-line skip summary instead of crashing
  - [ ] `tbd doctor` flags the corrupt file under "Issue validity" with the schema reason
  - [ ] `tbd doctor` "Temp files" now reports the correct path and detects leftover `.md.tmp-NNNN` files
  - [ ] `tbd create "<600 char string>"` returns the friendly "Invalid title" error instead of writing

## Related Beads

None — surfaced directly from issue #115.

Closes #115.

---
_Generated by [Claude Code](https://claude.ai/code/session_013LFvTrSBqryQoFaLc9ijdr)_

### DIFF
--- a/packages/tbd/src/cli/commands/create.ts
+++ b/packages/tbd/src/cli/commands/create.ts
@@ -19,7 +19,7 @@ import {
   addIdMapping,
   resolveToInternalId,
 } from '../../file/id-mapping.js';
-import { IssueKind } from '../../lib/schemas.js';
+import { IssueKind, IssueSchema } from '../../lib/schemas.js';
 import { parsePriority } from '../../lib/priority.js';
 import { resolveDataSyncDir } from '../../lib/paths.js';
 import { now } from '../../utils/time-utils.js';
@@ -47,6 +47,21 @@ class CreateHandler extends BaseCommand {
     // Validate title is provided (unless --from-file)
     if (!title && !options.fromFile) {
       throw new ValidationError('Title is required. Use: tbd create "Issue title"');
+    }
+
+    // Validate title length up-front against the schema constraint, so users
+    // get a clean error before we attempt to write. Without this, agents that
+    // dump a multi-paragraph description into the title field would get a raw
+    // ZodError later (or, before #115, would silently poison the data store).
+    if (title !== undefined) {
+      const titleCheck = IssueSchema.shape.title.safeParse(title);
+      if (!titleCheck.success) {
+        const reason = titleCheck.error.issues[0]?.message ?? 'invalid title';
+        throw new ValidationError(
+          `Invalid title (${title.length} chars): ${reason}. ` +
+            `Move detail into the description (-d / -f).`,
+        );
+      }
     }
 
     // Parse and validate options
--- a/packages/tbd/src/cli/commands/doctor.ts
+++ b/packages/tbd/src/cli/commands/doctor.ts
@@ -12,7 +12,7 @@ import { join } from 'node:path';
 
 import { BaseCommand } from '../lib/base-command.js';
 import { requireInit } from '../lib/errors.js';
-import { listIssues } from '../../file/storage.js';
+import { listIssuesDetailed } from '../../file/storage.js';
 import { readConfig } from '../../file/config.js';
 import type { Config, Issue, IssueStatusType } from '../../lib/types.js';
 import { resolveDataSyncDir, TBD_DIR, WORKTREE_DIR, DATA_SYNC_DIR } from '../../lib/paths.js';
@@ -73,9 +73,12 @@ class DoctorHandler extends BaseCommand {
       // Config may be invalid - will be caught by health checks
     }
 
-    // Load issues
+    // Load issues. Use listIssuesDetailed so doctor doesn't trigger
+    // listIssues's per-file console warnings (the validity check below
+    // already surfaces skipped files cleanly).
     try {
-      this.issues = await listIssues(this.dataSyncDir);
+      const result = await listIssuesDetailed(this.dataSyncDir);
+      this.issues = result.issues;
     } catch {
       // May fail if no issues yet
     }
@@ -113,8 +116,9 @@ class DoctorHandler extends BaseCommand {
     // Check 7: Orphaned temp files
     healthChecks.push(await this.checkTempFiles(options.fix));
 
-    // Check 8: Issue validity
-    healthChecks.push(this.checkIssueValidity(this.issues));
+    // Check 8: Issue validity (re-parses every issue file from disk so we
+    // catch schema violations that listIssues silently skipped — see #115).
+    healthChecks.push(await this.checkIssueValidity());
 
     // Check 9: Worktree health (with fix support)
     // Run BEFORE ID mapping check — worktree repair and data migration can
@@ -130,7 +134,8 @@ class DoctorHandler extends BaseCommand {
     if (dataLocationResult.status === 'ok' && dataLocationResult.message?.includes('migrated')) {
       this.dataSyncDir = await resolveDataSyncDir(this.cwd);
       try {
-        this.issues = await listIssues(this.dataSyncDir);
+        const result = await listIssuesDetailed(this.dataSyncDir);
+        this.issues = result.issues;
       } catch {
         // Will be caught by other health checks
       }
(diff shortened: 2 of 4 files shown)


jlevy:
Closing as superseded by #116 (merged), which fixed the underlying #115 bug with a similar but more refined approach (`onInvalidIssue` callback in `listIssues`, shared `validateIssueTitle` helper, `zod-error-utils.ts`).

The only items in this PR not already on main are two small `checkTempFiles` improvements in `doctor.ts`:
- Display path is wrong (`.tbd/issues` instead of the real `.tbd/data-sync/issues`)
- Filter misses `atomically`'s `.md.tmp-NNNN` leftovers, only catches `.tmp`

Those will be ported to a separate focused PR.

---

PR #121, opened 2026-05-17
feat: Use shared common-dir sync worktree

jlevy:
## Summary

Move tbd's local sync worktree from each checkout's `.tbd/data-sync-worktree/` into the repository Git common directory so the main checkout and linked worktrees share one `tbd-sync` worktree. This fixes the branch-ownership conflict seen when Codex alternates between the main checkout and Codex-created worktrees, while preserving f03 migration and failing closed for older clients after f04 upgrade.

## Changes

- Add the `f04` local layout format with `sync.storage: git-common-dir-v1` in `.tbd/config.yml` and matching `$GIT_COMMON_DIR/tbd/layout.yml` metadata.
- Resolve shared tbd paths from `git rev-parse --git-common-dir` and route issue reads/writes through `$GIT_COMMON_DIR/tbd/data-sync-worktree/.tbd/data-sync/`.
- Serialize shared sync worktree initialization, repair, migration, and mutating commands with the existing mkdir-based lockfile pattern.
- Migrate legacy per-checkout `.tbd/data-sync-worktree/` data into the shared worktree and remove legacy branch ownership before the shared worktree claims `tbd-sync`.
- Update setup, init, sync, status, doctor, save/import, uninstall, and command data-context flows to understand the shared layout.
- Add clearer newer-version errors for future `.tbd` or common-dir layout formats.
- Add unit and golden CLI coverage for linked worktrees, f03-to-f04 migration, old-client rejection, and future-format upgrade errors.
- Update the active spec, troubleshooting/design docs, and release changeset.

## Scope Guard

This PR is only the shared common-dir sync worktree implementation. The multi-agent skills and hooks planning work was moved to PR #131.

## Test Plan

- [x] TypeScript typecheck passed (`pnpm --filter get-tbd typecheck`).
- [x] Formatting passed (`pnpm format:check`).
- [x] Lint passed (`pnpm lint:check`).
- [x] Production build passed (`pnpm --filter get-tbd build`).
- [x] Unit/e2e test suite passed (`pnpm --filter get-tbd test`, 57 files / 1071 tests).
- [x] Full golden CLI suite passed (`pnpm --filter get-tbd test:tryscript`, 795 scenarios).
- [x] Pre-push hook passed before pushing the shared worktree implementation.
- [x] Pre-push hook passed after restoring this branch to worktree-only scope (`pnpm build:check`, `pnpm test`; 57 files / 1071 tests).
- [x] Manual scenario covered by golden tests: f03 legacy per-checkout sync worktree owns `tbd-sync`, linked worktree reproduces Git's duplicate branch ownership failure, first new-client write migrates to f04 shared common-dir layout, main checkout and linked worktree both create issues through one shared store.
- [x] Compatibility scenario covered by golden tests: an f03-era client rejects an f04 repo instead of writing the old per-checkout layout.
- [x] Future-format scenario covered by golden tests: a repo with `tbd_format: f99` reports that it requires a newer tbd and includes the supported format plus upgrade command.
- [x] Edge cases considered: unknown main checkout location, linked-worktree Git common-dir resolution, legacy unsynced data preservation, existing attached worktree branch ownership, stale older clients after migration, future layout format bumps.

## Related Beads

- tbd-vbrq: Design shared common-dir sync worktree
- tbd-vl06: Spec: Shared common-dir sync worktree implementation
- tbd-pxio: Phase 1 spike for shared Git common-dir attached sync worktree
- tbd-1fp5: Add Git common-dir path and layout helpers
- tbd-ke1g: Add f04 format and shared layout metadata guards
- tbd-0622: Migrate legacy per-checkout sync worktrees
- tbd-qflw: Update setup, doctor, status, and uninstall for shared layout
- tbd-pkwi: Add linked-worktree migration and versioning tests
- tbd-ytef: Update docs, troubleshooting, and release notes
- tbd-k5xm: Add shared common-dir migration compatibility scenario tests


jlevy:
# Senior Engineering Review — PR #121

## Summary

A well-scoped, well-tested implementation of the spec at `plan-2026-05-17-shared-common-dir-sync-worktree.md`. Code quality is high, the refactor of mutating commands to thread through a single `withDataSyncContext({ lock: true }, …)` is a clear improvement, and the test coverage (golden tryscript for linked worktrees, migration, future-format rejection; unit tests for `f04` migration; worktree-health test for linked worktrees) hits the cases that mattered. CI is green on all platforms.

I’d still hold the PR for two items below (one bug, one missing piece referenced by user-facing error text). Everything else is minor or non-blocking.

* * *

## Blocking issues

### 1. Lock timing invariant is violated → stuck commands on crashes

`packages/tbd/src/utils/lockfile.ts:42` documents:

> IMPORTANT: `timeoutMs` must be greater than `staleMs` so stale locks from crashed processes are always detected and broken before the timeout expires.

The new shared profile inverts that:

```ts
export const DATA_SYNC_LOCK_OPTIONS: Required<LockfileOptions> = {
  timeoutMs: 120_000,    // 2 min
  pollMs: 100,
  staleMs: 30 * 60_000,  // 30 min   ⟵ greater than timeoutMs
};
```

Concrete failure: if a sync process is SIGKILLed (OOM, terminal close, CI cancel) while holding `$GIT_COMMON_DIR/tbd/locks/data-sync.lock`, the next command waits 2 minutes, then throws `LockAcquisitionError` telling the user to “delete the lock directory manually.” The stale check (30 min) never fires within the timeout.

This isn’t a corner case — network hangs during `tbd sync` are exactly when crashes/cancels happen.

Options:
- Accept long sync runtimes as legitimate and set both windows higher (e.g. `timeoutMs: 35 * 60_000`), keeping the documented `timeoutMs > staleMs` invariant.
- Or land the heartbeat metadata path that the spec explicitly flags as the alternative (§Locking Model, “heartbeat metadata inside the lock directory so long-running sync operations do not get broken as stale”).
- Or at minimum, update the lockfile.ts invariant comment to explain why this profile is the documented exception and have `tbd doctor` surface a clear “lock looks stale, here’s the path to remove” hint so users aren’t stuck.

### 2. `CommonDirLayoutError` tells users to run `tbd doctor --fix`, but doctor doesn’t fix it

`packages/tbd/src/file/common-dir-layout.ts:60` and `:69` raise errors saying “Run `tbd doctor --fix` to repair.” The spec is also explicit (§Format And Layout Versioning, §Compatibility statement):

> If `.tbd/config.yml` and `$GIT_COMMON_DIR/tbd/layout.yml` disagree on `tbd_format`, normal mutating commands should fail closed and route repair through `tbd doctor --fix`.

`packages/tbd/src/cli/commands/doctor.ts` has no awareness of `layout.yml` — no `readCommonDirLayout`, no `writeCommonDirLayout`, no `withSharedDataSyncLock`. So if a user lands in this state (e.g., partial migration, half-applied upgrade, manually-edited layout), the error sends them to a tool that can’t help.

Either implement the diagnostic + `--fix` path in `doctor.ts`, or change the error message to describe the actual manual recovery. The first is the contract the spec promised.

* * *

## High-priority issues

### 3. Read-only commands now acquire the heavy 2-minute, 30-minute-stale lock

`atticList`, `atticShow`, `labelList`, etc. were rewritten to call `loadDataContext`, which always wraps in `withSharedDataSyncLock`. That serializes trivial reads behind any concurrent sync. The spec explicitly anticipates this (§Locking Model: “Read-only commands can eventually avoid the lock, but the first implementation should prefer correctness”), so this is conscious — but combined with issue (1), an interrupted sync can lock out **every** tbd read for 2 minutes. Worth either:
- splitting `loadDataContext` into a `lock: false` variant for pure reads (one-line change since `withDataSyncContext` already takes `lock`), or
- ensuring the lockfile fix above before users hit it.

### 4. Duplicate `WorktreeMissingError` class (pre-existing, but this PR re-touches both files)

Two unrelated classes with the same name:
- `packages/tbd/src/lib/paths.ts:400` — extends `Error`, this is the one actually thrown.
- `packages/tbd/src/cli/lib/errors.ts:86` — extends `CLIError`, dead code.
- `WorktreeCorruptedError` in `errors.ts:100` is also never thrown.

The PR edits both files. Cheap fix: delete the unused `errors.ts` versions or wire them up. Right now an IDE go-to-definition is ambiguous.

* * *

## Medium-priority issues

### 5. Redundant validation + redundant config write in setup migration path

`prepareDataSyncContext` (`packages/tbd/src/cli/lib/data-context.ts:91-123`):

```ts
const existingLayout = await readCommonDirLayout(...);
if (existingLayout) { validateCommonDirLayout(existingLayout, config); }   // ① validation
...repairWorktree...
if (existingLayout) { validateCommonDirLayout(existingLayout, config); }   // ② identical re-validation, dead
else { await writeCommonDirLayout(...); }
```

Block ② re-validates the **same in-memory snapshot** with no mutation in between. It’s dead code; drop it or re-read the layout if the intent was to detect interleaved writes.

Related: in `setup.ts:1112`, `withDataSyncContext(... async () => undefined)` calls `prepareDataSyncContext`, which writes the migrated config; then `setup.ts:1117` writes the config again. Harmless but confusing — pick one writer.

### 6. `paths.ts` cache returns stale fallback after worktree is created

`resolveDataSyncDir` (`packages/tbd/src/lib/paths.ts:435`) caches keyed on `(baseDir, allowFallback)`. When `allowFallback: true` returns the fallback direct path because the worktree didn’t exist yet, a later call with the same key returns the **cached fallback** even after the worktree is created. Production only calls `resolveDataSyncDir` with `allowFallback: false` (in `prepareDataSyncContext`), so this is dormant — but it’s a foot-gun for future callers (and for tests that don’t call `clearPathCache()`). Either cache only successful worktree resolutions, or never cache the fallback.

### 7. `WORKTREE_DIR` / `DATA_SYNC_DIR_VIA_WORKTREE` constants are misleading after this change

`packages/tbd/src/lib/paths.ts:59` defines `WORKTREE_DIR = '.git/tbd/data-sync-worktree'` (primary-checkout-only) with a comment that says “Production code should use `resolveSharedTbdPaths()`”. It’s still exported and used in `tests/worktree-health.test.ts:815`. The combination of name (sounds canonical) + “don’t actually use this” comment is exactly the kind of constant that grows new misuse. Suggest renaming to something like `PRIMARY_CHECKOUT_SHARED_WORKTREE_PATH` or marking `@internal` / test-only. Same for `DATA_SYNC_DIR_VIA_WORKTREE`.

### 8. `uninstall.ts` swallows the real reason for shared-paths failure

```ts
const sharedPaths = await resolveSharedTbdPaths(tbdRoot).catch(() => null);
```
(`packages/tbd/src/cli/commands/uninstall.ts:44`)

If we’re not in a git repo or `git rev-parse` fails for some other reason, the user sees only “Could not remove shared common-dir metadata” later — no clue why. Worth logging at `output.debug` level so `--debug` shows the real error.

### 9. Recurring `let didX = false; … if (!didX) return;` pattern across commands

`close.ts`, `update.ts`, `label.ts`, `dep.ts`, `attic.ts`, `create.ts` all repeat:

```ts
let didX = false;
await this.execute(async () => {
  await withDataSyncContext(..., async (ctx) => { ...; didX = true; });
}, '...');
if (!didX) return;
this.output.data(...);
```

The intent (skip success output on dry-run / no-op) is fine, but six near-identical copies will rot. A `runWithLock(ctx, fn): Promise<{ didMutate, displayId? } | null>` helper, or just letting the callback return a discriminated `{ kind: 'noop' | 'done', displayId }` and rendering uniformly, would compress this and remove the captured-variable-via-closure pattern (which currently means displayId can silently stay as the user-input form if mutation is skipped — see how `displayId = id` is the unconverted initial value).

### 10. Spec checkboxes still unchecked despite status “Implemented”

`docs/project/specs/active/plan-2026-05-17-shared-common-dir-sync-worktree.md` says “Status: Implemented” at the top but every `- [ ]` checkbox in the Implementation Plan and Testing Strategy is empty. Either tick the boxes that are actually done or move the spec to `done/`. (`tbd shortcut update-specs-status` exists for exactly this.)

* * *

## Low-priority / nits

- `packages/tbd/src/file/git.ts:1054` builds a backup branch name from a filename-timestamp then strips non-`[0-9A-Za-z-]` characters. The timestamp is already filename-safe, so the strip is dead. Not wrong, just noise.
- `packages/tbd/src/cli/commands/sync.ts:646` does `await import('node:fs/promises')` inline twice (one in clean-merge branch at :790 too). These are already statically imported elsewhere in the file — replace with top-level imports.
- `packages/tbd/src/file/git.ts:1029` does `git branch syncBranch head` when the local branch doesn’t exist. If a remote-tracking ref named `tbd-sync` exists but no local `refs/heads/tbd-sync`, this is fine; if anything weirder lurks (reflog-only, packed-refs glitches), the error surface is opaque. A defensive `--force` with a `branchExists` check, or a clearer error, would help during the migration phase.
- `packages/tbd/src/cli/commands/setup.ts:1112` — `withDataSyncContext(projectDir, { lock: true }, async () => undefined)` reads like dead code. A comment (“initializes shared layout under the lock”) would prevent a future cleanup PR from deleting it.
- `packages/tbd/src/lib/tbd-format.ts:413` — `describeMigration('f01')` returns 3 lines now, but the test description still says “(two steps)” (`packages/tbd/tests/tbd-format.test.ts:241`).
- `.changeset/shared-common-dir-worktree.md` is a `minor` bump but should arguably be highlighted in release notes given older clients break — add an upgrade note (the spec’s Rollout Plan lists exactly what should appear).
- The design doc still keeps `.tbd/backups/` in the diagram (`tbd-design.md` after the `└── backups/` line) but production migration backups now live under `$GIT_COMMON_DIR/tbd/backups/`. Clarify which (if either) is still used.

* * *

## What’s done well

- Spec-driven implementation that visibly tracks back to the spec (good cross-references in code comments).
- `f04` migration is **idempotent** and the staged write order (layout → config) matches the spec, so partial-failure recovery is well-defined.
- Golden tryscript at `cli-shared-common-dir-worktree.tryscript.md` actually reproduces the original branch-ownership failure mode in the `before` block, which is the right way to write a regression test for this class of bug.
- Old-client compatibility guard is tested both as a unit test (`tbd-format.test.ts`) and end-to-end in tryscript form (with a stand-in f03 client).
- Removing the double-print of `CLIError` messages in `base-command.ts` cleans up a longstanding minor wart.
- `migrateLegacyWorktreesToShared` correctly distinguishes ancestor / diverged / clean cases and bails to a backup branch rather than silently losing data on divergence.

## Recommendation

Hold for items **1** and **2**; merge after those are addressed. Items 3–6 can be follow-up beads (would be reasonable to ship together with #1 as a single lock-hardening PR). The rest are cleanups.

jlevy:
## Senior engineering review - PR #121 current head

I reviewed the current head (`1e1ff21`) after the follow-up commit that addresses the earlier review. CI is green across Benchmark, Coverage & Lint, DeepSource Secrets, and macOS/Ubuntu/Windows tests. There are no unresolved GitHub review threads. I did not see dependency changes, so the supply-chain install rules do not apply here.

### Summary

The high-level design is the right direction: a single repo-scoped sync worktree under `$GIT_COMMON_DIR/tbd/` is simpler and safer than trying to keep per-checkout worktrees coordinated. The attached shared worktree plus a common lock fits Git's own ownership model, removes the duplicate `tbd-sync` branch checkout problem, and gives us a clean second guard against stale old clients after the `f04` config migration.

I would still hold the PR for one locking bug before merge. The previous review items were mostly addressed well, but the read-only lock fix changed one failure mode into another: reads no longer wait behind the heavy lock, but they can still initialize, repair, migrate, and write layout/config without that lock.

### Blocking issue

#### 1. Read-only commands can still mutate shared sync state without the shared lock

`loadDataContext()` now calls `withDataSyncContext(tbdRoot, { lock: false }, ...)` at `packages/tbd/src/cli/lib/data-context.ts:164`, but the function it runs is still `prepareDataSyncContext()` at `packages/tbd/src/cli/lib/data-context.ts:87`.

That preparation path is not read-only:

- It migrates config in memory via `readConfigWithMigration()` at `packages/tbd/src/cli/lib/data-context.ts:88`.
- It checks and can repair missing/prunable shared worktrees at `packages/tbd/src/cli/lib/data-context.ts:96` and `packages/tbd/src/cli/lib/data-context.ts:100`.
- It writes `$GIT_COMMON_DIR/tbd/layout.yml` at `packages/tbd/src/cli/lib/data-context.ts:119`.
- It writes the migrated `.tbd/config.yml` at `packages/tbd/src/cli/lib/data-context.ts:123`.

That means commands like `tbd list`, `tbd ready`, `tbd search`, `tbd blocked`, `tbd stale`, `tbd stats`, and attic/label/dep read paths can race f03-to-f04 migration, missing-layout initialization, legacy worktree removal, and shared worktree creation without `$GIT_COMMON_DIR/tbd/locks/data-sync.lock`.

This violates the central safety property in the spec: migration/initialization/repair must be serialized. It also makes the race easiest to trigger during the exact first-use path that matters: two agents or an agent plus a user both run benign read commands from sibling worktrees after upgrade or fresh clone. One may be removing the legacy per-checkout worktree or claiming `tbd-sync` while the other is doing the same, but neither holds the shared lock.

Suggested fix:

- Split the current `prepareDataSyncContext()` into a pure read path and a locked ensure/migration path, or make `loadDataContext()` use a helper that first detects whether migration/repair/layout write is needed and only skips the lock when the shared layout and worktree are already valid.
- Keep the "reads should not wait behind a long sync" optimization only for the truly read-only steady state.
- Add a regression test that runs first-use read commands concurrently from two linked worktrees against an f03 repo or an f04 repo with missing `layout.yml`, and asserts exactly one shared worktree/layout is produced and no legacy/direct data path is written.

### High-priority follow-up

#### 2. Some direct setup/init/doctor repair paths still bypass the shared lock

The same design invariant is weaker in a few direct CLI paths:

- `tbd init` calls `initWorktree()` directly at `packages/tbd/src/cli/commands/init.ts:176`.
- fresh `tbd setup --auto` calls `initWorktree()` directly at `packages/tbd/src/cli/commands/setup.ts:1477`.
- `tbd doctor --fix` calls `repairWorktree()` directly at `packages/tbd/src/cli/commands/doctor.ts:839` and `initWorktree()` directly at `packages/tbd/src/cli/commands/doctor.ts:925`.
- `initWorktree()` itself runs `migrateLegacyWorktreesToShared()` at `packages/tbd/src/file/git.ts:1124`, so these direct paths can perform legacy migration and branch/worktree mutation.

This is less likely than the read-command race, but it is the same class of issue. I would prefer a single internal "ensure shared sync layout under lock" entry point so setup/init/doctor/data-context cannot drift. If `initWorktree()` stays public for tests, document it as requiring an outer lock or wrap the CLI callers in `withSharedDataSyncLock()`.

### Documentation / contract gap

The code now tells users to manually remove a mismatched layout file at `packages/tbd/src/file/common-dir-layout.ts:57` and `packages/tbd/src/file/common-dir-layout.ts:68`, which is a reasonable minimal recovery story. The spec still says mismatched config/layout should route repair through `tbd doctor --fix` at `docs/project/specs/active/plan-2026-05-17-shared-common-dir-sync-worktree.md:320` and marks that behavior complete at `docs/project/specs/active/plan-2026-05-17-shared-common-dir-sync-worktree.md:572`.

Either implement the doctor diagnostic/fix for layout mismatch, or update the spec to match the manual recovery contract. Since this repo treats specs as source-of-truth design artifacts, I would not leave the mismatch in place.

### Design assessment

I do not see a simpler general approach that handles the real constraints better than this one:

- Per-checkout worktrees with `tbd-sync` attached are the source of the branch ownership failure.
- Per-checkout detached worktrees avoid branch ownership but push complexity into manual ref advancement, dirty state, and compare-and-swap style ordering.
- A "main checkout owns the sync worktree" design is fragile in Codex/cloud/linked-worktree environments where the primary checkout may be unknown or not the place the agent should mutate.
- A daemon, database, or git-plumbing-only implementation would be more invasive than the problem requires.

The common-dir attached worktree is the right compromise: one physical sync worktree per Git repository, Git still owns branch/ref correctness, and the common lock can make mutation reasoning tractable. The remaining work is to make that lock boundary absolute around initialization, migration, repair, and writes.

### What looks good in the follow-up fixes

- The shared lock timing invariant is fixed (`timeoutMs` now exceeds `staleMs`).
- Duplicate/dead `WorktreeMissingError` and `WorktreeCorruptedError` definitions were removed.
- The stale fallback cache in `resolveDataSyncDir()` was fixed by not caching fallback direct paths.
- The primary-checkout-only constants are renamed and documented as non-canonical.
- `uninstall` now preserves useful debug context when shared path resolution fails.
- The changeset and design docs now call out the `f04`/older-client upgrade behavior.
- The golden tryscript is strong: it reproduces the old branch-ownership failure, migrates f03 to f04, proves main and linked worktree writes share one store, and checks old-client rejection.

### Residual risks / non-blocking suggestions

- The 30-minute stale-lock window is probably acceptable for this data size, but there is still no heartbeat. A live sync that hangs longer than `staleMs` can be broken by another process. I would either explicitly accept that risk in the spec or file a follow-up for heartbeat metadata.
- `doctor` still hides future-format config details behind "Invalid config file" rather than surfacing the upgrade message. Not a blocker for this PR, but it is worth tightening because `doctor` is where users go during broken upgrades.
- Coverage on the large CLI command files remains low per the report. The new golden tests cover the main workflow well, but the specific lock-boundary cases above need direct tests.

### Recommendation

Hold for item 1 before merge. Ideally fix item 2 in the same pass by centralizing shared-layout ensure/repair behind one locked helper, then either update or implement the doctor/layout-mismatch contract. After that, I would be comfortable merging this design.

PR #140, opened 2026-05-29
fix(release): use CHANGELOG section in GH Release body; gate tag on green main CI

jlevy:
Two release-process fixes surfaced when v0.2.0 went out.

## Summary

- `.github/workflows/release.yml` — fix the `Extract changelog for this version` step. v0.1.30 and v0.2.0 both shipped with the GH Release body set to the fallback `Release vX.Y.Z` string instead of the `## X.Y.Z` CHANGELOG section. Closes **tbd-xk7c**.
- `docs/publishing.md` — gate tagging on **main CI conclusion=success on the merge commit**, and explicitly verify the GH Release body after publish (catches a release.yml regression even before user review). Closes **tbd-uvvc**.

## Root cause — release.yml

The awk range pattern `/^## ${VERSION//./\\.}$/,/^## [0-9]/` collapses to a single line because the START pattern (`## 0.2.0`) **also matches the END pattern** (a `##` followed by a digit). awk's inclusive range therefore terminates on the same line it began, the pipeline writes a single heading, `head -n -1` drops it, and the empty-result fallback wins. The step exited 0 either way, so nothing surfaced.

Fix: the changelog extraction now lives in a unit-tested TypeScript module (`packages/tbd/src/utils/changelog.ts`) that matches the **exact** `## X.Y.Z` heading and scans to the next `## [0-9]` heading. `release.yml` calls a thin `tsx` wrapper (`packages/tbd/scripts/extract-changelog.ts`) and only wires its stdout into `GITHUB_OUTPUT`. The body is resolved **before** `npm publish` (so a wrapper failure fails the run before shipping) using a random output delimiter.

Verified locally on v0.2.0's section — the full body (f04 upgrade contract, multi-worktree notice, hardening section) is extracted. v0.2.0's GH Release body has been backfilled via `gh release edit`; v0.1.30 still has the fallback string (optional backfill).

## Root cause — publishing.md gate

I declared v0.2.0 "live" while main CI on the merge commit (`20303cf`) was still `in_progress` — 4/5 jobs green, the Windows lockfile flake hanging. The user correctly flagged this as a process failure: a release isn't "pushed successfully" if CI hasn't confirmed green on the merge commit.

`release.yml` is independent of `ci.yml` (it triggers on the tag and runs on `ubuntu-latest`), so the publish itself succeeds either way — but skipping the main-CI gate ships work that wasn't confirmed to pass tests on the merge commit.

publishing.md now states: **main CI on the merge commit MUST reach `conclusion=success` before tagging.** Every runnable path (Step 6 options + both quick references + the development.md summary) computes the merge SHA and waits on the CI run **for that exact SHA** before tagging. If a job hangs on a known flake, cancel and rerun. Don't tag past it.

## Review follow-ups (addressed)

- **[P1]** Release docs no longer let copy-paste paths tag before post-merge main CI is green; the gate filters by the merge commit SHA (an unfiltered `--limit 1` could return the previous run).
- **[P2]** `withLockfile` no longer renames a non-directory at the lock path — a regular file there now throws instead of being moved aside (which let the critical section run unprotected). Regression test added. (The atomic stale-break + resilient release for real lock directories remain.)
- Workflow hardening: extract body before publish, random `GITHUB_OUTPUT` delimiter, and a shell-free CLI smoke test for the wrapper.

## Related beads

- `tbd-vugg` — `tests/lockfile.test.ts > serializes concurrent access` Windows flake (addressed by the atomic stale-break + low-staleMs/timeout hardening in this PR)
- `tbd-72xm` — bump `pnpm/action-setup@v4` and `softprops/action-gh-release@v2` (Node 20 deprecation, June 2 2026)

## Test plan

- [ ] CI green
- [ ] After merge, the next release will exercise the fixed `Extract changelog` step. Confirm `gh release view <tag> --json body --jq .body | head` returns the `## X.Y.Z` section, not the fallback.

https://claude.ai/code/session_016Sc2W6J85NmVdbJCjpVbC9


.github/workflows/release.yml
```diff
@@ -45,9 +45,25 @@ jobs:
...
+          # followed by a digit), so the range collapses to a single line and
+          # the fallback "Release vX.Y.Z" body kicks in. Both v0.1.30 and v0.2.0
+          # were published with the empty fallback.
+          #
+          # Fix: explicitly print the start line and use a flag so the end
+          # pattern only fires on the NEXT `## [0-9]` heading.
           VERSION="${{ steps.version.outputs.version }}"
-          CHANGELOG=$(awk "/^## ${VERSION//./\\.}$/,/^## [0-9]/" packages/tbd/CHANGELOG.md | head -n -1)
+          ESCAPED_VERSION="${VERSION//./\\.}"
+          CHANGELOG=$(awk -v ver="^## ${ESCAPED_VERSION}\$" '
```
jlevy: We should not perform obscure bash commands inside of GitHub Actions workflows. This should be generally banned in our developer doc recommendations. This is a bad practice because it is impossible to debug reliably without the very slow process of running GitHub Actions. If something like this needs to be written, it should be a clean Python or TypeScript script that can be tested in isolation and then used by reference from a CI call. We already have various developer scripts like this, and this one should be no different.

jlevy: Fixed in 226713d. Moved the changelog extraction out of the workflow into a unit-tested module:

- `packages/tbd/scripts/extract-changelog.mjs` — pure exported `extractChangelogSection` / `resolveReleaseBody` (with a `.d.mts`), following the existing `git-version.mjs` script pattern.
- `packages/tbd/tests/extract-changelog.test.ts` — 10 cases including a regression for the awk-range collapse (full section vs single line), last-section-in-file, literal-dot matching, prerelease headings, CRLF, and the missing-version fallback.
- `release.yml` now only calls `node packages/tbd/scripts/extract-changelog.mjs "$VERSION"` and wires its stdout into the `GITHUB_OUTPUT` heredoc — no inline awk.

Also added a **"CI and GitHub Actions"** section to `docs/development.md` codifying the rule: no non-trivial inline shell (awk/sed/jq/regex/conditionals) in workflow `run:` steps — extract to a unit-tested script invoked by reference, citing this exact bug as the cautionary example.

---

jlevy (commented):
## Senior engineering review

### Findings

1. **[P1] The release docs still let the copy-paste paths tag before post-merge main CI is green.**

   The new gate is stated in `docs/publishing.md:193`, but the runnable examples still contradict it. The local quick reference still does `git push && git tag v0.2.0 && git push --tags` at `docs/publishing.md:282`, and the restricted flow still merges and immediately creates the tag at `docs/publishing.md:305`-`docs/publishing.md:307`. `docs/development.md:242`-`docs/development.md:245` also summarizes the release flow as "merge once CI is green" and then tag, without the new requirement that **main CI on the merge commit** has reached `conclusion=success`.

   There is a second, related footgun in the gate command itself: `docs/publishing.md:200` uses `gh run list --branch main --workflow=ci.yml --limit 1`, but it does not filter to the commit being tagged. Immediately after a push/merge, this can show the previous successful `main` run before the new run exists or completes. That recreates the exact failure mode this PR is meant to prevent.

   Suggested fix: make every runnable path compute the commit to be tagged and wait on CI for that SHA before tagging. For example, after direct push use `MERGE_SHA=$(git rev-parse HEAD)`; after PR merge use the PR merge commit SHA; then query with `gh run list --branch main --workflow=ci.yml --commit "$MERGE_SHA" ...`, verify `headSha == MERGE_SHA`, watch that run to completion, and tag only after `conclusion=success`. Update the quick references and the short `docs/development.md` release summary so the copy-paste path cannot bypass the policy.

2. **[P2] `withLockfile` now renames any stale path, including a regular file, and then runs the critical section.**

   The new atomic stale break is the right direction for real stale lock directories, but `packages/tbd/src/utils/lockfile.ts:210` only calls `stat()` and `packages/tbd/src/utils/lockfile.ts:213` then calls `breakStaleLock(lockPath)` without verifying the path is actually a directory. `breakStaleLock()` at `packages/tbd/src/utils/lockfile.ts:146`-`packages/tbd/src/utils/lockfile.ts:154` uses `rename()`, so if a regular file exists at the lock path, it is moved aside to `*.stale-<uuid>`, `removeLockDir()` silently fails to remove it, `mkdir(lockPath)` then succeeds, and the caller's protected callback runs.

   I reproduced this on `5610b12` by pre-creating `file.lock` as a regular file and calling `withLockfile(file.lock, ..., { staleMs: -1 })`: the callback ran, the original path disappeared, and a `file.lock.stale-...` sidecar was left behind. The previous `rmdir(lockPath)` behavior would not remove a file, so this is a regression in unexpected filesystem-state handling.

   Suggested fix: only break stale locks when `lockStat.isDirectory()` is true. If the path exists but is not a directory, do not rename it; either keep waiting until timeout or throw a specific error that says the lock path is occupied by a non-directory. Add a regression test that pre-creates a regular file at the lock path, asserts the callback does not run, and asserts the file remains in place.

### Design assessment

The changelog extraction approach is a solid improvement: the release-body logic is now a pure TypeScript function with unit coverage, and the workflow only wires stdout into `GITHUB_OUTPUT`. That is much easier to reason about than inline awk in Actions. The test cases cover the important regression surface: full section extraction, middle and last sections, literal dots, prereleases, CRLF, and missing-version fallback.

The stale-lock change is also conceptually right: rename-before-remove is the correct way to avoid multiple waiters breaking the same stale directory. It just needs the filesystem-type guard above so the atomic rename is only applied to directories created by this lock protocol.

### Strengths

- Good movement of non-trivial workflow logic out of YAML and into tested source.
- Good regression coverage for the changelog extractor, especially the literal-match and range-collapse cases.
- The lockfile concurrency test now checks actual overlap with `maxActive`, which is stronger than only checking paired order.
- Current PR CI is green across Ubuntu, macOS, Windows, coverage/lint, benchmark, and DeepSource.

### Documentation gaps

- The release-gate policy needs to be reflected everywhere a human or agent might copy commands: the main step-by-step flow, both quick references, and the short release summary in `docs/development.md`.
- `docs/publishing.md:168` still says `release.yml` "greps" for the changelog heading. That is minor, but now stale; it should say the release extractor matches the exact `## X.X.X` heading.
- The PR body still describes the fix as "flag-based extraction," while the current implementation is a TypeScript scanner. Not a merge blocker, but worth aligning before merge because this PR is process documentation as much as code.

### Unresolved comments

No unresolved review threads found. The prior inline concern about obscure shell in Actions is resolved and outdated in GitHub's thread state.

### Suggestions

- Add a small script-level smoke test for `packages/tbd/scripts/extract-changelog.ts`, not just the pure module, so the workflow's exact wrapper/default path stays covered outside tag CI.
- Consider extracting the changelog body before `pnpm -r publish` in `.github/workflows/release.yml`. If the wrapper or path ever fails, it is better to fail before publishing to npm than after the package has already been published.
- Consider using a unique `GITHUB_OUTPUT` delimiter instead of fixed `EOF`; it is unlikely but avoids release body corruption if a changelog line ever equals `EOF`.

### Validation performed

- `gh pr checks 140 --repo jlevy/tbd`: all checks passing.
- `pnpm install --frozen-lockfile` in a detached PR worktree.
- `pnpm --filter get-tbd typecheck`: passed.
- `pnpm --filter get-tbd test -- tests/extract-changelog.test.ts tests/lockfile.test.ts`: passed; due the script/runner behavior this executed the full Vitest suite, `60` files and `1102` tests.
- `pnpm exec tsx packages/tbd/scripts/extract-changelog.ts 0.2.0`: printed the `## 0.2.0` changelog section as expected.

I would address the two findings before merging. The core implementation is close; the release-process docs need to be mechanically copy-paste-safe, and the lockfile stale-break path needs to preserve unexpected non-directory state instead of moving it aside.

jlevy:
Thanks — both findings addressed in `4e4c147`.

**[P1] Release-tag CI gate is now copy-paste-safe.** Every runnable path computes the commit being tagged and waits on main CI **for that SHA** before tagging:
- `publishing.md` Step 6 now has a single reusable gate (`gh run list --commit "$MERGE_SHA"` → `gh run watch --exit-status` → verify `headSha`), and both Option A (`MERGE_SHA=$(git rev-parse HEAD)`) and Option B (merge-commit SHA) use it.
- Both Quick References and the `development.md` release summary updated to the same SHA-filtered gate (no more unfiltered `--limit 1` that could return the previous run).
- Fixed the stale "greps" wording → the extractor matches the exact `## X.X.X` heading.

**[P2] `withLockfile` no longer mishandles a non-directory at the lock path.** Stale-breaking now requires `lockStat.isDirectory()`; a regular file there throws `Lock path exists but is not a directory: …` instead of being renamed aside (which let the critical section run unprotected). Added a regression test that pre-creates a file, asserts the callback doesn't run, and asserts the file is left in place. Reproduced your `staleMs: -1` case → now rejects with the file intact.

**Suggestions taken:** extract the release body **before** `pnpm -r publish` (fail before shipping if the wrapper breaks), random `GITHUB_OUTPUT` delimiter, and a shell-free CLI smoke test for `scripts/extract-changelog.ts` (runs it via `node --import tsx`). Also aligned the PR body (TypeScript scanner, not "flag-based extraction") and the `publishing.md:168` wording.

Once CI is green on `4e4c147` I'll merge.

---

Issue #139, opened 2026-05-29
Sync: unrelated tbd-sync histories are undetected and unrecoverable (doctor reports healthy, push fails forever)


jlevy:
## Summary

A repo's `tbd-sync` branch can diverge from its remote with **no common ancestor** (two unrelated "Initialize tbd-sync branch" roots). When that happens:

- `tbd sync` push fails permanently ("Remote has conflicting changes"),
- `tbd doctor` reports the remote sync branch as **healthy / not diverged**, and
- there is no recovery path — `unrelated` / `allow-unrelated-histories` appear **zero** times in the codebase.

This is reachable in practice when two environments (e.g. a local machine and a cloud agent) each initialize the sync branch around the same time. I hit it for real, root-caused it against the bundled source of `get-tbd@0.1.30` (running binary `0.1.31-dev.28.3c2e7ca-dirty`), and **validated the proposed rescue end-to-end** (details at the bottom).

## Observed state (real case)

```
$ tbd sync --status
  ↑ 5 commit(s) ahead (to push)
  ↓ 6 commit(s) behind (to pull)

$ git merge-base tbd-sync origin/tbd-sync
>>> NO COMMON ANCESTOR

# two different roots, both titled "Initialize tbd-sync branch":
local  root 4a16878  committed 2026-05-26T07:14:20Z  by metaproc
remote root 7552fb9  committed 2026-05-25T23:03:45Z  by Claude   (~8h older)

# origin/tbd-sync tracking reflog:
4d97019 origin/tbd-sync@{0}: fetch: forced-update   <- remote history replaced
92f7a61 origin/tbd-sync@{1}: update by push
31da31b origin/tbd-sync@{2}: update by push
4a16878 origin/tbd-sync@{3}: update by push
```

`tbd doctor` showed `✓ Remote sync branch`, `✓ Sync status`, `✓ Sync consistency` throughout — i.e. it never flagged the unrecoverable state.

## Root cause

`initWorktree` is careful — it does a **live** existence check before creating an orphan:

```js
if (branchExists(syncBranch))            → attach worktree to local branch
else if (remoteBranchExists(remote, …))  → git fetch + create branch FROM origin/tbd-sync
else                                     → git worktree add --orphan   (fresh root)
```

So this is **not** a naive "never fetches before create" bug. The faults are:

### Fault 1 — the remote-existence check fails open

```js
async function remoteBranchExists(remote, branch, baseDir) {
  try {
[...]

Issue #137, opened 2026-05-28
tbd-sync race: independent inits in different environments create unrelated histories, sync cannot recover


jlevy:
## Summary

If two environments run `tbd setup --auto` independently before either has pushed `tbd-sync`, both create their own orphan `tbd-sync` branch with a fresh "Initialize tbd-sync branch" commit. The two branches have **identical trees but unrelated commit histories**, so subsequent pushes from the second environment fail with "Push failed: Push failed after 3 attempts. Remote has conflicting changes." `tbd sync` retries the same push three times and falls through to the outbox — it has no `--allow-unrelated-histories` path, so the divergence cannot be auto-recovered.

Hit on tbd v0.1.29 and v0.1.30.

## Reproducer (real-world)

1. Environment A starts (say at `T`). `tbd setup --auto` creates a local `tbd-sync` orphan branch and does **not** push it yet. At this point `origin/tbd-sync` does not exist.
2. Environment B starts at `T+~8h`. Same `tbd setup --auto` runs. Its `remoteBranchExists(origin, tbd-sync)` check returns `false`, so it also creates an orphan local `tbd-sync` and pushes first to origin.
3. Environment A later runs `tbd sync` → push fails. Three retries against the same commit don't help. Outbox fallback kicks in.

Both "Initialize tbd-sync branch" commits had **identical file trees** (same `.gitkeep`, `meta.yml`, `.gitattributes` blobs); only the commit metadata (author/date/parent) differed. The histories are forever unrelated:

```
$ git merge-base tbd-sync origin/tbd-sync; echo "exit: $?"
exit: 1
```

## Root cause

`initWorktree` in `dist/cli.mjs:1479–1525` (tbd v0.1.30) decides:

1. If local `tbd-sync` exists → use it.
2. Else if `origin/tbd-sync` exists (via `git ls-remote --exit-code`) → fetch and track it.
3. Else → `git worktree add --orphan -b tbd-sync …` and commit a fresh root.

The fresh-orphan path is never re-checked. As soon as origin/tbd-sync is created by another actor, the local stays on its own root forever, and `tbd sync` has no recovery for unrelated histories (`cli.mjs:1314` just returns "Push failed after \${MAX_PUSH_RETRIES} attempts. Remote has conflicting changes.").

`tbd doctor` in v0.1.30 does detect the state (`⚠ Sync consistency - diverged (N ahead, M behind)`) but its only advice is `Run: tbd sync to reconcile`, which can't.

## Recommendations

[...]

Issue #135, opened 2026-05-27
Silent fallback when sync worktree is missing: list shows 'No issues found', create writes to wrong (gitignored) location


jlevy:
## Summary

When the `tbd-sync` git worktree is missing/unhealthy (e.g. a fresh clone in an ephemeral container), `tbd` operations **degrade silently** instead of failing loudly or auto-healing:

- `tbd list` reports **"No issues found"** even though the repo has many issues on the `tbd-sync` branch — making the tracker look empty.
- `tbd create` **succeeds with `✓ Created rpp-XXXX`** but writes the issue files to a non-canonical location (`.tbd/data-sync/issues/` at the repo root, which is gitignored) instead of the worktree.
- A subsequent `tbd sync` then reports `✓ Already in sync` and the just-created issues do **not** appear in `tbd list` — so they look lost.

`tbd doctor` does correctly diagnose and fix this (`N issue(s) in wrong location [fixable]` → `tbd doctor --fix` migrates them), but by then an agent or user has already been given two misleading signals in a row: *"the tracker is empty"* and *"your new issues were saved."*

Observed on **tbd v0.1.30** (latest at time of writing).

## Environment

- Fresh clone in an ephemeral container (Claude Code on the web / remote execution environment).
- SessionStart hook installs tbd and runs `tbd prime`, but the `tbd-sync` worktree was not materialized, so `.tbd/data-sync-worktree/` was missing at session start.

## Reproduction (observed sequence)

1. Start a session in a fresh clone of a repo that already has issues on `tbd-sync`. `tbd status` shows:
   ```
   Worktree: .../.tbd/data-sync-worktree (unhealthy)
     Run: tbd doctor --fix
   ```
2. `tbd list` → **`No issues found`** (despite ~138 issues existing on `tbd-sync`).
3. Create issues, e.g. an epic plus children:
   ```
   tbd create "..." -t epic ...            # ✓ Created rpp-3u8q
   tbd create "..." --no-sync --parent ... # ✓ Created rpp-zjdh
   ...
   ```
   All succeed with no warning.
4. `tbd sync` → `✓ Worktree repaired successfully` / `✓ Already in sync`.
5. `tbd list` now shows the pre-existing issues but **not** the newly created ones.
6. `tbd doctor`:
   ```
   ✗ Data location - 13 issue(s) in wrong location (.../.tbd/data-sync/issues) [fixable]
       Found 13 issues in .tbd/data-sync/ (wrong)
[...]

PR #141, opened 2026-05-29
Plan: tbd-sync unrelated-history hardening (#139, #137, #135)

jlevy:
## Summary

Plan spec mapping out fixes for three related `tbd-sync` stability issues. **Planning only — no implementation yet**; this PR carries the spec and the tracking beads.

- **#139** — Unrelated `tbd-sync` histories (no common ancestor) are undetected (`doctor` reports healthy) and unrecoverable (push fails forever).
- **#137** — The init race that creates those unrelated histories; nothing pushes the fresh orphan to close the window.
- **#135** — Silent fallback when the sync worktree is missing (already largely fixed by the v0.2.0 shared-common-dir refactor; treated here as verification + regression coverage).

## Approach (settled with maintainer)

Three axes — **detection**, **prevention**, **rescue** — in two shippable phases:

- **Phase 1** — `remoteBranchExists` tri-state (absent vs check-failed, no fail-open), `checkRemoteBranchHealth.unrelated` flag, `doctor`/`sync` surfacing, best-effort push of the fresh orphan on init, and a #135 regression test.
- **Phase 2** — `rescueUnrelatedHistory`: non-destructive, issue-file-layer reconciliation wired into `tbd doctor --fix` (safety backup branch → adopt remote base → replay local-only issues + union `ids.yml` → fast-forward push).

Confirmed decisions: auto-fix non-destructive rescue; #135 = verify + regression tests; yes to best-effort orphan push on init.

## Tracking

Spec: `docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md`
Beads: epic `tbd-55sk` + 10 children (linked via `--spec`, with dependency ordering).

https://claude.ai/code/session_016Sc2W6J85NmVdbJCjpVbC9

---
_Generated by [Claude Code](https://claude.ai/code/session_016Sc2W6J85NmVdbJCjpVbC9)_

### DIFF
--- a/docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md
+++ b/docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md
@@ -58,14 +58,14 @@ recovery) — and locks each behavior in with tests.
 
 ### Current control flow (verified against current source, v0.2.0)
 
-**`initWorktree`** (`src/file/git.ts:1119`) decision tree:
+**`initWorktree`** (`packages/tbd/src/file/git.ts:1119`) decision tree:
 
 1. local `tbd-sync` exists → attach worktree to it.
 2. else `remoteBranchExists(remote, syncBranch)` → fetch + create tracking branch.
 3. else → `git worktree add --orphan` + seed structure + commit
    `"Initialize tbd-sync branch"`. **No push.**
 
-**`remoteBranchExists`** (`src/file/git.ts:708`):
+**`remoteBranchExists`** (`packages/tbd/src/file/git.ts:708`):
 
 ```ts
 try {
@@ -80,7 +80,7 @@ try {
 reachable, and exits with other non-zero codes (or throws) on auth/network/transient
 failure. Today both collapse to `false`.
 
-**`checkRemoteBranchHealth`** (`src/file/git.ts:1324`):
+**`checkRemoteBranchHealth`** (`packages/tbd/src/file/git.ts:1324`):
 
 ```ts
 try {
@@ -96,20 +96,21 @@ try {
 That lands in the catch and is reported as `diverged = false` — the worst case looks
 healthy.
 
-**`pushWithRetry`** (`src/file/git.ts:634`) treats `fetch first` / `rejected` /
-`non-fast-forward` as retryable, fetches + runs the field-level merge callback, and
-after `MAX_PUSH_RETRIES` (3) returns `"Remote has conflicting changes."` For unrelated
-histories the merge callback can’t help (the git merge has no base), so it burns three
-identical attempts and falls through to the outbox.
+**`pushWithRetry`** (`packages/tbd/src/file/git.ts:634`) treats `fetch first` /
+`rejected` / `non-fast-forward` as retryable, fetches + runs the field-level merge
+callback, and after `MAX_PUSH_RETRIES` (3) returns `"Remote has conflicting changes."`
+For unrelated histories the merge callback can’t help (the git merge has no base), so it
+burns three identical attempts and falls through to the outbox.
 
 **Recovery machinery that already exists** and we build on:
 
-- Safety-backup-branch idiom (`src/file/git.ts:1062`):
+- Safety-backup-branch idiom (`packages/tbd/src/file/git.ts:1062`):
   `git branch tbd-legacy-preserve-<ts> <head>` before a destructive op.
 - `repairWorktree` / corrupted-worktree backup to `sharedBackupsDir`
-  (`src/file/git.ts:1513+`).
+  (`packages/tbd/src/file/git.ts:1513+`).
 - `reconcileMappings`, `mergeIdMappings`, `resolveIdMappingConflicts`
-  (`src/file/id-mapping.ts`) and the field-level `mergeIssues` (`src/file/git.ts`).
+  (`packages/tbd/src/file/id-mapping.ts`) and the field-level `mergeIssues`
+  (`packages/tbd/src/file/git.ts`).
 - `attic/conflicts/` routing for issue-level conflicts.
 
 ### Why a file-layer rescue is safe (storage model)
@@ -124,7 +125,7 @@ merge must happen at the issue-file layer, not the git-history layer.
 
 ### #135 status (verified)
 
-`assembleDataContext` (`src/cli/lib/data-context.ts:172`) resolves with
+`assembleDataContext` (`packages/tbd/src/cli/lib/data-context.ts:172`) resolves with
 `allowFallback: false`, so a missing worktree throws `WorktreeMissingError` instead of
 silently using the gitignored `.tbd/data-sync/`. `withDataSyncContext` auto-repairs a
 `missing`/`prunable` worktree via `ensureSharedDataSyncLayout` → `repairWorktree`. So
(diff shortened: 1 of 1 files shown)


jlevy (commented):
## Senior engineering review

Summary: the detection / prevention / rescue split is the right architecture for #139/#137/#135, and I agree with the big decision to rescue at the issue-file layer instead of using `--allow-unrelated-histories` or force-pushes. I would tighten the spec before implementation, though: as written, it still leaves one #137 race open and the rescue algorithm can silently drop same-issue local edits.

### Findings

[P1] Handle a rejected first orphan push as a detected race, not an ignored best-effort failure.  
`docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:200` says to push the fresh orphan best-effort and ignore failure. That shrinks the race window, but it does not close the important interleaving: environment A probes `absent` and creates the orphan; environment B creates and pushes its orphan before A's immediate push; A's push is rejected non-fast-forward; if we ignore that rejection, A still returns from setup with a local `tbd-sync` that is unrelated to `origin/tbd-sync`. Since this happens during init before user issue writes, the safe handling is straightforward: on non-fast-forward/fetch-first from the immediate orphan push, fetch, verify the local branch only contains the initial scaffold (or no user issue files), and adopt the remote; otherwise fail loudly with the unrelated-history remediation. Please add a Phase 1 test for exactly this interleaving, not just the happy-path immediate push.

[P1] Rescue must merge same-ULID files, not only replay files absent on remote.  
`docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:238` defines the rescue delta as issue files present locally but absent on `origin/tbd-sync`, then resets to remote and reapplies only those files. That is safe for the real observed “57 remote + 5 local-only” case, but it is not a safe general rescue: if `issues/is-<ulid>.md` exists on both unrelated roots with different content, the local version is dropped by design. ULIDs make independent new-issue collisions unlikely, but same-path divergence is reachable via copied worktrees, restored backups, force-push/replacement history, or a user editing an issue that exists in both stores. The rescue should categorize `local-only`, `remote-only`, `both-identical`, and `both-different`; for `both-different`, run the issue-level merge/conflict-attic path or preserve both versions with an explicit conflict artifact. Add tests for same issue ID with different scalar fields, labels, and a true conflict.

[P2] Detect unrelated histories before/inside the pull/merge stage, not only after push failure.  
`docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:224` places the `tbd sync` handling in the push-failure block. In the current full-sync path, the earlier merge attempt can already hit `fatal: refusing to merge unrelated histories`, and `packages/tbd/src/cli/commands/sync.ts:865` currently catches that together with fetch failures as “may be first sync.” If implementation only adds a better message after `pushWithRetry`, sync will still do misleading work first and can leave diagnostics/debug output pointing at the wrong phase. The spec should require a pre-push health check after fetch, or an explicit unrelated-history branch in the merge catch, before file-level conflict resolution and retry logic run.

[P2] Specify the rescue locking and dirty-state contract.  
`docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:230` adds a mutating `doctor --fix` rescue that fetches, backs up, resets the shared worktree/branch, writes files, commits, and then pushes. Existing `initWorktree` and `repairWorktree` are documented and called under `withSharedDataSyncLock`; the rescue needs the same contract in the spec. It also needs preconditions for a clean data-sync worktree/no merge in progress, and failure semantics after backup but before commit. Without that, a concurrent `tbd create`/`tbd sync` from a sibling worktree can race the reset/replay window.

[P2] The tri-state remote probe section is internally ambiguous.  
`docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:182` says to keep `remoteBranchExists` as a thin boolean wrapper for existing callers including `init`, then `:184` says `initWorktree` should branch on the tri-state. Please make this unambiguous: any path that can create a local orphan must use `probeRemoteBranch` directly and must not collapse `check-failed` to `false`. The boolean wrapper is only acceptable for read-only/status-style callers where fail-closed behavior is not needed.

### Design assessment

The strongest parts of the plan are the tri-state remote probe, making `merge-base` exit status explicit, routing doctor/sync to `doctor --fix`, and avoiding a destructive force-push recovery. Those choices fit the existing storage model and the existing `mergeIssues` / `mergeIdMappings` / `reconcileMappings` machinery.

The main design tightening is to treat rescue as a real reconciliation algorithm, not just a file copy of local-only files. The implementation should be explicit about content buckets, ID mapping conflict policy, locking, clean-worktree preconditions, and post-rescue verification before it claims fast-forward push safety.

### Documentation gaps / cleanup

- The file references in the spec use `src/file/git.ts`, but the real paths are under `packages/tbd/src/...`. Since this PR is a plan/spec PR, accurate paths matter for implementers.
- The rollout says “draft PR” while PR #141 is not marked draft.
- Add the immediate-push rejected-race test to Phase 1 and the same-ID divergent rescue tests to Phase 2.

### Validation performed

- Checked out the PR head branch: `claude/eager-brown-tBntt` against base `main`.
- Verified PR CI via `gh pr checks`: Benchmark, Coverage & Lint, DeepSource Secrets, and Test on macOS/Ubuntu/Windows are all passing.
- Existing PR comments are bot-only (DeepSource + coverage); no unresolved human review comments found via `gh pr view --comments` / PR review comments API.
- Ran a local git repro with two unrelated orphan `tbd-sync` roots. `git merge-base tbd-sync origin/tbd-sync` exited `1`, and current `checkRemoteBranchHealth('origin', 'tbd-sync', repo)` returned `{"exists":true,"diverged":false,...}`, confirming the core #139 diagnosis.
- Ran the package Vitest suite: `59` files, `1092` tests passed.

jlevy:
Thanks — all five findings were sound; incorporated in `cb569e2` (spec-only, no implementation yet).

- **[P1] Rejected first orphan push → detected race.** Fault 1b now classifies the immediate push: success / transient (ignore) / non-fast-forward rejection. On rejection it fetches, and if the local branch is scaffold-only it adopts the remote, otherwise it fails loudly with the `tbd doctor --fix` remediation. Added the A-probes-absent / B-pushes-first / A-rejected interleaving test to Phase 1.
- **[P1] Rescue merges same-ULID files.** Rescue now categorizes by ULID into `local-only` / `remote-only` / `both-identical` / `both-different`; `both-different` runs through `mergeIssues` and routes true conflicts to `attic/conflicts/` (never dropped). Added Phase 2 tests for differing scalars, differing labels, and a true conflict.
- **[P2] Detect before the merge stage.** Added an explicit post-fetch unrelated-history check at `sync.ts:865` (splitting it out of the "may be first sync" catch) so detection happens before file-level conflict/retry; the push-failure-block check stays as defense in depth.
- **[P2] Locking + dirty-state contract.** Rescue now runs under `withSharedDataSyncLock` with clean-worktree / no-merge-in-progress preconditions, and documents post-backup/pre-commit failure semantics (reset is the only destructive step and always follows the backup branch, so it's restartable).
- **[P2] Tri-state ambiguity.** Clarified: orphan-creating paths call `probeRemoteBranch` directly and never collapse `check-failed` to `false`; the boolean `remoteBranchExists` wrapper is for read-only callers only (with a unit test asserting `initWorktree` doesn't route through it).

Docs cleanup: corrected all source paths to `packages/tbd/src/...`, resolved the `pushWithRetry` open question (detect up front, no wasted retries), and the PR is now marked draft.

---

PR #143, opened 2026-05-29
feat: tbd-sync unrelated-history detection, prevention, and rescue (#139, #137, #135)

jlevy:
## Summary

Implements the full plan spec `docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md`, hardening the `tbd-sync` layer along three axes — **detection**, **prevention**, **rescue** — so the branch can no longer silently end up unrecoverable or misreported as healthy.

- **#139** — unrelated `tbd-sync` histories (no common ancestor) were undetected (`doctor` said "healthy") and unrecoverable (push failed forever).
- **#137** — the init race that creates those histories (nothing pushed the fresh orphan to close the window).
- **#135** — silent fallback when the sync worktree is missing (already fixed in v0.2.0; here verified + locked with a regression test).

This is the **implementation** PR. The planning PR (#141, spec only) tracked the same work; this branch carries the spec plus the full implementation. Tracked by epic **`tbd-55sk`** + 11 children (all closed).

## What changed

**Phase 1 — detection + prevention + #135 regression**
- `git()` throws a structured `GitError` carrying `exitCode` (+ `exitCodeOf`), so callers branch on git's exit status instead of string-matching stderr. _(structural, tbd-as47)_
- `probeRemoteBranch` tri-state `present` / `absent` / `check-failed` (keyed on `ls-remote --exit-code` → 2); `remoteBranchExists` becomes a fail-closed wrapper. Fixes Fault 1's fail-open catch. _(tbd-ptj3)_
- `initWorktree` refuses to orphan on `check-failed` (no divergent local branch), still orphans local-only repos, and pushes the fresh orphan immediately (`pushFreshOrphan`) so "first init wins"; on a rejected race it adopts a scaffold-only remote or fails loudly toward `tbd doctor --fix`. _(#137, tbd-xkyh)_
- `checkRemoteBranchHealth` gains `unrelated` — distinguishing "no local branch" from "no common ancestor" via `merge-base` exit 1. _(Fault 2, tbd-nzqt)_
- `tbd doctor` reports unrelated histories as a hard ✗ routed to `--fix` (never `tbd sync`). _(tbd-fg5a)_
- `tbd sync` detects the unrelated state up front (post-fetch) and raises `UnrelatedHistoriesError` before any merge/retry — so `pushWithRetry` never burns attempts against an unrelated remote; push-failure block has defense-in-depth. _(tbd-xsb4)_
- #135 regression tryscript: missing worktree heals and reflects real issues, never writing under the gitignored `.tbd/data-sync/`. _(tbd-b6vi)_

**Phase 2 — non-destructive rescue**
- `rescueUnrelatedHistory`: preconditions (clean worktree, no merge in progress) → fetch → backup branch `tbd-backup-<ts>` → ULID-bucket categorization (`categorizeIssuesByUlid`, pure) → adopt `origin/tbd-sync` base → replay local-only, field-merge both-different (losing version preserved under `attic/conflicts/`), union + reconcile `ids.yml` → commit. Reconciliation is at the issue-file layer (ULIDs), never a git history merge; leaves a clean fast-forward push. _(#139, tbd-6l8r)_
- Wired into `tbd doctor --fix` under `withSharedDataSyncLock`. _(tbd-7eyl)_
- Tests: rescue divergence matrix + dirty/merge-in-progress preconditions _(tbd-e550)_, and an end-to-end tryscript `race → detect → doctor --fix → sync` _(tbd-xxk5)_.

## Test plan

- [x] `pnpm typecheck`, `lint`, `format` clean (pre-commit hooks on every commit)
- [x] `pnpm test` — **1126 vitest tests pass** (68 files; +new git-exit-code, probe-remote, init-orphan, unrelated-health, doctor-unrelated, unrelated-histories-error, rescue-categorize, rescue-unrelated, rescue-divergence)
- [x] `tryscript run 'tests/**/*.tryscript.md'` — **817 golden tests pass** (+new missing-worktree-135 and unrelated-rescue scenarios)
- [x] pre-push hook (build + full test + package-age) passed on push
- [x] Manual end-to-end: `doctor` flags unrelated → `doctor --fix` rescues (backup branch, both sides' issues survive) → `sync` fast-forwards

Developed TDD throughout (failing test per fault), structural and behavioral commits kept separate.

🤖 Generated with [Claude Code](https://claude.ai/code/session_016dQiPcejjsZXjPJgKtL78o)

---
_Generated by [Claude Code](https://claude.ai/code/session_016dQiPcejjsZXjPJgKtL78o)_

### DIFF
--- a/docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md
+++ b/docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md
@@ -58,14 +58,14 @@ recovery) — and locks each behavior in with tests.
 
 ### Current control flow (verified against current source, v0.2.0)
 
-**`initWorktree`** (`src/file/git.ts:1119`) decision tree:
+**`initWorktree`** (`packages/tbd/src/file/git.ts:1119`) decision tree:
 
 1. local `tbd-sync` exists → attach worktree to it.
 2. else `remoteBranchExists(remote, syncBranch)` → fetch + create tracking branch.
 3. else → `git worktree add --orphan` + seed structure + commit
    `"Initialize tbd-sync branch"`. **No push.**
 
-**`remoteBranchExists`** (`src/file/git.ts:708`):
+**`remoteBranchExists`** (`packages/tbd/src/file/git.ts:708`):
 
 ```ts
 try {
@@ -80,7 +80,7 @@ try {
 reachable, and exits with other non-zero codes (or throws) on auth/network/transient
 failure. Today both collapse to `false`.
 
-**`checkRemoteBranchHealth`** (`src/file/git.ts:1324`):
+**`checkRemoteBranchHealth`** (`packages/tbd/src/file/git.ts:1324`):
 
 ```ts
 try {
@@ -96,20 +96,21 @@ try {
 That lands in the catch and is reported as `diverged = false` — the worst case looks
 healthy.
 
-**`pushWithRetry`** (`src/file/git.ts:634`) treats `fetch first` / `rejected` /
-`non-fast-forward` as retryable, fetches + runs the field-level merge callback, and
-after `MAX_PUSH_RETRIES` (3) returns `"Remote has conflicting changes."` For unrelated
-histories the merge callback can’t help (the git merge has no base), so it burns three
-identical attempts and falls through to the outbox.
+**`pushWithRetry`** (`packages/tbd/src/file/git.ts:634`) treats `fetch first` /
+`rejected` / `non-fast-forward` as retryable, fetches + runs the field-level merge
+callback, and after `MAX_PUSH_RETRIES` (3) returns `"Remote has conflicting changes."`
+For unrelated histories the merge callback can’t help (the git merge has no base), so it
+burns three identical attempts and falls through to the outbox.
 
 **Recovery machinery that already exists** and we build on:
 
-- Safety-backup-branch idiom (`src/file/git.ts:1062`):
+- Safety-backup-branch idiom (`packages/tbd/src/file/git.ts:1062`):
   `git branch tbd-legacy-preserve-<ts> <head>` before a destructive op.
 - `repairWorktree` / corrupted-worktree backup to `sharedBackupsDir`
-  (`src/file/git.ts:1513+`).
+  (`packages/tbd/src/file/git.ts:1513+`).
 - `reconcileMappings`, `mergeIdMappings`, `resolveIdMappingConflicts`
-  (`src/file/id-mapping.ts`) and the field-level `mergeIssues` (`src/file/git.ts`).
+  (`packages/tbd/src/file/id-mapping.ts`) and the field-level `mergeIssues`
+  (`packages/tbd/src/file/git.ts`).
 - `attic/conflicts/` routing for issue-level conflicts.
 
 ### Why a file-layer rescue is safe (storage model)
@@ -124,7 +125,7 @@ merge must happen at the issue-file layer, not the git-history layer.
 
 ### #135 status (verified)
 
-`assembleDataContext` (`src/cli/lib/data-context.ts:172`) resolves with
+`assembleDataContext` (`packages/tbd/src/cli/lib/data-context.ts:172`) resolves with
 `allowFallback: false`, so a missing worktree throws `WorktreeMissingError` instead of
 silently using the gitignored `.tbd/data-sync/`. `withDataSyncContext` auto-repairs a
 `missing`/`prunable` worktree via `ensureSharedDataSyncLayout` → `repairWorktree`. So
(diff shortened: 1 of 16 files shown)


jlevy (commented):
## Senior engineering review

Summary: This is a thoughtful hardening pass. The overall architecture is right: carry git exit status structurally, fail closed before orphan creation, detect unrelated histories before retry/merge machinery, and recover at the issue-file layer rather than trying to force a git history merge. The test surface is also much stronger than the previous sync code. I would address the rescue edge cases below before merging, because they undermine the PR's core non-destructive/public-ID-stability guarantees.

### Strengths

- `GitError` + `exitCodeOf` is the right primitive; branching on `ls-remote`/`merge-base` exit codes is a major improvement over stderr matching.
- The `probeRemoteBranch` tri-state and init-time refused orphan path directly target the root race instead of only adding recovery after corruption.
- Detecting unrelated histories in `tbd sync` before push retry avoids misleading conflict work and gives the user the correct remediation.
- The PR adds targeted unit, integration, and tryscript coverage for the reported faults, including CI coverage across macOS, Ubuntu, and Windows.

### Issues

1. **Same-ULID rescue can still drop one side's edits without an attic artifact.**
   `packages/tbd/src/file/git.ts:1853`

   `rescueUnrelatedHistory` routes every same-ID/different-content issue through `mergeIssues(null, local, remoteIssue)`, then only preserves a losing whole issue when `fieldConflicts.length > 0`. The problem is that `mergeIssues(null, ...)` manufactures a synthetic base from the lower-version side when `created_at` matches. With two unrelated roots, that side is not a reliable common ancestor. If both environments independently edit the same issue from the same starting version and both end at the same version, the current code chooses `local` as the synthetic base, treats remote differences as one-sided changes, overwrites the local edits, and returns no field conflicts, so nothing is written to `attic/conflicts/`.

   Suggested fix: for unrelated-history rescue, either find a real issue ancestor if one exists or use a dedicated two-way merge path for `bothDifferent` that preserves the full non-winning version whenever the inputs are substantively different and no trustworthy base exists. Add a regression with same `id`, same `created_at`, same `version`, different substantive fields on both sides, and assert neither side is lost.

2. **Rescue can change existing remote public IDs on short-ID collisions.**
   `packages/tbd/src/file/git.ts:1865`

   The rescue adopts `origin/tbd-sync` as canonical, but then builds the merged ID map with `mergeIdMappings(localMapping, remoteMapping)`. `mergeIdMappings` gives the first argument precedence on short-ID conflicts, so a local-only mapping like `abcd -> localUlid` will override an existing remote mapping `abcd -> remoteUlid`. `reconcileMappings` then gives the remote issue a new generated short ID. That means an issue already present on the shared remote can silently change public ID after rescue, which is the opposite of treating the remote base as canonical.

   Suggested fix: give `remoteMapping` precedence during rescue, then reconcile local-only issues around it, e.g. `mergeIdMappings(remoteMapping, localMapping)` or a rescue-specific resolver that explicitly preserves remote public IDs and regenerates only conflicting local-only mappings. Add a test where local and remote unrelated roots use the same short ID for different ULIDs and assert the remote issue keeps its original short ID.

3. **No-op rescue reports failure after a successful reset/adoption.**
   `packages/tbd/src/file/git.ts:1875`

   After resetting the attached worktree to the remote base, rescue always stages and commits. If the unrelated histories contain no local-only issues, no divergent same-ID issues, and no mapping changes, the branch has already been adopted successfully but `git commit` exits non-zero because there is nothing to commit. `doctor --fix` will then report rescue failure even though the local branch now points at the remote base. This can happen with two empty/scaffold-only orphan roots, and possibly with substantively identical issue sets plus identical mappings.

   Suggested fix: after `git add -A`, check `git status --porcelain`; if clean, skip the commit and return a successful rescue result. Add a scaffold-only/no-op unrelated-history test.

### Documentation gaps

- `docs/project/specs/active/plan-2026-05-29-tbd-sync-unrelated-history-hardening.md:7` still says `Status: Draft`, and the implementation checklist starting around line 343 remains unchecked even though this PR claims full implementation. Either update the spec status/checklist for this PR or add a short implementation-status note explaining what remains intentionally open.

### Existing PR comments

No unresolved human review threads found. Current comments are bot-generated DeepSource and coverage reports.

### CI status

Passing at the time of review: Benchmark, Coverage & Lint, DeepSource Secrets, and tests on macOS, Ubuntu, and Windows.

jlevy:
Thanks for the careful review — all three rescue issues plus the doc gap are addressed in `5b49aac`.

**1. Same-ULID rescue could drop a side without an artifact** — Fixed. The rescue no longer keys preservation on `mergeIssues` reporting a field conflict. For every `bothDifferent` issue it now preserves to `attic/conflicts/` **each** side that is not substantively equal to the merged result, so an edit is never dropped silently even when both roots edited the same field from the same version (the synthetic-base case you described). New regression: same `id` / `created_at` / `version`, different titles on both sides — asserts both titles survive (one merged, one in attic).

**2. Rescue could change existing remote public IDs** — Fixed. The ID-map union is now `mergeIdMappings(remoteMapping, localMapping)` (remote precedence), so an issue already on the adopted canonical remote keeps its short ID; `reconcileMappings` regenerates only the colliding local-only mappings. New regression: local and remote use the same short ID for different ULIDs — asserts the remote ULID keeps `ab12` and the local-only one is regenerated.

**3. No-op rescue reported failure** — Fixed. After `git add -A`, the rescue checks `git status --porcelain` and skips the commit when clean (scaffold-only / identical roots), returning a successful result instead of failing on "nothing to commit". New regression: identical issue + identical `ids.yml` on both roots — asserts rescue succeeds and creates no empty commit (`tbd-sync` == `origin/tbd-sync`).

**Docs** — Spec status is now `Implemented (PR #143, epic tbd-55sk + 12 children closed)`, the implementation checklist is checked off, and an "Implementation notes" subsection records these three refinements.

Full suite green locally (1129 vitest + 817 tryscript) and the pre-push hook passed; CI re-running on the new commit.

🤖 Generated with [Claude Code](https://claude.ai/code/session_016dQiPcejjsZXjPJgKtL78o)

---

PR #157, opened 2026-06-03
Fix #155: structured field-level merge for bead YAML in tbd sync

### LINKED ISSUE #155: tbd sync (v0.2.2): concurrent epic child_order_hints/version edits leave git conflict markers in bead YAML → file silently skipped
## Summary

`tbd sync` (v0.2.2) corrupts a bead file when two sessions concurrently edit the **same epic's `child_order_hints` list** (and its `version` counter) before syncing. The sync delegates reconciliation to a **line-based git merge**, which leaves standard `<<<<<<< / ======= / >>>>>>>` conflict markers inside the bead's YAML. The markers make the file invalid YAML, after which `tbd` **silently skips** it — the epic and all of its child wiring disappear from the index. In one observed path the sync even reported success (`✓ Synced: received N updated`) while leaving the corrupted file behind.

Both conflicting fields are **semantically mergeable** (`child_order_hints` is a set/order-hint list; `version` is a monotonic counter), so this should be a clean auto-merge rather than a textual conflict.

## Environment

- `tbd --version`: **0.2.2**
- Sync mode: `git-common-dir-v1`, sync branch `tbd-sync`, worktree `.git/tbd/data-sync-worktree`
- Multiple agent sessions operating on the same git branch, each running `tbd` and `tbd sync`.

## Reproduction

1. Two sessions (A and B) start from the same epic state `E` (some `version: N`, some `child_order_hints: [...]`).
2. **Session A:** `tbd update <childA> --parent <epic>` (or `tbd create --parent <epic>`) → appends `childA` to `E.child_order_hints`, bumps `version` to `N+1`. Syncs.
3. **Session B (no intervening pull):** `tbd update <childB...> --parent <epic>` → appends different children, bumps `version` (to e.g. `N+8`). 
4. `tbd sync` in one of the sessions → conflict.

## What happens

The epic's YAML file (`.tbd/data-sync/issues/<epic-id>.md`) ends up with literal git conflict markers in two places:

```yaml
priority: 1
<<<<<<< HEAD
version: 68
=======
version: 75
>>>>>>> origin/tbd-sync
spec_path: ...
...
child_order_hints:
  - is-...                 # common prefix
<<<<<<< HEAD
  - is-01kt5dg0...         # session A's appended child
=======
>>>>>>> origin/tbd-sync
  - is-01kt5asj...         # session B's appended children
  - is-01kt5asm...
```

Then every subsequent `tbd` command prints:

```
Skipping invalid issue file: is-<epic>.md: Implicit keys need to be on a single line at line 8, column 1:
```

…and the epic (and its parent/child relationships) silently drop out of the working set.

Re-running `tbd sync` after a manual commit then surfaced it more loudly (good), but **re-introduced a marker and duplicated a child ID** in the list:

```
This is a bug in tbd sync. Please report it and manually resolve conflicts in:
  /.../.git/tbd/data-sync-worktree
```

```yaml
child_order_hints:
  - is-01kt4z4w...
<<<<<<< HEAD
  - is-01kt5dg0...
=======
>>>>>>> origin/tbd-sync
  - is-01kt5asj...
  ...
  - is-01kt5dg0...    # <-- duplicated
```

## Why this is a bug (not user error)

Concurrent multi-session edits to a shared epic are a headline use case ("never lose work across sessions"). The two conflicting fields are not textually conflicting in intent:

- `child_order_hints` — a **set** of child IDs (append-only ordering hints). Correct merge = **union** (dedup), preserving order where possible.
- `version` — a **monotonic counter**. Correct merge = **max** (or recompute).

A line-based git merge cannot know this, so it produces spurious conflicts and writes markers into structured data.

## Impact

- **Data appears lost:** the corrupted epic + its child links vanish from `tbd ready` / `tbd show` until manually repaired.
- **Silent in at least one path:** `✓ Synced: received N updated` was printed while a bead file was left invalid (the `Skipping invalid issue file` warning is easy to miss in normal output).
- Manual repair requires editing files inside `.git/tbd/data-sync-worktree`, resolving markers, completing the git merge, and re-syncing — not obvious to users.

## Suggested fix

1. **Semantic field-level 3-way merge** for bead YAML instead of git's textual merge:
   - list fields (`child_order_hints`, `labels`, `dependencies`) → **set union** (dedup, deterministic order),
   - `version` → **max** (or recompute post-merge),
   - scalar/notes → last-writer-wins or explicit conflict surfaced as a normal `tbd` conflict, never raw git markers.
2. **Fail loudly on import errors:** if any bead file fails to parse during/after sync, do **not** report `received N updated` — exit non-zero and name the offending file(s).
3. **Never write raw conflict markers into a `.md` bead file**; if a true conflict remains, store it out-of-band and present it via a `tbd` command.

## Workaround (what resolved it here)

Manually resolved in the sync worktree — replaced both conflict blocks with `version: 76` and the **union** of children (deduped), `git add` + `git commit` to complete the merge, then `tbd sync` → `✓ Already in sync`.

jlevy:
Fixes #155 — `tbd sync` writing git conflict markers into bead YAML on concurrent epic
`child_order_hints`/`version` edits, after which the file is silently skipped and the
epic disappears.

Spec: `docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md`
(tracked as epic `tbd-902g` + 6 children).

## Root cause (three coupled defects)
- **A — wrong strategy:** `child_order_hints` merged last-writer-wins, but it is the
  append-only parent→child wiring. Concurrent edits silently dropped a child.
- **B — conflict path never merged the remote:** both sync conflict sites read the
  *local* file as the "remote" side and passed `base = null`. With a null/synthetic base,
  `mergeIssues` never reaches its `union` branch — so the strategy fix alone is
  insufficient; a real `git merge-base` ancestor is required.
- **C — silent success over corruption:** marker-laden YAML fails to parse, `listIssues`
  skips it, and the `received N updated` tally (a `git diff` count) still printed success.

## The fix (6 commits)
1. **Union-merge `child_order_hints`** (+ update `tbd-design.md` §3.5).
2. **`mergeBeadAcrossRefs`** — structured three-way merge reading base/ours/theirs from
   git refs (`merge-base` + `git show`), so committed blobs (never a marker-corrupted
   working file) feed the existing `mergeIssues` engine with a real base.
3. **Pull-path rewrite** — enumerate conflicts via `--diff-filter=U`, resolve each via
   the helper (`HEAD`/`MERGE_HEAD`).
4. **Push-retry callback rewrite** — same helper against the branch tip and the
   freshly-fetched remote tip (this path has no conflicted index, so `merge-base` is the
   right primitive).
5. **Fail loudly** — `findInvalidBeads` + an `assertNoCorruptBeads` guard; sync now exits
   non-zero naming any unparseable bead instead of reporting success.
6. **e2e tests** — a real `git merge` conflict resolved to a clean union (integration),
   plus a two-session CLI tryscript that converges to the union with no markers.

Reuses the existing merge engine and attic conflict routing; consistent with the
unrelated-history rescue and workspace/import merges (full audit in the spec).

## Testing
- 1153 vitest tests green; 155 sync/child-order tryscripts green (incl. the new repro).
- The new tryscript was confirmed to **fail** under the old `lww` strategy (1 child
  survives) and **pass** with `union` (both children), so it genuinely guards #155.

## Notes
- The push-retry callback's writes are committed by `pushWithRetry` only in the normal
  flow; fully unifying that path with the pull path is flagged as a follow-up in the
  spec's Open Questions (low-risk; out of scope here).
- Beads (`tbd-902g` + children) are committed on the local `tbd-sync` branch but not
  pushed to `origin/tbd-sync` due to a pre-existing unrelated-history state there
  (`tbd doctor --fix` reconciles it non-destructively).

https://claude.ai/code/session_012pyBedVoUi1LVSpNvpqgfy

### DIFF
--- a/docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md
+++ b/docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md
@@ -1,0 +1,337 @@
+# Feature: Structured field-level merge for bead YAML in `tbd sync` (no conflict markers, no silent loss)
+
+**Date:** 2026-06-03 (last updated 2026-06-03)
+
+**Author:** Joshua Levy (with agent assistance)
+
+**Status:** Draft
+
+## Overview
+
+Issue **#155** reports that `tbd sync` (v0.2.2) corrupts a bead file when two sessions
+concurrently edit the **same epic’s `child_order_hints` list** (and its `version`
+counter) before syncing.
+Sync delegates reconciliation to git’s **line-based** merge, which writes literal
+`<<<<<<< / ======= / >>>>>>>` markers into the bead’s YAML. The markers make the file
+invalid YAML; `tbd` then **silently skips** the file (`Skipping invalid issue file: …`),
+so the epic and all of its child wiring drop out of the working set.
+In at least one path sync still printed `✓ Synced: received N updated` while leaving the
+corrupted file behind.
+
+Both conflicting fields are semantically mergeable — `child_order_hints` is an
+append-only **set** of child IDs and `version` is a **monotonic counter** — so this
+should be a clean auto-merge, never a textual conflict.
+
+The fix is to stop relying on git’s textual merge for bead `.md` files and instead route
+every conflicted bead through the project’s **existing** field-level three-way merge
+engine (`mergeIssues`), fed with a **real common-ancestor base** so set-union actually
+combines both sides.
+Sync must also **fail loudly** when any bead fails to parse, rather than reporting
+success.
+
+## Goals
+
+- Concurrent edits to a shared epic’s `child_order_hints` merge to the **union** of both
+  sides (deduped), and `version` to the **max** — automatically, with no manual repair.
+- A bead `.md` file is **never** left on disk (or committed) containing git conflict
+  markers.
+- If any bead file fails to parse during or after a sync, sync **exits non-zero** and
+  names the offending file(s) — it never prints `received N updated` over a corrupted
+  store.
+- All bead merges in the codebase use one consistent field-level engine and strategy
+  table; no path silently drops a side without an attic artifact.
+
+## Non-Goals
+
+- No change to the on-disk storage format (`f04`) or the `git-common-dir-v1` layout.
+- No new interactive merge UI.
+- No change to the unrelated-history rescue flow
+  (`plan-2026-05-29-tbd-sync-unrelated-history-hardening.md`); that path already
+  preserves every losing side to `attic/conflicts/` and is reused as-is.
+- Not introducing a git custom merge **driver** for issue files (see Open Questions for
+  why the post-merge resolution approach is preferred for tbd’s ephemeral-clone use
+  case).
+- Issue #135 (silent fallback on a missing worktree) is tracked separately and already
+  largely fixed; out of scope here.
+
+## Background
+
+### The three coupled defects (verified against current source, v0.2.x)
+
+The single user-visible symptom is produced by three distinct defects that compound:
+
+**Defect A — `child_order_hints` uses the wrong merge strategy.** `FIELD_STRATEGIES`
+(`packages/tbd/src/file/git.ts:387`) declares `child_order_hints: 'lww'`, and
+`tbd-design.md` §3.5 documents it as “soft ordering, LWW”. But `child_order_hints` is
+the parent→child wiring — an append-only set of internal child IDs.
+LWW silently discards one session’s appended child on every concurrent edit.
+The correct strategy is `union` (dedupe), exactly like `labels` and `dependencies`.
+`version` (`max`) and `updated_at` (`max`) are already correct.
+
+**Defect B — the conflict path never actually merges the remote, and passes no base.**
+Both conflict-resolution call sites build the merge from the wrong inputs:
+
+- Pull/merge path (`packages/tbd/src/cli/commands/sync.ts:763-785`): after a failed
+  `git merge`, for each local issue it does
+  `const remoteContent = await git('show', '…origin…:…/<id>.md')` only to test
+  existence, then reads the **local** file as the “remote” side
+  (`readIssue(this.dataSyncDir, id)`) and calls
+  `mergeIssues(null, localIssue, remoteIssue)`. So local and “remote” are identical and
+  `base` is `null`.
+- Push-retry callback (`packages/tbd/src/cli/commands/sync.ts:531-561`): identical
+  mistake — reads `readIssue(this.dataSyncDir, id)` as the remote and passes
+  `base = null`.
+
+`base = null` matters because of how `mergeIssues` behaves (below): with a
+null/synthetic base, the `union` strategy **never combines both sides**.
+
+**Defect C — corrupted files are silently skipped, and sync still reports success.**
+Once git writes markers into the epic file it is invalid YAML. `listIssues`
+(`packages/tbd/src/file/storage.ts:110-128`) catches the parse error, prints
+`Skipping invalid issue file:` via `console.warn`, and `continue`s. Consequences:
+
+1. The corrupted epic is **excluded** from `localIssues`, so the conflict loop in Defect
+   B never even visits it — the markers persist on disk.
+2. The epic vanishes from `tbd list`/`ready`/`show`.
+3. The “received N updated” tally comes from a `git diff` file count (`sync.ts:643-646`,
+   `parseGitDiff`), not from successfully parsed issues, so sync reports
+   `✓ Synced: received N updated` even though a bead is unreadable.
+
+The existing `<<<<<<<` safety check (`sync.ts:848-864`) does catch markers **if they
+reach `git add`** — that is the “louder” second path the reporter saw — but the primary
+failure leaves the file unstaged and the warning easy to miss.
+
+### Why strategy + base are coupled (the key insight)
+
+`mergeIssues` (`packages/tbd/src/file/git.ts:519`) only runs a field’s strategy when
+**both** sides differ from the base:
+
+```text
+if (localVal == baseVal && remoteVal == baseVal) continue;   // unchanged
+if (localVal == baseVal) merged = remoteVal;                 // only remote changed
+if (remoteVal == baseVal) merged = localVal;                 // only local changed
+// else: both changed -> apply strategy (union/max/lww)
+```
+
+When `base` is `null`, `mergeIssues` synthesizes a base from the **lower-version side**
+(`git.ts:529-533`). That synthetic base then *equals one of the two sides entirely*, so
+for `child_order_hints` one side always satisfies `localVal == baseVal` and the code
+takes the **other** side wholesale — the `union` branch is never reached.
+**Therefore changing the strategy to `union` is necessary but not sufficient: sync must
+also supply a real common-ancestor base so both sides’ additions are present as deltas
+and the union actually combines them.**
+
+With a real base `[x]`, `local=[x,A]`, `remote=[x,B]`: both differ → `union` →
+`[x,A,B]`. Correct.
+
+### Audit of every bead-merge site (for consistency)
+
+The user asked that the plan be consistent with all existing merging.
+Full inventory:
+
+| Site | Base today | Reads remote correctly? | Verdict |
+| --- | --- | --- | --- |
+| `mergeIssues` engine (`git.ts:519`) | — | — | Keep; flip `child_order_hints` → `union`. All callers benefit. |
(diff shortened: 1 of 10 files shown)


jlevy (commented):
## Senior engineering review

Summary: This is a strong direction overall: the PR reuses the existing field-level merge engine, adds a real ref-based merge primitive, and covers the main #155 two-session reproduction. I would hold merge for the first two issues below because they leave important parts of the PR's stated behavior either broken or brittle.

## Strengths

- Reusing `mergeIssues` plus real `merge-base` inputs is the right architecture. It keeps the merge strategy table centralized instead of adding another YAML-specific merge implementation.
- The pull-path regression coverage is valuable: the new temp-repo merge test and tryscript both exercise real git behavior rather than only unit-level merge logic.
- No dependency changes in this PR, so there is no supply-chain/package-age concern to review.

## Design assessment

The core pull-path design is sound: let git identify conflicted files, then replace line-based conflict output for bead files with a structured merge read from committed refs. That is simpler and more portable than a custom merge driver, especially for fresh/ephemeral clones where git config may not be installed.

The weaker part is the push-retry path. The PR routes the callback through `mergeBeadAcrossRefs`, but the surrounding transaction still does not actually integrate the remote branch before retrying the push. That makes the callback look fixed while the branch state remains non-fast-forward.

## Issues

### [P1] Push retry computes merges but never commits or advances the sync branch

References: `packages/tbd/src/cli/commands/sync.ts:589`, `packages/tbd/src/cli/commands/sync.ts:591`, `packages/tbd/src/file/git.ts:801`, `packages/tbd/src/file/git.ts:803`, `packages/tbd/src/file/git.ts:810`

When the initial push is rejected, `pushWithRetry` fetches the remote and calls `onMergeNeeded()`. The new callback writes `result.merged` into the data-sync worktree, but `pushWithRetry` then immediately loops and retries the same push ref. It never stages/commits those writes, never performs a real merge commit or reset/replay, and never advances local `tbd-sync` to include `origin/tbd-sync`. The next `git push refs/heads/tbd-sync:refs/heads/tbd-sync` is still non-fast-forward, so this path fails after the retry limit even if the structured merge was clean.

This matters because the PR explicitly lists push-retry callback rewrite as part of the fix. To make this path correct, the retry path needs to complete the transaction before retrying: integrate the remote ref into the local sync branch, write structured resolutions, save any attic conflicts, stage and commit the result, assert no corrupt beads, then retry the push. Alternatively, route this through the same real `git merge` plus structured resolution flow as the pull path.

Also check the user-facing result handling around conflicts. In `pushChanges`, `result.success === false` with `result.conflicts` currently prints `Push completed with ... conflict(s)`, but the push did not complete.

### [P1] `child_order_hints: 'union'` can throw on valid null/undefined states

References: `packages/tbd/src/file/git.ts:389`, `packages/tbd/src/file/git.ts:648`, `packages/tbd/src/file/git.ts:650`, `packages/tbd/src/lib/schemas.ts:171`, `packages/tbd/src/cli/commands/update.ts:405`, `packages/tbd/src/cli/commands/update.ts:407`, `packages/tbd/src/cli/commands/update.ts:421`

The schema allows `child_order_hints` to be nullable/optional, and the CLI intentionally writes `null` when a user clears order hints via `tbd update --child-order ""`. After this PR changes the field strategy to `union`, the generic union branch still assumes both values are arrays:

```ts
unionArrays(localVal as unknown[], remoteVal as unknown[])
```

A valid concurrent merge such as one side clearing hints (`null`) while the other side appends or reorders children (`string[]`) will enter the "both changed" branch and try to spread `null`/`undefined` in `unionArrays`, throwing instead of resolving. This is especially likely for legacy parents with missing hints or users who use the existing clear operation.

The fix should normalize nullable union fields before merging, or give `child_order_hints` a field-specific merge function. For example: treat `null`/`undefined` as `[]` for union purposes, and decide whether an empty merged result serializes as `null` or `[]`. Add tests for `undefined + array`, `null + array`, and `clear + concurrent append`.

### [P2] `mergeBeadAcrossRefs` treats parse failures as missing files

References: `packages/tbd/src/file/git.ts:1808`, `packages/tbd/src/file/git.ts:1811`, `packages/tbd/src/file/git.ts:1825`, `packages/tbd/src/file/git.ts:1828`, `packages/tbd/src/cli/commands/sync.ts:594`, `packages/tbd/src/cli/commands/sync.ts:595`

The helper catches every error while reading/parsing `theirsRef` and returns `null`, with a comment saying the bead is absent on the other side. It does the same broad catch for the base blob and treats it as no base. Those are different cases: a git pathspec/object absence is a merge input shape; a committed bead that exists but fails `parseIssue` is corruption and should propagate into the fail-loud path.

This is more visible in the push callback because it catches helper errors and only debug-logs "merge skipped". A bad remote bead or unexpected parse failure can therefore be silently treated as absent/skipped in the exact code paths this PR is hardening. Please distinguish missing git objects from parser errors, and do not swallow parser errors in the sync merge callbacks.

## Documentation gaps

- `docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md:7` still says `Status: Draft`, and the implementation checklist at `docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md:250` through `docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md:276` is entirely unchecked even though the PR implements most of it. Update the spec status/checklist before landing, or move it according to the repo's Speculate workflow.
- The spec says `version` should merge to the max at `docs/project/specs/active/plan-2026-06-03-tbd-sync-structured-bead-merge.md:34`, but the implementation/tests bump to `max + 1` for substantive merges (`packages/tbd/tests/merge.test.ts:223`). If `max + 1` is intended, update the spec and PR text so future reviewers do not chase a false mismatch.

## Test gaps

- The new #155 tryscript covers the successful two-session union path well.
- `packages/tbd/tests/sync-fail-loud.test.ts` only proves `findInvalidBeads()` detects a bad file. It does not prove the CLI exits non-zero or suppresses `Synced: received ...` when sync itself leaves or encounters an invalid bead, despite that being a stated requirement in the spec. Add an end-to-end CLI regression for that failure behavior.
- Add coverage for the push-retry path after a remote advances between fetch and push. That is the path most likely to catch the transaction bug above.
- Add nullable `child_order_hints` merge coverage as described in the P1 issue.

## Unresolved comments

No unresolved human review comments found. Existing PR comments are automated DeepSource and coverage reports.

## CI status

Current PR checks are passing: Benchmark, Coverage & Lint, DeepSource Secrets, and tests on macOS, Ubuntu, and Windows. I also ran `git diff --check origin/main...HEAD` in a detached PR worktree; it passed.

jlevy:
Thanks for the thorough review — all three issues, both doc gaps, and the test gaps are addressed. Each was reproduced first (TDD red), then fixed.

**[P1] Push-retry never committed/advanced the sync branch** — `86cdff1`
Extracted the pull path's real `git merge` + structured resolution into one shared `mergeRemoteIntoSyncBranch()` and routed both paths through it, so a rejected push now *integrates and commits* the remote into local `tbd-sync` before retrying (no more looping on the same non-fast-forward commit). `pushWithRetry` now keeps retrying after integrating — field conflicts are attic-preserved, not a stop signal — and the misleading "Push completed with … conflict(s)" message is replaced with an accurate one (it only prints on an actual successful push). New e2e `cli-sync-push-retry-155.tryscript.md` drives `tbd sync --push` against an advanced remote and asserts the push actually lands (origin receives the integrated result).

**[P1] `child_order_hints: 'union'` threw on null/undefined** — `49483de`
The union branch now coerces non-array inputs to `[]` (union ignores deletions, so a concurrent clear-vs-append resolves to the append). Tests added for `null + array` and `undefined + array`.

**[P2] `mergeBeadAcrossRefs` treated parse failures as missing files** — `f9b8964`
Separated the git read from the parse: only a missing git object → "absent" (keep ours); a committed-but-unparseable bead now propagates, and the sync merge loops fail loudly (`SyncError`) instead of debug-skipping. Test added for a corrupt committed bead.

**Docs** — `c9c8042`
Spec marked Implemented, Phase 1/2 boxes checked, a Phase 3 section added for this review round, and the `version` wording reconciled to "max, then +1 on a substantive merge" (the existing `mergeIssues` contract). The unify-the-paths open question is now resolved (done, not deferred).

**Test gaps**
- Fail-loud CLI regression — `3e0a8a1` (`cli-sync-fail-loud-155.tryscript.md`): a sync that pulls in an invalid bead exits non-zero, names the file, and prints no success summary.
- Push-retry-after-remote-advances coverage — `86cdff1` (the new tryscript above).
- Nullable `child_order_hints` merge coverage — `49483de`.

Full suite green locally: 1156 vitest tests + 143 sync tryscripts. The push-retry and fail-loud tryscripts both exercise the real binary end-to-end (and the union tryscript was verified to fail under the old `lww` strategy, so it genuinely guards the regression).

---

PR #160, opened 2026-06-03
Fix #158: unrelated-history rescue tolerates a dirty sync worktree

jlevy:
Fixes #158 — a transiently dirty internal sync worktree made `tbd doctor --fix` refuse the unrelated-history rescue, and `tbd doctor` bounced the user between `tbd sync` and `tbd doctor --fix` in a loop.

Spec: `docs/project/specs/active/plan-2026-06-03-unrelated-rescue-dirty-worktree.md` (epic `tbd-lmqu` + 5 children).

## Root cause
The rescue (`rescueUnrelatedHistory`) already snapshots the working tree before its `reset --hard` and replays local work, and it holds the shared lock — so a dirty worktree (tbd's own uncommitted data-sync writes) was already safe to proceed through. The refusal was over-cautious, and its "commit or stash them" message was misleading (the user never touches the internal worktree). Separately, doctor's `checkSyncConsistency` flagged unrelated histories as "diverged → Run: tbd sync", which loops against the unrelated-history error that says "Run tbd doctor --fix".

## Changes
1. **Rescue tolerates a dirty worktree** (`740999f`): commit the pending data-sync state first (so the `tbd-backup-*` branch captures it faithfully and the reset is safe), then proceed. Still refuse a genuine **merge-in-progress**, now with a terminating instruction. Foreign-path guard uses a git exclude pathspec (robust).
2. **Break the advice loop** (`71cc37b`): `divergenceFinding()` routes unrelated divergence to `tbd doctor --fix`, never `tbd sync`.
3. **Exit-code audit** (`1a80072`): a push failure now exits non-zero **unless** the changes were durably saved to the outbox (the permanent soft-fail recovery path keeps exit 0). Transient/unknown failures that saved nothing previously exited 0, hiding hard failures from CI/scripts. Local commits still remain on `tbd-sync`.
4. **Tests**: dirty-worktree + merge-in-progress rescue (`rescue-dirty-worktree.test.ts`); unrelated-aware `divergenceFinding`; e2e tryscripts for the one-shot dirty rescue, the non-zero exit on unrelated histories, and the non-zero exit on an unsaved push failure; updated the missing-remote goldens.

## Notes
- The report's secondary "sync exits 0 on unrelated histories" was **already fixed on `main`** (up-front `UnrelatedHistoriesError`, exit 1) — verified and locked in with a test.
- Behavior change: full-sync push failures without a durable outbox save now exit non-zero (goldens updated accordingly).

## Testing
[...]

PR #163, opened 2026-06-04
Fix brittle ids.yml count assertion (quoted numeric short IDs)

jlevy:
Root-causes and fixes the intermittent `cli-workspace-save.tryscript.md` failure that surfaced on #160 and #162. **Not a flake — a brittle test assertion**, debugged systematically.

## Root cause
The two "mappings contain exactly N entries" assertions counted lines with `grep -c '^[a-z0-9]'` over `ids.yml`. Each mapping line is `shortid: ulid` — but when a short ID is **all-numeric** (e.g. `2253`), the YAML serializer writes the key **quoted** (`"2253": 01kt9j10…`). That line starts with `"`, so the key-anchored pattern misses it and counts N−1.

Reproduced deterministically (iter 36 of 60) — the workspace `ids.yml` had **all 3** entries, one of them `"2253": …`, and the old `grep` returned 2:
```
04f7: 01kt9j10nmwxc0x2hqxb9fhx95
"2253": 01kt9j103f384mbc62dznda8hg
2awc: 01kt9j10cgkzdf108ge1j40mgz
```

`generateShortId` draws 4 chars from `[0-9a-z]` via crypto `randomBytes`, so an all-numeric short ID occurs at ≈ (10/36)⁴ ≈ 0.6% per ID → ≈ **1.8%** across the 3 issues in a run. That matches the observed ~1-in-60 rate exactly.

**The product is correct end-to-end** — `save` copies all three mappings, and `loadIdMapping` round-trips the quoted key (it uses a real YAML parser). Only the test assertion was wrong.

## Fix
Count by the **ULID value** (`grep -cE ': [0-9a-z]{26}$'`), which is immune to key quoting, for both the `my-backup` and `outbox` assertions, with an explanatory note.

## Validation
- 100 create→save runs with the new pattern: **0 failures**; 3 of those runs hit a quoted all-numeric short ID (which the old pattern would have failed).
- `cli-workspace-save.tryscript.md`: 36 passed × 3 runs.

This removes the recurring CI noise across PRs (no product change needed).

https://claude.ai/code/session_012pyBedVoUi1LVSpNvpqgfy

---
_Generated by [Claude Code](https://claude.ai/code/session_012pyBedVoUi1LVSpNvpqgfy)_

### DIFF
--- a/packages/tbd/tests/cli-workspace-save.tryscript.md
+++ b/packages/tbd/tests/cli-workspace-save.tryscript.md
@@ -96,8 +96,12 @@ $ ls .tbd/workspaces/my-backup/issues/ | wc -l | tr -d ' '
 
 # Test: Workspace mappings contain exactly 3 entries
 
+Count by the ULID value, not the key: an all-numeric short ID (e.g. `2253`) is
+YAML-serialized as a quoted key (`"2253":`), so a `^[a-z0-9]` key match would undercount
+it (~1.8% of runs). The mapping itself is correct either way.
+
 ```console
-$ grep -c '^[a-z0-9]' .tbd/workspaces/my-backup/mappings/ids.yml
+$ grep -cE ': [0-9a-z]{26}$' .tbd/workspaces/my-backup/mappings/ids.yml
 3
 ? 0
 ```
@@ -129,8 +133,11 @@ $ ls .tbd/workspaces/outbox/issues/ | wc -l | tr -d ' '
 
 # Test: Outbox mappings contain exactly 3 entries (not all mappings)
 
+Count by the ULID value (see note above): a quoted all-numeric short-ID key must still
+be counted.
+
 ```console
-$ grep -c '^[a-z0-9]' .tbd/workspaces/outbox/mappings/ids.yml
+$ grep -cE ': [0-9a-z]{26}$' .tbd/workspaces/outbox/mappings/ids.yml
 3
 ? 0
 ```


jlevy:
## Senior Engineering Review

### Summary
No blocking issues found. This is a narrow test-only fix: it changes the brittle `ids.yml` mapping-count assertion to count ULID values instead of unquoted key lines, which matches the product behavior because the actual loader parses YAML.

### Design assessment
The approach is appropriate for the failure mode. Counting `: [0-9a-z]{26}$` keeps the assertion focused on the persisted mapping entries and avoids coupling the test to YAML key quoting. I considered whether this should instead parse YAML in the tryscript, but for this smoke-level assertion the value-pattern check is simpler and still covers the intended behavior without adding test complexity.

### Strengths
- The PR correctly treats this as a test assertion bug rather than changing production serialization or ID loading.
- The explanatory note at `packages/tbd/tests/cli-workspace-save.tryscript.md:99` documents why key-anchored counting is unsafe.
- The same fix is applied consistently to both the named workspace and outbox assertions.

### Issues
No code or test issues found.

### Documentation gaps
None. This does not change user-facing behavior or architecture docs.

### Unresolved comments
No unresolved review threads. Existing comments are automated DeepSource and coverage reports only.

### Validation
- Local: `pnpm --filter get-tbd build` passed.
- Local: `pnpm --filter get-tbd exec tryscript run tests/cli-workspace-save.tryscript.md` passed, 36 assertions.
- CI: all reported checks pass: Benchmark, Coverage & Lint, DeepSource: Secrets, and Test on ubuntu-latest, macos-latest, and windows-latest.

PR #165, opened 2026-06-08
feat: Detect unwritable shared data-sync lock in doctor (#164)

### LINKED ISSUE #164: doctor should detect shared data-sync lock unwritable in Codex/agent sandboxes
## Summary

`tbd` v0.2.2 correctly stores its sync worktree under the Git common directory
and uses a mkdir-based shared lock at `$GIT_COMMON_DIR/tbd/locks/data-sync.lock`.
In a Codex app worktree, that path can be outside the thread's writable sandbox
even when the checkout itself is writable. The result is that read-only
diagnostics look healthy, but write-side commands fail immediately with `EPERM`
when trying to create the shared lock directory.

This seems likely to affect Codex and potentially other coding-agent
environments that use Git worktrees where the visible checkout lives under an
agent-managed workspace but the Git common dir lives in the user's original repo
directory.

## Observed environment

- `tbd --version`: `0.2.2`
- current checkout: `/Users/levy/.codex/worktrees/c58d/trading`
- `git rev-parse --git-common-dir`: `/Users/levy/wrk/aisw/trading/.git`
- `tbd status --json` reports:
  - `worktree_path`: `/Users/levy/wrk/aisw/trading/.git/tbd/data-sync-worktree`
  - `worktree_healthy`: `true`
  - `worktree_status`: `valid`

The installed v0.2.2 source matches the intended design:

- `buildSharedTbdPaths()` builds `$GIT_COMMON_DIR/tbd`,
  `$GIT_COMMON_DIR/tbd/data-sync-worktree`, and
  `$GIT_COMMON_DIR/tbd/locks/data-sync.lock`.
- `withSharedDataSyncLock()` wraps shared data-sync operations and calls
  `withLockfile(paths.sharedLockPath, ...)`.
- `withLockfile()` uses atomic `mkdir(lockPath)` as a directory lock, not
  `flock`.

## Reproduction / evidence

In the same Codex desktop thread:

```sh
tbd status --json
```

succeeds and reports the shared worktree healthy.

```sh
tbd doctor --json
```

also reports these checks as OK:

```json
{
  "name": "Worktree",
  "status": "ok",
  "path": "/Users/levy/wrk/aisw/trading/.git/tbd/data-sync-worktree"
}
```

and:

```json
{
  "name": "Common-dir layout",
  "status": "ok",
  "path": "/Users/levy/wrk/aisw/trading/.git/tbd/layout.yml"
}
```

But the current sandbox cannot create the lock directory:

```sh
mkdir /Users/levy/wrk/aisw/trading/.git/tbd/locks/codex-sandbox-final-probe.lock
```

fails with:

```text
mkdir: /Users/levy/wrk/aisw/trading/.git/tbd/locks/codex-sandbox-final-probe.lock: Operation not permitted
```

A write-side `tbd` command fails for the same reason:

```text
Failed to create issue: EPERM: operation not permitted, mkdir '/Users/levy/wrk/aisw/trading/.git/tbd/locks/data-sync.lock'
```

This is not a stale-lock or lock-contention case. If another process held the
lock, the `mkdir` would return `EEXIST` and enter the wait/stale-break path. Here
it returns `EPERM`, so the write fails before the critical section begins.

## Why Codex can hit this

The Codex manual documents that local commands run in an OS-enforced sandbox. In
`workspace-write`, Codex can write the active workspace and configured writable
roots, but not arbitrary sibling paths. It also documents
`sandbox_workspace_write.writable_roots` and newer permission profiles as ways to
extend the writable boundary without disabling sandboxing entirely.

Codex app worktrees are created under `$CODEX_HOME/worktrees`, while Git
worktrees share the original repository's Git metadata. That means a
Codex-managed checkout can be writable while `$GIT_COMMON_DIR/tbd` lives
elsewhere and is read-only from the sandbox's perspective.

This may also affect other agents with similar sandbox models:

- agent checkout/worktree is writable;
- Git common dir is outside that workspace;
- shared tbd sync state and locks live under the common dir;
- the agent can read the common dir but cannot write the lock path.

## Expected behavior

`tbd doctor` should detect that the current process cannot acquire or create the
shared data-sync lock, and should report a targeted remediation before users
discover it through a write command failure.

Write-side commands should also turn `EPERM`/`EACCES` on
`$GIT_COMMON_DIR/tbd/locks/data-sync.lock` into a clearer error message.

## Proposed changes

1. Add a `doctor` health check, e.g. `Shared data-sync lock writability`:

   - resolve `paths = resolveSharedTbdPaths(cwd)`;
   - ensure/probe `paths.sharedLocksDir` if appropriate;
   - attempt
     `mkdir(join(paths.sharedLocksDir, ".tbd-doctor-probe-<pid>-<random>.lock"))`;
   - immediately `rmdir` the probe directory;
   - on `EPERM`/`EACCES`, warn/error that the current process cannot create
     `$GIT_COMMON_DIR/tbd/locks/data-sync.lock`;
   - keep this as a direct filesystem capability probe rather than relying only
     on environment variables, because this Codex desktop shell did not expose a
     `CODEX_*` variable.

2. Improve write-side lock errors:

   When `withSharedDataSyncLock()` fails with `EPERM`/`EACCES` on the shared lock
   path, include the resolved Git common dir and a short explanation: the
   checkout may be writable while the shared tbd state under `$GIT_COMMON_DIR/tbd`
   is outside the agent sandbox.

3. Add Codex-specific guidance when Codex integration is detected:

   - suggest adding the shared tbd dir to Codex writable roots or a named
     permission profile for future sessions;
   - mention that project-local `.codex/config.toml` is loaded only for trusted
     projects;
   - warn that absolute paths under `.git` are machine-local and should probably
     not be committed blindly;
   - suggest sandbox escalation as the immediate workaround for the current
     session.

4. Consider a setup-time note or opt-in setup helper:

   `tbd setup --codex` could print a warning if the repo is a linked worktree and
   `$GIT_COMMON_DIR/tbd` is outside the project root. A more aggressive option
   would be an explicit opt-in command that edits user-level Codex config, but
   that should probably require confirmation because it is machine-local and
   security-sensitive.

## Possible Codex remediation text

For a future Codex session, the user may need a config entry along these lines,
adjusted for their machine and current Codex permission model:

```toml
sandbox_mode = "workspace-write"

[sandbox_workspace_write]
writable_roots = ["/Users/levy/wrk/aisw/trading/.git/tbd"]
```

or a named permission profile that extends `:workspace` and adds the same path as
an additional workspace root. This should be validated because Codex also
protects `.git` paths in default workspace-write mode, so the robust guidance may
need to be profile-based or otherwise explicit about protected Git metadata
paths.

## Current workaround

Run write-side `tbd` commands with sandbox escalation, or start Codex with a
configuration that grants write access to `$GIT_COMMON_DIR/tbd`. `tbd status` and
most read-only checks can still pass without that access, which is why a doctor
check is important.

jlevy:
## Summary

In an agent sandbox (e.g. Codex worktrees), the checkout can be writable while
`$GIT_COMMON_DIR/tbd` — which holds the shared sync worktree and the data-sync
lock — lives outside the sandbox's writable boundary. Read-only commands and
`tbd doctor` reported healthy, but the first write died with a raw
`EPERM: … mkdir '…/locks/data-sync.lock'`. This adds a `doctor` check that
surfaces the condition, makes the write-side failure a clear fatal error, and
hardens `doctor` so one failing check never aborts the whole report.

Fixes #164 (items 1–3; the security-sensitive `tbd setup --codex` config-writer in
item 4 is intentionally deferred — see Notes).

## Changes

- **`doctor` — new "Shared lock writability" check** (`doctor.ts`): probes the
  shared locks dir by creating and removing a uniquely named probe directory,
  mirroring `withSharedDataSyncLock`. An unwritable lock (`EPERM`/`EACCES`) is a
  hard **error** finding (sets exit 1) because every write command must take this
  lock. Remediation text is sandbox-aware (writable roots / sandbox escalation)
  when the Git common dir lives outside the checkout, and generic
  (filesystem-permissions) otherwise. The probe is self-contained and never throws.
- **`doctor` never aborts on one check** (`doctor.ts`): every health and
  integration check now runs under a `safeCheck` guard, so an unexpected throw
  becomes an error *finding* and the rest of the report still runs — doctor must
  list every issue, not just the first.
- **Clear fatal write error** (`common-dir-layout.ts`): `withSharedDataSyncLock`
  translates `EPERM`/`EACCES` on the shared lock path into a
  `SharedLockUnwritableError` that names the lock path, the shared tbd dir, and
  the remediation. The match is path-scoped so an unrelated `EPERM` thrown by the
  critical section (e.g. writing issue data) passes through unchanged.
- **Tests**: pure-function unit tests for the finding classifier, the
  outside-checkout detector, and the error message
  (`doctor-lock-writability.test.ts`); an e2e happy-path test asserting a normal
  repo reports the lock writable and `doctor` still lists the full check set
  (`doctor-lock-writability-e2e.test.ts`).

## Test Plan

- [x] Unit tests pass — `doctor-lock-writability.test.ts` (10 tests)
- [x] E2e happy-path test passes — `doctor-lock-writability-e2e.test.ts`
- [x] Full suite green (`pnpm test`) — 74 files / 1172 tests
- [x] Build succeeds (`pnpm build`), typecheck + lint + prettier clean
- [x] **Live `EPERM` reproduction** (`chattr +i` on the locks dir, since CI runs
      as root and can't reproduce permission denial via `chmod`):
  - Write command (`tbd create`) fails with the new
    `SharedLockUnwritableError` message and exit 1 — not a raw `mkdir` error.
  - `tbd doctor` flags "Shared lock writability" as `error` (exit 1) while still
    listing all 19 health checks (no abort).
- Edge cases considered:
  - Common dir **inside** the checkout → generic permission guidance (no sandbox
    framing); common dir **outside** → sandbox/writable-roots guidance.
  - Unrelated `EPERM` from the critical section is not misclassified as a lock
    problem (path-scoped match).
  - Steady-state read commands still take no lock, so they keep working in a
    sandbox where the worktree/layout are already valid (matches #164).

### Reviewer note on reproduction

The sandbox EPERM path is not reproducible with `chmod` under CI's root, so the
classification logic is exercised as pure functions; the happy path is covered
e2e; and the wired EPERM behavior was verified manually with `chattr +i`.

## Related Beads

None. The deferred `tbd setup --codex` writable-roots helper (issue #164 item 4,
marked "Consider" in the issue) is intentionally out of scope here — it edits
machine-local, security-sensitive Codex config and warrants its own change with a
confirmation UX. Happy to file a bead for it if desired.

https://claude.ai/code/session_01HQnFgn8byLa88Sw2FuUHxV

---
_Generated by [Claude Code](https://claude.ai/code/session_01HQnFgn8byLa88Sw2FuUHxV)_

### DIFF
--- a/packages/tbd/src/cli/commands/doctor.ts
+++ b/packages/tbd/src/cli/commands/doctor.ts
@@ -8,7 +8,7 @@
 
 import { Command } from 'commander';
 import { access, mkdir, readdir, readFile, rmdir, unlink } from 'node:fs/promises';
-import { isAbsolute, join, relative, resolve, sep } from 'node:path';
+import { join } from 'node:path';
 import { randomUUID } from 'node:crypto';
 
 import { BaseCommand } from '../lib/base-command.js';
@@ -17,7 +17,12 @@ import { listIssues, type InvalidIssueFile } from '../../file/storage.js';
 import { IncompatibleFormatError, readConfig } from '../../file/config.js';
 import { prepareDataSyncContext } from '../lib/data-context.js';
 import type { Config, Issue, IssueStatusType } from '../../lib/types.js';
-import { resolveSharedTbdPaths, TBD_DIR, DATA_SYNC_DIR } from '../../lib/paths.js';
+import {
+  isCommonDirOutsideProject,
+  resolveSharedTbdPaths,
+  TBD_DIR,
+  DATA_SYNC_DIR,
+} from '../../lib/paths.js';
 import { detectDuplicateYamlKeys } from '../../utils/yaml-utils.js';
 import {
   getClaudePaths,
@@ -120,18 +125,6 @@ export function divergenceFinding(
     message: `diverged (${ahead} ahead, ${behind} behind)`,
     suggestion: 'Run: tbd sync to reconcile',
   };
-}
-
-/**
- * True when the Git common dir lives outside the project checkout — the linked
- * worktree shape that lets a checkout be writable while `$GIT_COMMON_DIR/tbd`
- * (shared sync state + lock) is not. This is the generic signal behind the
- * Codex-sandbox case in #164; it does not depend on any `CODEX_*` env var, which
- * that sandbox did not expose.
- */
-export function isCommonDirOutsideProject(gitCommonDir: string, projectRoot: string): boolean {
-  const rel = relative(resolve(projectRoot), resolve(gitCommonDir));
-  return rel === '..' || rel.startsWith(`..${sep}`) || isAbsolute(rel);
 }
 
 /**
--- a/packages/tbd/src/cli/commands/doctor.ts
+++ b/packages/tbd/src/cli/commands/doctor.ts
@@ -7,8 +7,9 @@
  */
 
 import { Command } from 'commander';
-import { access, readdir, readFile, unlink } from 'node:fs/promises';
-import { join } from 'node:path';
+import { access, mkdir, readdir, readFile, rmdir, unlink } from 'node:fs/promises';
+import { isAbsolute, join, relative, resolve, sep } from 'node:path';
+import { randomUUID } from 'node:crypto';
 
 import { BaseCommand } from '../lib/base-command.js';
 import { requireInit } from '../lib/errors.js';
@@ -120,6 +121,79 @@ export function divergenceFinding(
     suggestion: 'Run: tbd sync to reconcile',
   };
 }
+
+/**
+ * True when the Git common dir lives outside the project checkout — the linked
+ * worktree shape that lets a checkout be writable while `$GIT_COMMON_DIR/tbd`
+ * (shared sync state + lock) is not. This is the generic signal behind the
+ * Codex-sandbox case in #164; it does not depend on any `CODEX_*` env var, which
+ * that sandbox did not expose.
+ */
+export function isCommonDirOutsideProject(gitCommonDir: string, projectRoot: string): boolean {
+  const rel = relative(resolve(projectRoot), resolve(gitCommonDir));
+  return rel === '..' || rel.startsWith(`..${sep}`) || isAbsolute(rel);
+}
+
+/**
+ * Build the "Shared lock writability" finding from a probe result.
+ *
+ * `code` is the errno from attempting to create a directory under the shared
+ * locks dir, or undefined on success. EPERM/EACCES is a hard error: every write
+ * command must acquire this lock, so an unwritable lock path breaks all writes
+ * (the #164 Codex-sandbox case) — and a lock tbd needs but cannot take is a
+ * fatal condition, not a soft warning. Any other probe failure is reported as a
+ * warning since it cannot be positively interpreted. Never `fixable`: tbd cannot
+ * widen a sandbox or change filesystem permissions itself.
+ */
+export function buildLockWritabilityFinding(params: {
+  code: string | undefined;
+  sharedLockPath: string;
+  sharedLocksDir: string;
+  sharedTbdDir: string;
+  gitCommonDir: string;
+  projectRoot: string;
+}): DiagnosticResult {
+  const { code, sharedLockPath, sharedLocksDir, sharedTbdDir, gitCommonDir, projectRoot } = params;
+
+  if (!code) {
+    return { name: 'Shared lock writability', status: 'ok', path: sharedLocksDir };
+  }
+
+  if (code === 'EPERM' || code === 'EACCES') {
+    const outsideProject = isCommonDirOutsideProject(gitCommonDir, projectRoot);
+    const details = [
+      `Cannot create the shared data-sync lock (${code}): ${sharedLockPath}`,
+      'Read-only commands work, but every write command (create, update, sync) needs',
+      'this lock, so they will fail until the lock path is writable.',
+    ];
+    if (outsideProject) {
+      details.push(
+        `The checkout (${projectRoot}) is writable, but the shared tbd state under`,
+        `${sharedTbdDir} lives in the Git common dir (${gitCommonDir}) outside it —`,
+        'a common situation in agent sandboxes such as Codex worktrees.',
+      );
+    }
+    const suggestion = outsideProject
+      ? `Grant write access to ${sharedTbdDir} (in an agent sandbox such as Codex, add it ` +
+        `to the writable roots), or re-run the write command with sandbox escalation.`
+      : `Ensure ${sharedTbdDir} is writable by this user (check filesystem permissions).`;
+    return {
+      name: 'Shared lock writability',
+      status: 'error',
+      message: `lock path not writable (${code})`,
+      path: sharedLockPath,
+      details,
+      suggestion,
+    };
+  }
+
+  return {
+    name: 'Shared lock writability',
+    status: 'warn',
+    message: `unable to verify (${code})`,
+    path: sharedLocksDir,
+  };
+}
 import { formatHeading } from '../lib/output.js';
 import {
   renderRepositorySection,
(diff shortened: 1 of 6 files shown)


jlevy (commented):
## Senior Engineering Review

I reviewed PR #165 against issue #164, checked out `claude/exciting-turing-2Puzl`, inspected the touched code in the surrounding `doctor`, common-dir path, lockfile, and data-context paths, and ran focused plus representative local validation.

My assessment: this is the right shape for the fix. The important choice is that `doctor` performs a real filesystem capability probe instead of trying to infer Codex or sandbox state from environment variables. The write-side change is also appropriately path-scoped: EPERM/EACCES on the shared lock becomes an actionable fatal error, while EPERM from the critical section itself is not misreported as a lock problem. Deferring the `tbd setup --codex` writable-roots helper also makes sense because that would edit machine-local, security-sensitive agent config and needs a separate confirmation UX.

### Findings

No blocking findings.

### Non-blocking observations

- The `SharedLockUnwritableError` write-side message is optimized for the Codex/linked-worktree case. In a plain filesystem-permission case where `$GIT_COMMON_DIR/tbd` is inside the checkout, it may still say the shared state is outside the process writable area. `doctor` already has the better inside/outside classifier, so a later refinement could pass the project root into the write-side error and mirror that wording. I do not consider this blocking for #164 because the target failure mode is the Codex sandbox case.
- While validating the new failure path with `--json`, a failed `create` emitted two JSON error objects. That comes from the existing `BaseCommand.execute()` plus top-level CLI error handler both rendering the failure after wrapping it, so it appears pre-existing rather than introduced by this PR. This path just makes it easy to notice.

### Documentation And Comments

- No PR-specific documentation gap found. The changed doctor output golden was updated, and the runtime diagnostic includes the actionable remediation. The broader design docs already describe the common-dir shared lock architecture; they do not appear to enumerate every current `doctor` check precisely.
- Existing PR comments/reviews: only automated DeepSource and coverage comments. `gh api repos/jlevy/tbd/pulls/165/comments` and `/reviews` returned empty, so I did not find unresolved human review feedback.

### Validation

- `pnpm install --frozen-lockfile --ignore-scripts` passed from the committed lockfile.
- `pnpm audit` ran per repo policy and reports existing advisories in the current dependency graph; this PR does not touch dependencies or the lockfile.
- `pnpm build` passed.
- `pnpm --filter get-tbd test doctor-lock-writability.test.ts doctor-lock-writability-e2e.test.ts` passed: 11 tests.
- `pnpm typecheck` passed.
- `pnpm lint:check` passed.
- Live Codex-style sandbox check: `node packages/tbd/dist/bin.mjs doctor --json` in this worktree exits 1 with `Worktree` and `Common-dir layout` OK, plus `Shared lock writability` error on EPERM with writable-roots/sandbox-escalation guidance.
- Live write-side check: `node packages/tbd/dist/bin.mjs create "codex lock probe discard" --no-sync --json` fails before issue data writes with the new `SharedLockUnwritableError` message.
- `pnpm test` locally passed 73/74 files and 1170 tests before `doc-references.test.ts` hit Vitest's default 10s `beforeAll` hook timeout in this sandbox. Rerunning that same file with `--hookTimeout=30000` passed, so I am treating that as a local timeout artifact rather than a PR regression.
- GitHub CI is green: Benchmark, Coverage & Lint, DeepSource: Secrets, and Test on macOS/Ubuntu/Windows all pass.

Recommendation: merge.

jlevy:
Addressed the two non-blocking observations from the review:

1. **Write-side error wording** (`SharedLockUnwritableError` always implying the sandbox case) — fixed in 899583a. Hoisted `isCommonDirOutsideProject` to `lib/paths.ts` (neutral layer) so the write-side error and the doctor finding share one classifier; the error now takes the project root and mirrors doctor's inside/outside wording: sandbox/writable-roots guidance when `$GIT_COMMON_DIR/tbd` is outside the checkout, plain filesystem-permission guidance when inside. Verified live with `chattr +i` that the inside-checkout case now reads *"…is not writable by this process … check filesystem permissions"* with no sandbox framing.

2. **Double JSON error object on a failed `--json` command** — confirmed pre-existing (`BaseCommand.execute()` renders + rethrows a wrapped `CLIError`, then the top-level CLI handler renders again), not introduced by this PR. Left out of scope here and tracked separately (bead `tbd-rtmj`) rather than folded in.

Full suite green locally (1173 tests); CI re-running on 899583a.

---

PR #175, opened 2026-06-13
docs: address cli-agent-skill-patterns gaps from issue #173 (pprose audit)

### LINKED ISSUE #173: cli-agent-skill-patterns: five small gaps surfaced by an L2-ish CLI (pprose)
## Context

I (re-)read `cli-agent-skill-patterns.md` on `origin/main` (~1509 lines, hugely expanded since v0.1.29) while comparing it against the implementation of [`pprose`](https://github.com/jlevy/practical-prose) — a Python CLI that self-installs as an Agent Skill into `.agents/skills/`, `.claude/skills/`, and a marker-bounded block in `AGENTS.md`.

Verdict first: the doc is in great shape. Codex is covered down to `core-skills/src/loader.rs` scopes; non-npm pinning idioms (`uvx --from pkg@<ver>`, `pipx run pkg==<ver>`, `go run mod@<ver>`) are all in §6.7; the `git config`-style implicit-when-unambiguous scope rule is codified in §6.6.2; the marker-bounded `AGENTS.md` block with `format=fNN` versioning is in §2 and §6.6; the integration ladder in §6.0 cleanly separates L0/L1/L2/L3.

The five items below are small but real things I bumped into where the doc's current shape forced a less-clean choice than the actual best practice. Each one is self-contained — feel free to triage independently.

## 1. There is an unnamed rung between L2 and L3

§6.0's ladder:

> L2 — self-installing skill, discovery-dirs only. […] stops there: **no managed `AGENTS.md` block**, no hooks, no `prime`/`setup` context machinery.
>
> L3 — full self-installing CLI platform. Everything in L2 **plus** a managed `AGENTS.md` block, lifecycle hooks (`SessionStart`/`PreCompact`), `prime` and `setup` commands, format-versioned migration with a forward-compatibility guard, and usually a knowledge-injection meta-skill.

There is a useful middle: **L2 + a compact managed `AGENTS.md` block, but no hooks, no `prime`/`setup`, no DocCache.** A small always-on bootstrap block (the tool exists, here is its trigger description, here is the pinned fallback runner) is a worthwhile lift with none of the L3 platform machinery. `pprose` lives here: it writes `.agents/skills/`, `.claude/skills/`, and a ~13-line marker-bounded `AGENTS.md` block — but no hooks, no `prime`, no DocCache.

Today the ladder forces such tools to either drop AGENTS.md (L2) or appear to take on the whole L3 platform burden. Suggestion: split L2 into **L2a** (discovery-dirs only — `qmd`) and **L2b** (discovery-dirs + managed AGENTS.md block — `pprose`), keeping L3 (full platform — `tbd`) where it is. Or just acknowledge the rung in a sentence and give it a name.

## 2. Format-versioning is presented as L3-only, but it is useful any time a tool writes a generated artifact

§6.0 closes with:

> The self-upgrade and format-versioning rules in §6.6 apply **only to L3** — L0–L2 never need them.

This reads too sharply. The `format=fNN` discipline (begin-line stamp on `AGENTS.md` blocks; `DO NOT EDIT … format=fNN` marker on generated `SKILL.md` files; refuse-to-clobber-newer guard) is just as load-bearing at L2 or L2b — anywhere a tool writes a generated artifact it may need to upgrade later. `pprose` stamps `format=f01` on both its `SKILL.md` files and its `AGENTS.md` block and uses the same forward-compat guard `tbd` uses, despite being well short of L3.

Suggestion: reframe §6.6's format-versioning as "any rung that writes generated artifacts" rather than tying it to L3. The hooks/prime/setup machinery is what is L3-specific; the format stamp is universal hygiene.

## 3. §6.7 does not address: what version does the *generator* bake when the generator itself is a dev build?

§6.7 is thorough on pinned zero-install runners, but assumes the generator can pin to its own running version. In practice, a developer running `mytool install` from an editable/dev checkout has a running version like `0.1.1.dev49+abc1234` that **was never published**, so `uvx mytool@0.1.1.dev49+abc1234` cannot resolve. Baking the dev pin into the generated skill produces a skill that fails the moment it runs.

`pprose` handles this with two pieces:

- A `DISCOVERY_VERSION` constant — the last real PyPI release — used as the fallback pin in generated artifacts when the running version is not a publishable release ([source](https://github.com/jlevy/practical-prose/blob/main/tools/pprose/src/pprose/install.py#L50-L57)).
- A PEP-440 release-form check (`is_pypi_release`) that gates whether the running version is bakeable. Dev / pre-release / local versions fall back to `DISCOVERY_VERSION` ([source](https://github.com/jlevy/practical-prose/blob/main/tools/pprose/src/pprose/install.py#L86-L112)).

Worth one paragraph in §6.7, since this is non-obvious and easy to get subtly wrong (a single works-on-my-machine install can ship a broken pin to a teammate's clone).

## 4. Marker-block example inconsistency: what is `surface=agents-md` for?

§2 example (line 206) and §6.6 example (line 697) both show:

```markdown
<!-- BEGIN MYCLI INTEGRATION format=f02 surface=agents-md -->
```

…but most prose discussing the marker only mentions `format=fNN`. `pprose`'s own marker is `<!-- BEGIN PPROSE INTEGRATION format=f01 -->` with no `surface=`. The doc never explains:

- What `surface=` is for (one would assume disambiguating between AGENTS.md vs SKILL.md, but the doc explicitly says the artifact type is identified by **location**, not by an in-file tag).
- Whether it is load-bearing for the detection regex or just a human-readable marker.
- Whether a tool author should emit it.

Either drop `surface=` from the examples or add a sentence explaining what it does. (The doc's own §6.6 reasoning — "the artifact's identity is clear from its location … so no in-file `surface=` tag is needed beyond the load-bearing `format=fNN`" — suggests dropping it.)

## 5. Multi-block collapse: what to do if multiple stale blocks accumulate?

§2 covers the "AGENTS.md exists but has no marker block" case nicely:

> If `AGENTS.md` already exists without your markers, preserve the file and append a compact marked block instead of overwriting user content.

But the spec is silent on **multiple stale blocks** — e.g. a marker-prefix change over time, or an older format that was not collapsed on the prior install. `pprose` handles this by finding every match of the `BEGIN PPROSE INTEGRATION … END PPROSE INTEGRATION` regex, replacing all of them with one current block at the position of the **first** stale block, and discarding the others ([source](https://github.com/jlevy/practical-prose/blob/main/tools/pprose/src/pprose/install.py#L401-L433)).

This is not hypothetical — a tool that shipped a `BEGIN PPROSE BLOCK` marker in v0.0.x and switched to `BEGIN PPROSE INTEGRATION` in v0.1.x will have users with both blocks until upgraded. Worth a sentence in §2 or §6.6.

---

Happy to PR any of these — let me know if you would rather take prose suggestions inline or close as wontfix on any.

Cross-link: noticed only one open issue (#161, "Know that; fetch how" naming) — these are independent of that proposal.

jlevy:
Closes #173.

Folds the five pprose-surfaced gaps from #173 into the `cli-agent-skill-patterns` guideline, plus a wider distribution audit that keeps our research current. Each guideline edit is tagged with the ladder rung it touches, so the §0 simple baseline stays untouched. All five gaps were verified against pprose&#39;s working `install.py` (not just the issue prose) before editing.

## Guideline edits (the five gaps)

1. **Name the L2b variant** (§6.0, §0.3) — a self-installing skill that writes a compact managed `AGENTS.md` block but takes on none of the L3 platform (no hooks/`prime`/`setup`/DocCache). Named as a **variant within L2** (reference: `pprose`); the four-rung ladder **L0–L3 is unchanged, with no L2a/L2b renumber** — the owner chose the lighter single-named-sentence form over a full ladder split. `qmd` stays the plain-L2 reference.
2. **Format-versioning is artifact-driven, not L3-only** (§6.0 closer, §6.6) — any rung that writes a merged/managed artifact (an `AGENTS.md` block, or a committed generated skill shared across versions) needs the format stamp and forward-compat guard. Hooks/`prime`/`setup`/DocCache stay L3-specific.
3. **Generator-side dev-build pin rule** (§6.7) — bake the running version only if it is a real, resolvable published release; otherwise fall back to a known-good published pin (pprose&#39;s `DISCOVERY_VERSION` + `is_pypi_release`). A dev/editable checkout&#39;s version (`0.1.1.dev49+abc1234`) never resolves via `uvx`/`npx`.
4. **Drop the redundant `surface=` tag** (§2, §6.6) — both marker examples now carry only `format=fNN`; artifacts are identified by location. (pprose already dropped it; `tbd`/`flowmark` still emit it — see follow-on below.)
5. **Multi-block collapse** (§2, with a §6.6 pointer) — on install, collapse stale duplicate blocks to one at the first match, matching a small set of known legacy begin-prefixes.

**Current-practice corrections** (§6.6 native-scanning table): Cursor now scans `.agents/skills/` natively (verified against Cursor docs); native per-agent skill dirs (`.cursor/skills/`, `.github/skills/`, `.gemini/skills/`, Google Antigravity `.agent/skills/`, …) are proliferating. §11/§12 and References updated (pprose L2b variant, qmd L2, taste-skill/anthropics L0).

## Research (the broader audit)

- **New** `research-2026-06-13-skill-distribution-landscape.md` — channel taxonomy, a distribution decision matrix (when each rung/channel is right), case studies across L0/L2/L2b/L3 (taste-skill, anthropics/skills, qmd, pprose, tbd), and the skills.sh leaderboard reality: the most-installed skills are simple L0/L1 vendor knowledge skills (Vercel `find-skills` ~2.0M, Anthropic `frontend-design` ~540K, Microsoft Azure ~5.8M), not self-installing platforms. Distribution complexity does not track adoption. The doc records both shapes considered for gap 1 (the L2a/L2b split vs. a single named sentence) and which was adopted.
- **New** `plan-2026-06-13-cli-skill-guideline-pprose-gaps.md` — maps each gap to its edit; Phase 1 marked done and the gap-1 Open Question resolved to the single named variant.
- **Refreshed** three stale research docs (foundational CLI-as-skill brief; standard-paths native-dirs note; meta-skill 15K→1% budget correction).

## Scope notes

- All docs were formatted with the repo&#39;s flowmark (`format:md`); new content strictly follows the common-doc guidelines (no spaced em dashes; &#34;and&#34; not `+`/`&amp;` in new prose; Title Case headings). Pre-existing `+` in untouched sections of the guideline was left alone to keep this diff substantive rather than a repo-wide style sweep.
- **Follow-on (not in this PR):** gap 4 implies a small `tbd` generator change — `setup.ts` emits `surface=agents-md` on the `AGENTS.md` begin line, which the guideline now calls unnecessary. Dropping it is safe (block is rewritten whole; detection is prefix-anchored) but is a code change with golden/drift-test updates, best tracked separately.

https://claude.ai/code/session_01MvbozVS4yEWoR4pkJCMSkJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01MvbozVS4yEWoR4pkJCMSkJ)_

### DIFF
--- a/docs/project/research/current/research-2026-06-13-skill-distribution-landscape.md
+++ b/docs/project/research/current/research-2026-06-13-skill-distribution-landscape.md
@@ -1,0 +1,466 @@
+# Research Brief: Agent-Skill Distribution Landscape and Popular-Skill Patterns
+
+**Last Updated**: 2026-06-13
+
+**Status**: Complete
+
+**Related**:
+
+- [CLI as Agent Skill](./research-cli-as-agent-skill.md) (foundational research behind
+  the guideline)
+- [Agent Skills Standard Paths](./research-agent-skills-standard-paths.md)
+- [Skills vs Meta-Skill Architecture](./research-skills-vs-meta-skill-architecture.md)
+- Guideline this informs: `tbd guidelines cli-agent-skill-patterns`
+  ([source](../../../../packages/tbd/docs/guidelines/cli-agent-skill-patterns.md))
+- Issue motivating this pass: [jlevy/tbd#173](https://github.com/jlevy/tbd/issues/173)
+  (five gaps surfaced by the `pprose` L2-ish CLI)
+
+* * *
+
+## Executive Summary
+
+`cli-agent-skill-patterns` is a strong, current guideline, but it was shaped mostly by
+two reference points: `tbd` itself (a full L3 platform) and `qmd` (an L2
+self-installer). This brief widens the evidence base by auditing how a spread of
+**actually popular and actually shipped** skills are distributed, then maps the
+resulting decision framework.
+It exists to make sure our research reflects mid-2026 practice and to ground the five
+edits proposed in issue #173.
+
+Two headline findings.
+First, **the most-installed skills in the ecosystem are simple L0/L1 knowledge skills**,
+not platforms: the skills.sh all-time leaderboard is led by Vercel’s `find-skills`
+(~2.0M installs), Anthropic’s `frontend-design` (~540K), and Microsoft’s Azure skills
+(~5.8M across the set).
+Distribution complexity does not correlate with adoption; the opposite, if anything.
+Second, there is a **real, populated rung between the guideline’s L2 and L3** that we
+currently under-name: a self-installing tool that writes a compact managed `AGENTS.md`
+block (with format stamping and a forward-compat guard) but takes on none of the L3
+platform machinery (no hooks, no `prime`/`setup`, no DocCache).
+`pprose` is a clean reference for it, and its install code already solves three problems
+our guideline either ties to L3 or leaves unspecified.
+
+The practical output is a distribution decision matrix (which channel/rung fits which
+tool), a set of current-practice corrections (native per-agent skill dirs are
+proliferating; Cursor now scans `.agents/skills/` natively), and concrete backing for
+the issue #173 edits.
+
+**Research Questions**:
+
+1. How are popular, real-world skills actually distributed, and does distribution
+   complexity track adoption?
+
+2. What distribution channels exist in mid-2026, and when is each the right choice?
+
+3. Is the L0–L3 ladder complete, or is there a missing rung between L2 and L3?
+
+4. Which install-time concerns (format stamping, dev-build version pinning, marker
+   collapse, redundant in-file tags) are universal vs rung-specific?
+
+5. What has changed in the ecosystem since the guideline’s last substantive update
+   (native per-agent skill directories, Cursor’s discovery behavior)?
+
+* * *
+
+## Research Methodology
+
+### Approach
+
+Source-level audit of cloned repositories (via the `checkout-third-party-repo` shortcut,
+into `attic/`), cross-checked against the skills.sh leaderboard and curated catalogs.
+Where the issue cited `pprose` source lines, each claim was verified against the working
+code rather than the prose.
+Ecosystem-wide claims (per-agent skill paths, Cursor discovery) were checked against
+vendor docs and flagged where only community catalogs corroborate them.
+
+### Sources
+
+- **Cloned and read at source** (`attic/`): `Leonxlnx/taste-skill`,
+  `jlevy/practical-prose` (pprose), `tobi/qmd`, `anthropics/skills`,
+  `VoltAgent/awesome-agent-skills`, `heilcheng/awesome-agent-skills`.
+- **Leaderboard / discovery**: [skills.sh](https://skills.sh) all-time install rankings.
+- **Vendor docs**: Cursor Agent Skills docs (native `.agents/skills/` discovery), Codex
+  skills loader scopes (already cited in the guideline).
+- **Internal**: the guideline, the foundational research brief, the `pprose` install
+  spec (`plan-2026-05-30-pprose-install-scopes-and-surfaces.md` in the pprose repo).
+
+* * *
+
+## Research Findings
+
+### 1. What is actually popular (and what shape is it?)
+
+**Status**: ✅ Complete
+
+The skills.sh all-time leaderboard (mid-2026) is dominated by knowledge/best-practice
+skills published by vendors, not by self-installing platforms:
+
+| Rank | Skill | Publisher | Installs | Rung |
+| --- | --- | --- | --- | --- |
+| 1 | `find-skills` | vercel-labs/skills | ~2.0M | L0 (meta: finds other skills) |
+| 2 | `frontend-design` | anthropics/skills | ~540K | L0 (prompt and scripts) |
+| 3 | `vercel-react-best-practices` | vercel-labs | ~474K | L0/L1 |
+| 4 | `agent-browser` | vercel-labs | ~447K | L1 (points at a tool) |
+| 5 | `microsoft-foundry` | microsoft/azure-skills | ~390K | L1 (vendor SDK/CLI) |
+
+Microsoft’s Azure skill set alone reports ~5.8M installs.
+The leaders are vendor knowledge skills (React, Azure, Remotion best-practices) and
+design skills.
+
+**Assessment**: Adoption is driven by *usefulness and reach*, not by install machinery.
+The highest-leverage distribution move is shipping a spec-compliant `SKILL.md` that
+skills.sh and the scrapers can find.
+This validates the guideline’s “stop as low as you can” stance and argues the simple
+baseline (§0) should stay the loud default.
+It also means our L3-heavy reference set (`tbd`) is the *exception*, and the guideline
+should not read as if self-installing platforms are the norm.
+
+* * *
+
+### 2. The distribution channels, and when each is right
+
+**Status**: ✅ Complete
+
+Distinct, composable channels observed in the wild (a single public repo can satisfy
+several at once):
+
+| Channel | Mechanism | Best when | Seen in |
+| --- | --- | --- | --- |
+| **Repo-root `skills/<name>/SKILL.md`** | `npx skills add owner/repo` (skills.sh) scans it; scrapers index it | Always—the universal, zero-infra baseline | taste-skill, anthropics/skills, qmd |
+| **Claude/Codex plugin marketplace** | `.claude-plugin/marketplace.json` (plus `.agents/plugins/...`); `/plugin marketplace add` | You want a *bundle* (skills, MCP, and hooks) installable as a unit, or in-Claude discovery | anthropics/skills, qmd, taste-skill |
+| **Native per-agent skill dir** | Commit `.claude/skills/`, `.cursor/skills/`, `.github/skills/`, … | Targeting one agent that reads only its own dir (Claude Code) | All (Claude mirror) |
+| **Self-installing CLI (`tool install`)** | Tool writes its own skill into discovery dirs | The skill ships *with* a binary and should track its version | qmd (L2), pprose (L2b), tbd (L3) |
+| **Managed `AGENTS.md` block** | Marker-bounded, format-stamped block | You want an always-on bootstrap pointer that every agent reads | pprose (L2b), tbd (L3) |
+| **Agent-specific instruction file** | `.github/copilot-instructions.md`, `GEMINI.md` | A target agent auto-reads a bespoke file | taste-skill (Copilot) |
+| **`llms.txt` manifest** | Plain catalog of skills and one-line descriptions | Cheap machine-readable index for crawlers/agents | taste-skill |
+| **MCP server** | `npx`/`uvx` server in marketplace or agent config | No CLI fits, or you need OAuth/multi-tenant/remote | qmd (alongside CLI) |
+
+**Assessment**: The guideline covers most of these but frames the repo-root and
+skills.sh path as one option among many.
+The evidence says it is *the* default and everything else is additive.
+Two channels are under-documented: agent-specific instruction files
+(`copilot-instructions.md`) and `llms.txt`, both of which taste-skill uses to widen
+reach at near-zero cost.
+
+* * *
+
+### 3. Case studies across the ladder
+
+**Status**: ✅ Complete
+
+- **taste-skill (L0, 13-skill collection)**—pure prompt skills, no binary.
+  Distributed through *five* channels simultaneously: `npx skills add`, a Claude plugin
+  marketplace (`.claude-plugin/marketplace.json` and `plugin.json`),
+  `.github/copilot-instructions.md`, a `skills/llms.txt` manifest, and manual copy.
+  **Versioning is by stable install-name, not format stamps**: the v2 rewrite keeps the
+  same `name: design-taste-frontend` (so re-running `npx skills add` overwrites in
+  place) and preserves v1 under a separate install-name `design-taste-frontend-v1` for
+  callers that pinned to its behavior.
+  No `format=fNN`, no `DO NOT EDIT`—none is needed, because nothing writes a managed
+  file it must later migrate.
+  The folder name and the frontmatter `name` deliberately differ (e.g.
+  `brutalist-skill/` → `name: industrial-brutalist-ui`); the install-name is the
+  contract.
+- **anthropics/skills (L0 and bundled scripts)**—the official examples.
+  Repo-root `skills/<name>/SKILL.md`, some with `scripts/` (xlsx,
+  web-artifacts-builder).
+  Primary distribution is a **Claude plugin marketplace** that groups the 18 skills into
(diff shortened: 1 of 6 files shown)


jlevy:
## Senior Engineering Review: Skill Distribution Guidance

I reviewed PR #175 (`docs: address cli-agent-skill-patterns gaps from issue #173`) at head `2c54b531af967074be541c48d4339c80c7575760`, including the diff, PR discussion, local project docs, the tbd workflow guidance, and the external skill-distribution sources cited or implied by the PR.

### Bottom line

I agree with the PR's main direction: the guideline should strongly recommend **stopping at the lowest integration rung that works**. Most tools should ship a small repo-root `skills/<name>/SKILL.md` and maybe a self-installing skill mirror. Very few should copy tbd's full L3 setup with hooks, `prime`, `setup`, DocCache, and a meta-skill.

I would request changes before merge because a few internal contradictions still weaken the recommendation.

### Findings

1. **[P1] §6.6 still scopes upgrade/migration guidance to “L3 only.”**

   `packages/tbd/docs/guidelines/cli-agent-skill-patterns.md:571-577` correctly says format versioning is artifact-driven and applies to any rung that writes a generated artifact it may later need to upgrade or refuse to clobber. That explicitly includes an L2b managed `AGENTS.md` block.

   But `packages/tbd/docs/guidelines/cli-agent-skill-patterns.md:816-819` still says:

   - “Upgrade existing installs deliberately (L3 only).”
   - “An L2 tool that only writes discovery-dir skills can skip all of this...”

   That conflicts with the new L2b definition. An L2b tool writes a managed `AGENTS.md` block, so it needs the §6.6 format stamp and forward-compatibility guard even though it is not L3. I would rename this section around the artifact, e.g. “Upgrade managed generated artifacts deliberately,” and change the parenthetical to “An L2a tool...” or “A tool that only writes fully-overwritten discovery-dir skills...”

2. **[P2] The refreshed standard-path research still contains stale Cursor/native-scan claims.**

   `docs/project/research/current/research-agent-skills-standard-paths.md:144-149` says Cursor now scans `.agents/skills/` natively, but the table and precision note still say otherwise:

   - `:203` says Cursor is `.agents/skills/` “per skills CLI support” and that path docs are thinner.
   - `:217-224` says Cursor is reached via `npx skills add`, not native scan.

   Cursor's current docs corroborate native loading from `.agents/skills/`, `.cursor/skills/`, `~/.agents/skills/`, and `~/.cursor/skills/`, plus compatibility loading from `.claude/skills/` and `.codex/skills/`. The research doc should reconcile this instead of carrying both claims.

   More generally, I would keep the distinction explicit: “verified vendor-native scan” vs “Vercel installer support.” Vercel's supported-agent table is useful evidence for distribution reach, but it should not be treated as source verification for every vendor's native loader.

3. **[P3] The plan/spec still reads like pre-implementation state.**

   `docs/project/specs/active/plan-2026-06-13-cli-skill-guideline-pprose-gaps.md:10` still says this is a draft proposal and the guideline edits are not applied. `:202-210` still lists open questions that the PR has already answered, including the L2a/L2b split and Cursor native-scanning confirmation.

   I would update this to “implemented/review” state, mark the decisions as decided, or move it out of active draft state. Otherwise future agents will treat already-settled PR decisions as unresolved planning work.

### Research corroboration

The main research claims check out.

- [skills.sh](https://skills.sh/) corroborates the public agent-skills ecosystem and leaderboard shape, including large install counts for small knowledge/prompt skills.
- [Vercel Skills supported agents](https://www.mintlify.com/vercel-labs/skills/guides/supported-agents) corroborates the broad installer/distribution story and the importance of `.agents/skills/` as a portable project-level convention.
- [Cursor Agent Skills docs](https://cursor.com/docs/skills) corroborate native Cursor loading from `.agents/skills/`, `.cursor/skills/`, and user-level variants.
- [pprose install.py](https://github.com/jlevy/practical-prose/blob/main/tools/pprose/src/pprose/install.py) corroborates the L2b pattern: `.agents/skills/`, `.claude/skills/`, marker-bounded `AGENTS.md`, `format=fNN`, a forward-compatibility guard, collapse of stale duplicate managed blocks, and release-pinned runner instructions.
- [qmd](https://github.com/tobi/qmd) supports the L2a pattern: installable skill plus plugin/MCP packaging without a managed `AGENTS.md` block.
- [anthropics/skills](https://github.com/anthropics/skills) and [taste-skill](https://github.com/Leonxlnx/taste-skill) support the repo-root skills and plugin-marketplace distribution examples.

The nuance I would add: the ecosystem is fragmenting into many native directories, but the best response is not for every CLI to write every native path itself. The durable baseline is still canonical repo-root skills plus `.agents/skills/` portability and a `.claude/skills/` mirror where Claude Code support matters. Let broader installers or plugin packaging fan out additional per-agent mirrors unless the tool has a strong reason to own that surface.

### Holistic recommendation

I would make the guideline's recommendation ladder even sharper:

1. **Default: repo-root canonical skill.** Ship `skills/<name>/SKILL.md` in the repo. Keep it small, installable, and readable by humans and crawlers.
2. **Skill routes to CLI.** A skill should teach the agent when and how to invoke the CLI, not duplicate full CLI help or become a second product surface.
3. **Portable project mirror: `.agents/skills/`.** Use this as the cross-agent project path.
4. **Claude mirror: `.claude/skills/`.** Required if Claude Code support is first-class, because Claude Code does not treat `.agents/skills/` as the canonical native location.
5. **L2a: self-installing skill only.** A CLI may write its own discovery-dir skill and stop there. Full overwrite is enough; no migration apparatus needed.
6. **L2b: self-installing skill plus managed `AGENTS.md`.** Only add this when the tool needs always-on bootstrap instructions. Once it writes into a human-authored project-instruction file, it needs marker bounds, `format=fNN`, a forward guard, deterministic output, and duplicate-block cleanup.
7. **L3: full agent platform.** Hooks, `prime`, `setup`, DocCache, and meta-skills are justified only for tools with broad lifecycle/state/knowledge-library needs. tbd qualifies; most CLIs do not.
8. **Plugin marketplaces are packaging channels, not the baseline.** Use them for bundled skills/MCP/hooks/trust/discovery. Do not make plugin packaging the required path for a basic skill to work.

### What tbd should do to dogfood this

I would turn the PR's recommendations into a small tbd dogfooding roadmap:

1. **Fix this PR's internal contradictions before merge.**
   - Rename or re-scope the §6.6 “L3 only” upgrade section.
   - Reconcile Cursor/native-scan research.
   - Update the active spec status and open questions.

2. **Make tbd's own distribution contract explicit.**
   - State that tbd is intentionally L3, not the baseline example for ordinary CLIs.
   - Document the exact surfaces tbd owns: generated skill(s), managed `AGENTS.md` block, hooks, setup/prime, DocCache/meta-skill.
   - State which lower rungs tbd recommends for simpler tools and why.

3. **Generate and test tbd's own canonical skill from one source.**
   - Ensure the source of truth for the tbd skill can produce repo-local `.agents/skills/tbd/SKILL.md` and `.claude/skills/tbd/SKILL.md` consistently.
   - Add/keep drift tests so generated skill content cannot silently diverge from bundled docs.
   - Keep the skill small: it should route agents to `tbd prime`, `tbd shortcut`, and `tbd guidelines`, not embed the whole manual.

4. **Clean up tbd's managed marker format.**
   - The PR says `surface=` is redundant because artifact identity comes from file location and marker name. I agree.
   - tbd should follow through with a real migration: accept old markers with `surface=...`, emit the cleaner marker going forward, collapse duplicates, and guard newer `format=fNN` blocks.

5. **Make format migration artifact-driven in code and tests.**
   - Test managed `AGENTS.md` blocks independently from hook files and skill files.
   - Treat each generated artifact as having its own format/migration contract where needed.
   - Avoid implying that format stamps exist only because a tool is L3.

6. **Improve setup/doctor output around rungs and surfaces.**
   - `tbd setup`/`doctor` should be able to explain what surfaces are installed and why tbd uses L3.
   - The message should reinforce the public recommendation: simpler tools usually want L1, L2a, or L2b.

7. **Add a visible follow-up bead for the tbd generator cleanup.**
   - I would not force the generator migration into this docs PR if the intent is to keep it pure docs.
   - But the follow-up should be tracked explicitly: remove/deprecate `surface=`, add compatibility tests, verify generated output, and update docs/examples after the migration lands.

### Suggested merge stance

I would not block on the tbd generator cleanup if it is tracked as a follow-up. I would block on the three documentation consistency issues above, because they are in the recommendation surface this PR is changing.

jlevy:
Thanks for the detailed review. Addressed the three blocking findings in `1208ee5` (on top of the gap-1 revision in `02d4a8e`):

**P1 — §6.6 "L3 only" upgrade scope.** Retitled "Upgrade existing installs deliberately (L3 only)" → **"Upgrade managed generated artifacts deliberately"**, and rewrote the parenthetical around the *artifact*, not the rung: a plain-L2 tool (discovery-dirs only, full overwrite) still skips migration, but the **L2b variant** — which merges a managed `AGENTS.md` block into a human-authored file — needs the format stamp and forward-compat guard *even though it isn't L3*. No longer contradicts the artifact-driven rule in §6.0.

**P2 — Cursor native-scan contradiction.** Verified against the current Cursor Agent Skills docs: native recursive scan of `.agents/skills/` and `.cursor/skills/` (+ the `~/` variants), with compatibility loading from `.claude/skills/` and `.codex/skills/`. Moved Cursor from the installer-reach bucket to **verified-native** in both the path table and the precision note, and made the **verified-vendor-native vs. Vercel-installer-reach** distinction explicit (Vercel's supported-agent table is installer reach, not proof a vendor's own loader reads the path).

**P3 — spec reads pre-implementation.** Flipped Status from "Draft (not yet applied)" → **"Implemented … this spec is a record, not an open proposal."** (The Open Questions and Phase 1 checklist were already marked resolved in `02d4a8e`.)

Non-blocking items, considered:

- **Holistic recommendation ladder** — the guideline already frames plugin marketplaces as a packaging/publishing channel rather than the baseline (§1, the §6.6 Codex note, and the "Anthropic's own marketplaces are the exception" note at the end). Happy to fold your sharper 8-point ladder into one explicit summary block as a follow-up if you'd like it stated in a single place.
- **Dogfooding roadmap (7 pts)** — tracked as Phase 2 follow-ons in the spec: explicit tbd distribution contract, single-source canonical skill + drift test, artifact-driven format migration in code/tests, the `surface=` generator migration, and surface/rung output in `setup`/`doctor`. Kept out of this pure-docs PR per your merge stance.

One heads-up: since your review (made at `2c54b53`), the gap-1 shape changed per owner direction to a **single named "L2b" variant within L2** — no L2a/L2b ladder renumber; `qmd` is the plain-L2 reference, `pprose` the L2b one. Your holistic rec lists L2a/L2b as distinct rungs, so if you'd prefer the full split instead, say the word and I'll switch it back (it's a small change).

---

PR #196, opened 2026-07-20
Bead watch and external issue sync: tbd changes/watch, comments, tbd mirror

jlevy:
## What

Implement the bead-watch half of the plan as a fully validated Phase 1. Phase 2 remains
gated and has not started.

**Phase 1 — Watch.**

- `tbd changes --since <commit> [selection] [--json]` is the pure, read-only
  sync-branch snapshot diff with deterministic per-bead field deltas and exit 0/3.
- `tbd watch --bead|--label|--spec|--status|--ready|--all` polls the remote with
  `git ls-remote`, fetches only on movement to a private ref, emits the same report,
  and exits 0 for change, 3 for timeout, or 1 for error (exit 2 stays reserved for usage errors).
- The `watch-beads` shortcut includes a race-free watch-then-spawn daemon and bounded
  or background in-session recipes for Claude Code and Codex.
- Watch never touches the hidden data-sync worktree or its lock, `FETCH_HEAD`, or the
  configured sync refs.

**Phase 2 — Sync.** Comments, external mirror adapters, and reference workflows remain
unchecked in the spec and outside this implementation commit.

## Spec and validation

- Plan: `docs/project/specs/active/plan-2026-07-19-bead-watch-and-external-sync.md`
- Phase 1 validation:
  `docs/project/specs/active/valid-2026-07-19-bead-watch-phase-1.md`

## Plan

- [x] Senior engineering review; make baseline, selection, JSON, and private-fetch
  semantics explicit in the spec
- [x] Create and sync the Phase 1 epic and child beads (`tbd-m1at`)
- [x] Implement and test `tbd changes` (`tbd-q1em`)
- [x] Implement and test `tbd watch` (`tbd-l467`)
- [x] Add and package the `watch-beads` shortcut (`tbd-h4tf`)
- [x] Record the Claude Code background-wake demo (`tbd-q7rf`)
- [x] Record the Codex watch-then-spawn demo and command limits (`tbd-hb3p`)
- [x] Record the cross-agent conversation and concurrent-sync safety demo (`tbd-2y7v`)
- [x] Run local quality gates: formatting, lint/typecheck, build, 95 files / 1,407
  tests, publint, and the 14-day package-age check
- [x] Address the formal senior review (R1–R6 and S2–S3 fixed; S1 deferred as `tbd-fuiw`)\n- [x] Confirm the implementation CI on macOS, Linux, Windows, coverage/lint, and
  benchmark
- [x] Address the round-2 senior review (N1–N5 and S-B/S-C fixed; S-A deferred as
  `tbd-35nh`)
- [ ] Phase 2: comments, extensions merge, mirror collector/projector/closer, reference
  workflow, and dev-workspace end-to-end test

No dependency or lockfile changed. The setup audit's pre-existing dev-tool findings
remain tracked in `tbd-ujyy` and are intentionally outside this PR.

---

> [!NOTE]
> **Medium Risk**
> New Git remote polling and private-ref fetch paths must not disturb concurrent `tbd sync` or the hidden worktree; the PR adds explicit safety tests and hardening, but watch behavior is operationally sensitive for unattended agents.
> 
> **Overview**
> **Phase 1** adds read-only bead change detection and blocking watch so agents can wake on remote sync-branch updates without webhooks or touching the data-sync worktree. **Phase 2** (comments, `tbd mirror`, external sync) is documented in the spec but not implemented in this diff.
> 
> **`tbd changes`** diffs committed sync-branch snapshots between `--since` and the local tip: per-bead field deltas (including text hunks), shared selectors (`--bead`, `--label`, `--spec`, `--status`, `--ready`, `--all`), human/JSON output, and exit **0** vs **3** for match vs no match.
> 
> **`tbd watch`** polls `git ls-remote`, fetches only after tip movement into a **private ref** (no `FETCH_HEAD` or configured sync ref mutation), runs the same report, and exits on change, timeout (**3**), or error. It validates unknown bead IDs before an unbounded poll, tolerates brief remote outages, and reclaims stale private refs.
> 
> Shared logic lives in `issue-changes`, worktree-free snapshot reads via batched `git cat-file`, and extracted `issue-selection` reused by `list`/`ready`. Docs add the **`watch-beads`** shortcut (watch-then-spawn and in-session recipes), README/CLI help, changelog, and Phase 1 plan + validation transcripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac48a63ecf563605d924e7de7c438d1683fdcf39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy (commented):
## Senior Engineering Review — PR #196 (Bead watch, Phase 1)

**Scope:** All 23 changed files at head `62338b7` (+2,882/−53): new `tbd changes`/`tbd watch` commands, diff engine (`issue-changes.ts`), snapshot reader (`sync-branch-changes.ts`), poller (`bead-watch.ts`), shared selection predicates, `list`/`ready` refactors, 4 test files, `watch-beads` shortcut, spec + validation docs, README/CHANGELOG/tbd-docs. Reviewed from a local checkout with the repo's `review-code` guideline set (general, comments, error-handling, testing, TypeScript); Phase 2 (mirror/sync) is spec-only and reviewed as design, not code.

**Summary:** This is a high-quality, well-scoped increment. The architecture is exactly right for the stated goal: a pure diff core (`createIssueChangesReport`) under an imperative shell with injectable poll/fetch/clock dependencies, a strict fail-loud snapshot reader, a stateless blocking-watch contract (`--since` chaining, gh-style exit codes), and genuinely paranoid Git safety (private collision-resistant refs, `--no-write-fetch-head`, never touching the data-sync worktree or lock) — all pinned by sentinel tests. The `list`/`ready` refactor into shared predicates was verified to preserve prior semantics exactly (including the dangling-blocker edge, which cannot occur since blocker IDs come from the same issue collection). Spec, validation transcript, and shortcut docs are exemplary.

**Verdict: Approve with follow-ups** — R1 and R2 are worth fixing before undraft/merge (both small); R3 before adoption on large/Windows repos; the rest are low.

### Findings

**R1 (Medium, UX/correctness) — unknown `--bead` ID is not validated until first remote movement.**
`tbd watch --bead tbd-TYPO` (without `--since`) performs no initial report; bead IDs are only resolved inside `createIssueChangesReport`, which first runs after the remote tip moves. A typo'd ID therefore blocks silently — potentially for hours — instead of failing fast (`packages/tbd/src/file/bead-watch.ts:119-147`, `packages/tbd/src/lib/issue-changes.ts:212-227`). With `--since`, the immediate comparison surfaces the error; without it, the failure path violates the error-handling rule that failure behavior is part of the feature.
Fix: when `selection.kind === 'beads'`, pre-resolve IDs at watch startup against the local sync-branch snapshot (or the first observed tip) and exit 1 on unknown IDs before entering the poll loop; the per-report resolution stays as the authority.

**R2 (Medium, robustness) — interrupted watches orphan private refs.**
Ctrl-C is the normal way to stop a blocking watch, and SIGINT skips the `finally` cleanup, so any invocation that fetched at least once leaves `refs/tbd/watch/<pid>-<uuid>` behind permanently, pinning fetched objects against GC and accumulating without bound (`packages/tbd/src/file/bead-watch.ts:52,100-107`). The CHANGELOG's "removes its private ref in a `finally` path" overstates this slightly.
Fix (pick one): best-effort startup sweep deleting `refs/tbd/watch/*` whose embedded PID is no longer alive (`process.kill(pid, 0)`); and/or install SIGINT/SIGTERM handlers that delete the ref before exiting. Either way, reword the CHANGELOG line to "on normal exit" until done.

**R3 (Medium, performance — non-blocking for this repo, blocking for large ones) — O(2N) subprocess spawns per report, repeated on every unrelated movement.**
`readSnapshot` runs one `git show` per issue file per endpoint (`packages/tbd/src/file/sync-branch-changes.ts:43-69`), and the baseline-advance loop runs a full two-snapshot report on every remote movement that doesn't match the selection (`bead-watch.ts:138-146`). At this repo's ~450 beads that is ~900 spawns per movement tick; on Windows (where this PR already had to allow slower fixtures) that can approach a minute of spawn overhead per tick during busy multi-agent activity.
Fix: stream blobs through one `git cat-file --batch` process per snapshot; additionally, when the selection doesn't require full graphs (`all`/`beads`/label/spec/status without `--ready`), narrow parsing to `git diff --name-only <since> <tip> -- <issues-dir>` changed paths. `--ready` legitimately needs both full graphs.

**R4 (Low, output size) — hunks carry the entire field as context.**
`createTextHunks` emits one hunk spanning the whole field, so a one-line append to a 500-line `notes` field emits all ~500 lines as context (`packages/tbd/src/lib/issue-changes.ts:143-186`). Watch reports are designed to be piped into agent prompts, where this inflates tokens with a 50KB-body worst case.
Fix: bound context to a few lines around the changed span (unified-diff style), keeping determinism.

**R5 (Low, output noise) — created/deleted reports include null→null fields.**
For created/deleted beads every one of the 19 normative fields is emitted, including pairs where both sides normalize to `null` (e.g. `assignee: null -> null` on create) (`packages/tbd/src/lib/issue-changes.ts:229-241`).
Fix: in the created/deleted branch, skip fields where both normalized values are `null`.

**R6 (Low, edge) — explicit empty-string `--spec` now activates the filter.**
The shared-predicate refactor passes `options.spec ?? null`, so `--spec ""` becomes an active filter in `list`/`changes`/`watch`, where the old `list` code's truthiness check ignored it (`packages/tbd/src/cli/commands/list.ts:206-213`, `packages/tbd/src/cli/lib/change-selection.ts:44`). Only reachable by explicitly passing an empty string.
Fix: treat `''` as absent (or reject it as a usage error).

- **S1 — Phase 2 reconciliation with the sibling plan.** A parallel spec landed on `claude/linear-bead-sync-plan-tct4hn` (`plan-2026-07-20-linear-bead-sync-pilot.md`, being rebased on top of this branch) covering the same Phase 2 territory with different choices: `tbd bridge` vs `tbd mirror` naming, a first-class `linked` field with a one-external-source-per-bead invariant vs label + HTML-marker bindings, and base-snapshot echo suppression vs content hashes (complementary, not conflicting). Before Phase 2 starts, fold these into one plan of record so the command-group name and binding representation are decided once.
- **S2 — Document watch behavior across sync-branch history rewrites.** The unrelated-history rescue path can rewrite the sync branch; a running watch will then exit 1 with the not-an-ancestor error by design. Worth one troubleshooting line in `watch-beads`/tbd-docs so users know to restart without `--since`.
- **S3 — `tbd changes` local-tip prerequisite.** On a fresh clone before any sync, `refs/heads/tbd-sync` may not exist and `tbd changes` errors with the raw ref name. A hint ("run `tbd sync` first") would make the failure self-explanatory.

### False positives / do not fix

- **Timeout-path cleanup is safe:** `git update-ref -d` on the never-created private ref exits 0 (verified empirically), so a pure-timeout watch correctly exits 2; the `finally` cleanup cannot mask it.
- **`ready` refactor is semantics-preserving:** old inline logic and `readyIssueIds` agree on all inputs, including missing-blocker references (impossible by construction — blocker IDs are the iterating issues' own IDs).
- **Low vitest coverage on `changes.ts`/`watch.ts`/`change-selection.ts` is an instrumentation artifact:** these paths are exercised end-to-end by `cli-changes.test.ts`/`cli-watch.test.ts` through spawned CLI subprocesses, which the coverage reporter cannot see.
- **`--timeout 0 --since <commit>` acts as a one-shot remote comparison** — coherent and useful, matches the spec's immediate-comparison rule.

### Design assessment

The polling-over-`ls-remote` transport with fetch-on-movement is the right zero-infrastructure choice, and the blocking one-shot contract (exit, report, chain via `--since`) is better for agent wake-ups than a streaming daemon would have been: stateless, resumable, and composable from any shell. The edge-triggered `--ready` (computed independently from each endpoint's full dependency graph, with "became ready" rendering for beads whose own fields didn't change) is a subtle piece of design done correctly. Alternatives (fs-watchers, git hooks, webhook receivers) are all worse on at least one of: sandbox compatibility, remote visibility, or infrastructure burden. The strict fail-loud snapshot validation (nested paths, filename/ID mismatch, mapping bijection, unmapped issues) correctly refuses to produce plausible-but-wrong reports from a corrupted store — consistent with the lesson of the structured-merge spec.

### Documentation

Spec (`plan-2026-07-19-bead-watch-and-external-sync.md`) and validation (`valid-2026-07-19-bead-watch-phase-1.md`) are the standard other PRs should be held to. README, CHANGELOG, tbd-docs, and the help transcript are all updated. Gaps: the S1 spec reconciliation and the S2/S3 troubleshooting lines above; CHANGELOG wording per R2.

### CI status

All checks green at `62338b7`: Benchmark, Coverage & Lint, Test (macOS/Ubuntu/Windows), DeepSource Secrets. DeepSource AI Review shows "not requested" (pending by configuration, not a failure). PR is draft.

jlevy:
Addressed review findings in commit `69145dc`.

| Finding | Disposition |
| --- | --- |
| R1 — unknown `--bead` can poll indefinitely | Fixed. A watch without `--since` validates static IDs against the local committed sync snapshot before the first remote query. A watch with `--since` keeps the immediate two-snapshot union authoritative, including deletions. CLI coverage proves an unknown ID wins even when the remote branch is absent. |
| R2 — SIGINT can orphan private refs | Fixed. Normal completion still deletes the invocation ref in `finally`; each new watch also sweeps `refs/tbd/watch/<pid>-*` entries whose owner PID is dead. Unrecognized or live-owner refs are left alone. The changelog/spec now state the actual lifecycle. |
| R3 — one `git show` per issue per endpoint | Fixed. Each strict endpoint read now lists the tree once and reads all issue/mapping blobs through one `git cat-file --batch` subprocess. The test asserts two batch reads total for a two-endpoint report. Full-graph parsing is intentionally retained because the contract requires fail-loud validation of the snapshot and `--ready` requires both complete dependency graphs; changed-path narrowing would weaken that invariant without a separately validated cache. |
| R4 — unbounded text hunk context | Fixed. Deterministic Myers line diffs are grouped into unified-style hunks with at most three context lines; tests cover distant edits and a one-line append to 500-line notes. |
| R5 — null-to-null created/deleted fields | Fixed. Fields normalized to `null` at both endpoints are omitted. |
| R6 — empty `--spec` compatibility | Fixed. `--spec ""` is absent for list/change selection semantics; watch still correctly reports that no selector was supplied. |
| S1 — reconcile Phase 2 designs | Deferred explicitly as `tbd-fuiw` and added as the first unchecked Phase 2 spec item. Phase 2 remains unimplemented. |
| S2 — history-rewrite recovery | Documented: restart without `--since` after sync recovery invalidates the saved ancestry. |
| S3 — missing local sync tip | Fixed with an actionable `Run tbd sync first` error and CLI coverage. |

The review’s noted false positives required no code changes: timeout cleanup already traverses `finally`, ready semantics remain edge-triggered, subprocess CLI coverage is an instrumentation artifact, and timeout `0` with `--since` remains a coherent one-shot comparison.

Validation: `pnpm precommit` passed (format, lint/typecheck, build, 95 files / 1,407 tests); Vitest coverage passed; publint and the 14-day package-age check passed. Five unrelated tryscript fixtures timed out only under the 29-minute all-files load and then passed 68/68 when rerun individually.

jlevy:
Follow-up validation: the Windows watch-safety fixture timeout adjustment landed in 4fb2f0d. GitHub Actions run 29768482548 completed successfully: coverage/lint, benchmark, Ubuntu, macOS, and Windows all pass. DeepSource Secrets also passes; the separate DeepSource AI Review status remains informational/pending because an AI review was not requested.

jlevy:
Addressed the review findings in `52867bd` (on `claude/linear-bead-sync-plan-tct4hn`, the combined branch / PR #197 that carries this PR's code — per the plan, fixes ride the branch moving forward rather than force-updating this draft):

- **R1: fixed** — `tbd watch --bead` without `--since` now performs one startup fetch and a tip-to-tip report, so unknown bead IDs exit 1 immediately instead of blocking until the first remote movement. Unit test (injected deps) + e2e test (`exits 1 immediately on an unknown --bead ID`).
- **R2: fixed** — new `sweepStaleWatchRefs` runs at watch startup and deletes `refs/tbd/watch/*` whose embedded PID is no longer alive (live and unrecognized refs untouched); CHANGELOG claim reworded to "on normal exit" plus the sweep. Test covers dead/live/legacy refs.
- **R3: deferred** — tracked as `tbd-293h` (status: deferred): replace per-file `git show` with `git cat-file --batch` and/or changed-path narrowing when the selection doesn't need full graphs. Perf hardening before large-repo adoption; not blocking Phase 1.
- **R4: fixed** — `TEXT_HUNK_CONTEXT_LINES = 3` bounds hunk context around the changed span with adjusted `old_start`/`new_start` and counts; small-text hunks (including the spec's JSON example and the validation transcripts) are byte-identical to before. Bounded-context test added.
- **R5: fixed** — created/deleted reports omit fields that are `null` on both sides; test renamed and updated to pin the omission.
- **R6: fixed** — explicit empty `--spec` is treated as "no filter" in `list`/`changes`/`watch`, restoring `list`'s historic behavior.
- **S1: fixed** (in `d21a701`, prior commit) — the two Phase 2 plans are reconciled: `tbd bridge` naming, first-class `linked` field with a one-source-per-bead invariant, complementary echo suppression, comments model adopted; addenda recorded in both specs.
- **S2: fixed** — `watch-beads` documents the history-rewrite behavior (baseline-not-an-ancestor → exit 1; restart without `--since`).
- **S3: fixed** — `tbd changes` on a repo without a local sync branch now says: `Local sync branch 'tbd-sync' not found. Run \`tbd sync\` first to create it.`

Contract refinements are recorded in the 2026-07-20 addendum of `plan-2026-07-19-bead-watch-and-external-sync.md`. Verification: full suite 94 files / 1405 tests pass (6 new), lint + typecheck + build clean. Findings were tracked as beads `tbd-83e7`, `tbd-4v45`, `tbd-293h`, `tbd-m06c`, `tbd-o5d3`, `tbd-5e6e` under parent `tbd-k6bb` (all closed except the deferred `tbd-293h`).

jlevy:
## Senior Engineering Review — PR #196 (Bead watch, Phase 1) — round 2

**Scope:** All 24 changed files at head `53e8edd` (+3,481/−59), which is `4fb2f0d` merged with `origin/main` (`c458c21`, the PR #198 CLI-ergonomics release). Reviewed from a local checkout of the merged head with the repo's `review-code` guideline set (general, comments, error-handling, testing, TypeScript). This round verifies the fix commit `69145dc` and everything since against the prior review at `62338b7` (R1–R6, S1–S3), reviews the merge commit itself, and re-examines the full diff with fresh eyes. Prior findings are referenced, not re-litigated.

**Summary:** The prior review's verdict holds after the fixes and the merge. R1–R6 are implemented as claimed and each is pinned by tests (details in the verified list below). The batch snapshot reader added for R3 is carefully engineered: strict response framing, missing-blob and truncation failures are loud, and object IDs are deduplicated. The merge with main is semantics-preserving: the `ready.ts` conflict kept this branch's `readyIssueIds()` extraction (main's side was brace-only restyling of the code the branch replaced), the help transcript was re-wrapped for the wider column introduced by variadic `show <ids...>` and is proven exact by the passing tryscript goldens, and the new watch code was brought under main's re-enabled `curly` rule. New findings this round are operational, centered on the watch exit-code contract and long-run robustness of the daemon recipe.

**Verdict: Approve with follow-ups** — N1 and N2 are worth deciding before the watch exit-code contract ossifies (PR #197 chains on this code, and the author has stated fixes ride that branch; addressing them there is fine). N3–N5 are low.

### Findings

**N1 (Medium, API contract / robustness) — exit 2 is ambiguous between "timeout" and "usage error", and the daemon recipe spins hot on the latter.**
`tbd watch` exits 2 on timeout (`packages/tbd/src/cli/commands/watch.ts:59-62`, documented at `packages/tbd/docs/tbd-docs.md:592`), but every tbd usage error also exits 2 via `ValidationError` (`packages/tbd/src/cli/lib/errors.ts:39-48`): missing selector, `--interval` below minimum, invalid `--status`, conflicting selectors. The `watch-beads` watch-then-spawn recipe treats exit 2 as "timeout, poll again" (`packages/tbd/docs/shortcuts/standard/watch-beads.md:60-64`), so any adaptation of the recipe with a bad flag value re-invokes `tbd watch` in a tight loop forever — the usage error returns immediately, so the `while true` spins hot instead of failing on the first iteration. (Unknown flags are commander errors with exit 1 and are handled correctly.)
Fix (pick one): (a) re-code watch's timeout as exit 3, mirroring `tbd changes`' "no matching changes" exit 3 — the "nothing happened" family stays distinct from usage errors on every command; update spec, docs, tests, and the recipe; or (b) keep the contract and harden the recipe: on exit 2, `continue` only if stderr was empty, otherwise abort.

**N2 (Medium, operational robustness) — one transient network failure kills an unattended watch.**
Every poll tick spawns `git ls-remote`; any single failure — DNS blip, TLS hiccup, transient 5xx from the host — propagates out of `getRemoteTip`/`fetchRemoteTip` (`packages/tbd/src/file/bead-watch.ts:103-138`) and exits the watch with code 1, and the daemon recipe exits on 1. A watch designed to wake agents over hours violates error-handling Principle 8 (classify transient vs permanent) at exactly the layer where it matters most: an hour-3 blip silently ends all future wakes. Startup failures (validation, missing branch, missing baseline) should stay fail-fast; this is only about steady-state polling.
Fix (pick one): bounded consecutive-failure retries with backoff inside the poll loop (K consecutive errors → exit 1, single successes reset the counter); or keep the code as-is and change the shipped daemon recipe to restart on exit 1 with backoff and a retry cap, with a line of rationale.

**N3 (Low, docs placement) — the S2 guidance landed in the spec, not where users troubleshoot.**
The prior review's S2 asked for one troubleshooting line in `watch-beads`/tbd-docs ("after a sync-branch history rewrite, watch exits 1 with the not-an-ancestor error; restart without `--since`"). The fix commit added that sentence to the plan spec only (`docs/project/specs/active/plan-2026-07-19-bead-watch-and-external-sync.md`); neither `watch-beads.md` nor the tbd-docs watch section mentions it, and the PR checklist counts S2 as fixed. The error message itself (`Baseline <sha> is not an ancestor of tip <sha>`, `packages/tbd/src/file/sync-branch-changes.ts:290-294`) also carries no recovery hint, unlike the neighboring missing-branch path which does.
Fix: add the one-line troubleshooting note to `watch-beads.md` (and/or the tbd-docs watch section), and consider appending "restart the watch without --since to establish a new baseline" to the not-an-ancestor error.

**N4 (Low, UX of the R1 fix) — `--bead` startup validation against a stale local snapshot rejects real remote beads without a hint.**
Without `--since`, watch validates explicit IDs against the local committed sync branch before polling (`packages/tbd/src/file/bead-watch.ts:91-102`). A bead created remotely after the user's last `tbd sync` fails with `Unknown issue ID: <id>` (`packages/tbd/src/lib/issue-changes.ts:304-333`) even though watching it is exactly what the user wants, and the message doesn't say the snapshot may just be stale. The author's PR #197 revision (startup fetch + tip-to-tip report) resolves this structurally.
Fix: append "if it was created recently, run `tbd sync` first" to the unknown-ID error on this path, or adopt the #197 startup-fetch approach when the branches reconcile.

**N5 (Low, avoidable work) — the `--since <current-tip>` fast path reads the same snapshot twice.**
When `--since` equals the observed remote tip, the initial comparison calls `createChangesReportFromRefs` with identical endpoints (`packages/tbd/src/file/bead-watch.ts:172-181`), which reads and parses the full snapshot twice in `Promise.all` (`packages/tbd/src/file/sync-branch-changes.ts:295-298`) to produce a guaranteed-empty report. Also, `readyIssueIds` is computed for both endpoints on every report even for `all`/`beads` selections that never consult them (`packages/tbd/src/lib/issue-changes.ts:395-396`).
Fix: short-circuit identical endpoints to an empty report after resolving refs; compute ready sets lazily or only for `filter` selections with `ready`.

- **Myers diff worst-case memory.** `diffTextLines` stores a snapshot of the furthest-x map per edit distance (`packages/tbd/src/lib/issue-changes.ts:155-179`), which is O(D²) for a fully rewritten field — a 1,000-line rewritten `notes` field allocates on the order of millions of map entries during one diff. A distance cap that falls back to a whole-field replace hunk would bound this while preserving the PR's zero-dependency stance.
- **One exit-code table.** `changes` (0 = changes, 3 = none), `watch` (0 = change, 2 = timeout, 1 = error), and the repo-wide usage-error 2 are each documented separately; a single table in the tbd-docs sync section would make the N1 collision visible at a glance and give agent authors one thing to read.
- **Managed skill listings.** The generated `SKILL.md` shortcut tables (`.claude/skills/tbd/SKILL.md`, `.agents/skills/tbd/SKILL.md`) don't yet list `watch-beads`; presumably regenerated at release/setup time, but worth confirming it happens before this ships.

### False positives / verified — do not fix

- **All six prior findings are implemented and test-pinned.** R1: startup validation via `validateBeadSelectionAtRef` plus CLI test `rejects an unknown --bead before entering the poll loop` (exit 1 even with the remote branch absent). R2: `removeStaleWatchRefs` PID sweep at startup plus `finally` deletion; the reclaim test injects liveness so there is no PID-collision flake, and the real sweep's EPERM handling errs on the safe side (keeps the ref). R3: one `git cat-file --batch` per snapshot endpoint; the framing parser fails loud on missing blobs, truncation, and trailing bytes. R4: hunk context bounded to three lines with deterministic range merging, pinned by two tests. R5: null→null suppression on created/deleted beads, pinned. R6: empty `--spec` treated as absent in `list`/`changes`/`watch`, pinned by `change-selection.test.ts`.
- **The merge is semantics-preserving.** `readyIssueIds()` agrees with the old inline `ready` logic on all inputs (dangling blocker references remain impossible by construction — blocker IDs are the iterating collection's own IDs); main's conflicting hunk was brace-only. The re-wrapped help transcript is byte-exact per the passing tryscript goldens. The `list.ts` shared-filter refactor maps the old truthiness semantics exactly, including the R6 empty-string guard.
- **Git safety sentinels are comprehensive.** The e2e safety test byte-compares `FETCH_HEAD`, asserts the local sync branch and remote-tracking ref SHAs are unchanged, places sentinels in the hidden data-sync worktree and lock directory, and asserts no `refs/tbd/watch/*` remain after completion.
- **Empty stdout on timeout in `--json` mode is the contract, not a bug** — exit 2 is the signal, tested by `exits 2 with no stdout when the timeout elapses`.
- **Injection safety:** refs pass through `--end-of-options` and exact refspecs; all git invocations are spawn-arg arrays with a scrubbed environment (`gitSafeEnv`), no shell interpolation.
- **`--timeout 0 --since <commit>` one-shot comparison** remains coherent and covered, as in the prior review.

### Design assessment

Concur with the prior review's assessment of the architecture (pure diff core under an imperative shell with injected poll/clock/fetch boundaries, private collision-resistant refs, stateless `--since` chaining). Two additions from this round: the batch object reader is the right shape for R3 — strict framing rather than trusting `git cat-file` output, one process per endpoint, bounded buffering with an explicit 50 MB cap; and the edge-triggered `--ready` selection composed with the baseline-advance loop remains the subtlest logic in the PR and is the best-tested. The one design wrinkle worth deciding deliberately is the exit-code family split across the two commands (N1): `changes` chose 3 for "nothing", `watch` chose 2, and 2 is already the repo-wide usage-error code.

### Documentation

Spec and implementation are consistent after `69145dc` (batch reads, bounded hunks, created/deleted field suppression, ref reclamation, and the R2 CHANGELOG rewording all match the code). README, CHANGELOG, tbd-docs, help transcript, and the validation doc are updated and accurate at the merged head. Gaps: N3 (S2 guidance is spec-only), plus the two doc suggestions above.

### CI status

All checks green at `53e8edd` (post-merge): Test (macOS/Ubuntu/Windows), Coverage & Lint, Benchmark, DeepSource Secrets. PR remains draft; `mergeable_state` is clean after the merge with main.

---

jlevy:
Addressed the round-2 review findings in `f71b1cf`:

| Finding | Disposition |
| --- | --- |
| N1 — timeout/usage exit-code collision | Fixed, option (a). Watch timeout now exits 3 via a shared `EXIT_NO_MATCHING_CHANGE` constant used by both `changes` and `watch`, so exit 2 always means a usage error. Spec, tbd-docs, the `watch-beads` daemon recipe (`-eq 3`), and the `--timeout` help text updated; CLI test pins exit 3. |
| N2 — one transient poll failure kills an unattended watch | Fixed. An established watch tolerates a bounded run of consecutive failed remote polls (each failed poll waits the normal interval, counter resets on success) before exiting 1 with the last cause; startup validation and the first remote read still fail fast. Two unit tests pin ride-out and abort-at-cap. Spec, tbd-docs, and watch-beads note the behavior. |
| N3 — S2 guidance was spec-only | Fixed. The restart-without-`--since` troubleshooting line now appears in `watch-beads.md` and the tbd-docs watch section, and the not-an-ancestor error itself carries the recovery hint; test updated to pin it. |
| N4 — stale-local `--bead` rejection lacked a hint | Fixed. Startup validation failures against the local snapshot now append "If the bead was created since the last sync, run tbd sync first" (scoped so the missing-branch path does not double-hint); CLI test pins it. PR #197's startup-fetch approach remains the structural fix when the branches reconcile. |
| N5 — identical-endpoint double read; unconditional ready sets | Fixed. `createChangesReportFromRefs` reads one snapshot when `since === tip` (selection validation still runs, pinned by a new test including the unknown-ID case), and ready sets are computed only for `--ready` selections. |
| S-A — Myers diff worst-case memory cap | Deferred — tracked as `tbd-35nh`. |
| S-B — consolidated exit-code table | Fixed. New "Change-command exit codes" table in tbd-docs. |
| S-C — SKILL.md tables missing watch-beads | Fixed. Row added to both managed skill listings, matching generator output. |

All checks green at `f71b1cf`: Test (macOS/Ubuntu/Windows), Coverage & Lint, Benchmark, DeepSource Secrets. Full suite is 98 files / 1,427 tests.

---

jlevy:
## Superseded by #197 — please don't merge this one first

Heads-up before this lands: this branch and #197 had both been running review rounds on the same watch implementation in parallel (here: `69145dc` "harden" + `f71b1cf` N1–N5; there: `52867bd` R1–R6), which left two independently evolved forks of one feature — ~627 diverged lines, with the same behavior under different names (`removeStaleWatchRefs` here, `sweepStaleWatchRefs` there). Merging them conflicted in 14 files, so whichever landed second would have silently reverted the other's fixes.

At the repo owner's direction, **#197 is now the superset**. This branch's lineage won wholesale for every overlapping source file, test, and command doc — N1's `EXIT_NO_MATCHING_CHANGE`, N2's bounded poll failures, N5 plus the `git cat-file --batch` snapshot reads, and pre-poll `--bead` validation are all carried over intact.

`git merge-base --is-ancestor` confirms every commit here is now contained in #197 (0 outstanding). #197 is green and `MERGEABLE`.

One thing from here did **not** survive as-is, and is worth knowing regardless of which PR lands: `cleanup` still throws from inside a `finally` in this branch's `bead-watch.ts`, so a failed `update-ref -d` replaces a successful `{kind:'changed', report}` return — the watch detects the change the caller was blocking on, then exits 1 without reporting it. #197 makes it best-effort, matching the reclaim path. If this branch is ever revived independently, that fix needs to come with it.

Suggest closing this as superseded once #197 lands.

---

PR #197, opened 2026-07-20
process: plan Linear ↔ bead bidirectional sync pilot + tbd watch foundation

jlevy:
## Summary

Planning/docs-only PR: a refined, testable plan for bidirectional sync between a linked subset of beads and Linear, composed with the first slice of `tbd watch` for agent coordination — plus consolidation of the research docs that back it.

**New spec: `docs/project/specs/active/plan-2026-07-20-linear-bead-sync-pilot.md`**
- `linked` field on beads (provider + UUID, per design doc §8.7) with a **one-external-source-per-bead invariant**: CLI link guard + merge collapse rule (newest `linked_at` wins, loser to attic), so every sync is one bead↔tracker pair against one base — never multi-source reconciliation.
- **Generic `tbd bridge` command group** (`link/unlink/import/sync/status`): one `tbd bridge sync` covers all configured providers; provider inferred from refs; shared `bridges:` config block. Linear is the only adapter in the pilot; GitHub is designed to be a mechanical second adapter (every provider-facing seam ships generic).
- Bridge state on the `tbd-sync` branch with per-link base snapshots → true 3-way field-level diffs, echo suppression (no self-write loops with plain API-key auth), and LWW+attic conflict handling consistent with the existing merge engine.
- Field mapping tables: status via Linear `WorkflowState.type` (with `blocked`/`deferred` round-tripping via paired labels), bijective priority map, labels/title/description; polling transport (Linear webhooks require a public HTTPS endpoint, so they're a later daemon phase).
- `tbd watch --json` foundation: JSONL bead-change events with `last_actor` attribution for anti-recursion, documented dispatch conventions (assignee=agent, label=mode, status=lifecycle), and a QA playbook proving the loop: Linear issue → import → watch wakes agent → bead closed → Linear shows completed.
- Phased plan with prerequisites: fix `extensions` merge (`lww` → `deep_merge_by_key` per design §3.5 — current whole-object LWW at `packages/tbd/src/file/git.ts:407` can silently drop a namespace) and a `tbd_format` compatibility gate (Zod strip mode means older CLIs silently destroy new frontmatter fields on rewrite).

**Research consolidation**
- `api-references-bridge-integrations.md`: new §5 with Linear API facts verified 2026-07-20 (GraphQL endpoint/auth, 5,000 req/hr + complexity limits, webhook delivery constraints, Agents platform / `AgentSession` delegation, priority + workflow-state model); cross-links to the monitors/kernel/orchestration briefs and the new spec.
- `research-2026-06-04-agent-issue-monitors.md`: next-steps checked off with an addendum noting the extension sketch graduated into this spec.

**Tracking** (synced on `tbd-sync`, not in this diff): epic `tbd-g305` with dependency-wired phase beads (`tbd-le2l`, `tbd-80wy`, `tbd-fjr0`, `tbd-klb6` → `tbd-7czg` → `tbd-atv5` → `tbd-ii8p`, deferred `tbd-bbv0`); stale `tbd-7am3` closed.

## Context

Review found the pre-rewrite GitHub sync epic (`tbd-68cw`, closed 2026-02) did not survive the v2 rewrite — current tbd has no external tracker support — so this spec is the new plan of record for external tracker sync, carrying forward that epic's lessons (4-phase sync ordering, status mapping tables, CLI gating pattern).

https://claude.ai/code/session_014hwBVi9rmUqsmP8DE6FCLE

---
_Generated by [Claude Code](https://claude.ai/code/session_014hwBVi9rmUqsmP8DE6FCLE)_

---

> [!NOTE]
> **Medium Risk**
> New remote-polling CLI path and a large contract surface (selectors, baselines, private fetches); mitigated by read-only design and broad tests, but agents will depend on correct exit-code and JSON semantics.
> 
> **Overview**
> **Ships Phase 1 bead watching** and documents the Linear ↔ bead pilot that builds on it.
> 
> **CLI:** Adds read-only **`tbd changes`** (diff committed sync-branch snapshots with selectors and stable JSON) and **`tbd watch`** (poll remote tip, private-ref fetch, exit on match). Shared selection/filter logic is extracted for `list`/`ready`; human + JSON reports and gh-style exit codes. **`watch-beads`** shortcut and README/docs cover watch-then-spawn and in-session agent patterns.
> 
> **Safety:** Watch does not touch the working tree, hidden data-sync worktree, or sync refs; sweeps stale private refs from dead processes; **`--bead`** fails fast on unknown IDs.
> 
> **Planning (no bridge code yet):** New/updated specs for bead watch + external sync and the **Linear bidirectional sync pilot** (`linked` field, `tbd bridge`, 3-way echo suppression, polling-first). Research brief gains **Linear API §5**; agent-monitor research is reconciled with the pilot plan. Agent hook fallbacks bump **`get-tbd@0.4.1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31314e21e5ca741e6723a765acda5716da1a31d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy:
## Senior Engineering Review — PR #197

**Scope:** full branch review (`tbd shortcut review-code`, GitHub PR scope) plus a merge of `origin/main`, run after the branch went `CONFLICTING`. Reviewed the 12 branch commits, the two plan specs, the research consolidation, the shipped `changes`/`watch` implementation, and the bead tracking behind the epic.

**Verdict:** **Approve with changes — the changes are applied on this branch.** The design is sound and the implementation is well-factored. The merge with main was not clean, and two of the conflicts were *semantic* — invisible to git, and both would have shipped broken. Those are fixed, along with a contract ambiguity that made the project's own recommended agent loop spin forever on a typo.

**CI status:** locally, `pnpm run ci` (format:check → lint:check → build → test) is green at `06511a5` — **97 test files / 1425 tests** — and the pre-push gate (check / build / test / package-age) passed. The PR went from `CONFLICTING` to `MERGEABLE`; GitHub Actions (including the Windows and macOS matrix, which I can't reproduce here) is still running as I post this, and I'm watching it.

---

### Merge: `origin/main` → this branch

Main moved 20 commits ahead (v0.4.2 release, the TS/JS quality floor v2 lint overhaul, agent CLI ergonomics round 2). Seven files conflicted:

| File | Resolution |
| --- | --- |
| `.claude/{hooks,scripts}`, `.codex/*` | Generated agent surfaces differing only in the zero-install pin. Took main's `0.4.2`; the branch's `0.4.1` predates the release. |
| `packages/tbd/CHANGELOG.md` | Union — the branch's `Unreleased` above main's released `0.4.2`. |
| `ready.ts` | Took the branch's extraction into `readyIssueIds()`. Main's side was **purely** the new curly-brace rule applied to the old inline logic, so the refactor is the only semantic intent. |
| `cli-setup.tryscript.md` | Union of the branch's `changes`/`watch` entries with main's widened help column, then regenerated via `tryscript --update` — the regeneration produced a **zero diff**, confirming the hand-merge byte-for-byte. |

---

### Findings

#### R1 — `--label` crashes on first use for both new commands · Blocker · **fixed**

`packages/tbd/src/cli/commands/changes.ts:68`, `packages/tbd/src/cli/commands/watch.ts:79`

This is the semantic conflict the merge could not see, and it is a trap main already fell into and documented.

Main's new strict preset enables `@typescript-eslint/no-useless-default-assignment`, which **errors** on Commander's `(value, previous: string[] = []) => …` accumulator — the declared type isn't optional, so the default looks useless. Its autofix strips the `= []`. But Commander passes `undefined` on the first occurrence when the option has no default, so the stripped form throws. Main hit exactly this in `fd657f2` and fixed it in `89e3ff9` across four sites. These two new commands are the **last two sites**, written with the old form, so they land on the wrong side of that fix.

Verified on the merged build, before the fix:

```
$ tbd changes --since HEAD --label foo
Error: previous is not iterable
$ tbd watch --label foo
Error: previous is not iterable
$ tbd list --label foo          # control: fixed on main
Error: Not a tbd repository (…)  # ← parses fine, reaches the repo check
```

This also broke `cli-changes.test.ts:170` (usage error expected exit 2, got 1) — the failure that made the post-merge suite red.

**Fix:** applied main's idiom — accumulator typed `string[] | undefined`, spread `previous ?? []`. Honest under `strictTypeChecked` rather than hiding `undefined` behind a default parameter. `--label` is a headline selector for both commands, so this was 100% broken on first use.

#### R2 — timeout and usage errors shared exit code 2, spinning the documented agent loop · High · **fixed**

`packages/tbd/src/cli/commands/watch.ts:60`

`watch` exited **2** on timeout. `ValidationError` also exits 2, CLI-wide. And `watch-beads` ships the unattended loop as:

```bash
watch_status=$?
if [ "$watch_status" -eq 2 ]; then continue; fi   # ← retries usage errors forever
```

So a mistyped flag becomes an infinite tight loop that never wakes anyone. Verified both collide:

```
tbd watch --ready --interval 5   -> exit 2   ("must be at least 10 seconds")
tbd watch --bead x --label y     -> exit 2   ("cannot be combined")
tbd watch --all --timeout 0      -> exit 2   (timeout)   ← indistinguishable
```

**Fix:** timeout now exits **3** — the code `tbd changes` already uses for "nothing matched", which is exactly what a timeout is, just discovered by waiting. Both commands now share one contract, hoisted into `NO_CHANGES_EXIT_CODE` so they can't drift:

> **0** matched · **3** nothing matched · **2** usage error · **1** operational error

The documented loop retries on 3 and is now correct by construction. `watch` is unreleased (CHANGELOG: `Unreleased`), so this cost no compatibility — but it is the last cheap moment to make the change. Updated the manual, the shortcut prose and its loop, both spec statements of the contract, the CHANGELOG, and the timeout test; the existing usage-error test now carries a comment tying it to the retry loop so a re-collision fails with the reason attached.

#### R3 — ref cleanup could discard the change report it just produced · Medium · **fixed**

`packages/tbd/src/file/bead-watch.ts:143`

`cleanup` threw on a failed `update-ref -d`, and it runs in a `finally`. A throw there **replaces** a successful `{kind:'changed', report}` return — the watch would detect the change the caller was blocking on, then exit 1 without reporting it. For the watch-then-spawn loop that turns a successful wake into a crash.

Ref bookkeeping is already best-effort everywhere else: `sweepStaleWatchRefs` deliberately swallows both its listing and its delete failures ("sweep failures must never prevent the watch itself from running").

**Fix:** cleanup is now best-effort too, matching. A ref left by a failed delete is collected by the next watch's sweep once the PID is gone — the same path an interrupted watch already relies on. Cleanup still runs *through* `finally`, preserving the validation plan's requirement.

#### R4 — the extensions-merge prerequisite gated nothing · Medium · **fixed**

`tbd-le2l` — "Fix extensions merge: lww → deep_merge_by_key per design §3.5" — had **no dependency edge at all**, despite the PR body listing it as dependency-wired into Phase 1 and the spec listing it under "Phase 0 — prerequisites". `tbd ready` would have offered Phase 1 work with a documented data-loss hazard (whole-object LWW silently dropping a namespace, `git.ts:407`) still open.

**Fix:** wired `tbd-le2l → tbd-7czg`, matching its two Phase 0 siblings. Phase 1 is now correctly blocked by all three.

#### R5 — the reconciliation bead was still open after being delivered · Low · **fixed**

`tbd-fuiw` ("S1: Reconcile Phase 2 external-sync designs") sat `deferred` even though **§6a of the pilot spec is that reconciliation**, delivered in this PR (`d21a701`). **Fix:** closed with a reason pointing at §6a and the commit.

#### R6 — PR body and bead store disagree on `tbd-klb6` · Low · **your call, not changed**

The PR body lists `tbd-klb6` (last_actor) among the beads gating `tbd-7czg` (Phase 1). The store has it gating `tbd-ii8p` (Phase 3).

I did **not** override this, because unlike R4 it looks deliberate and it is defensible: Phase 3 is where the spec says `last_actor` is consumed. But the spec also files it under "Phase 0 — prerequisites", and §6 has the bridge stamping `last_actor: linear-bridge` on its own bead writes — which is Phase 1/2 behavior. So either the edge should move to `tbd-7czg` or the PR prose should be corrected. Worth a decision.

#### R7 — `readyIssueIds` diverges from the predicate it replaced, in unreachable territory · Informational · **not changed**

`packages/tbd/src/lib/issue-selection.ts:60`

The extraction treats a **missing** blocker as unresolved (`issueById.get(id)?.status !== 'closed'` → `undefined !== 'closed'` → blocks), where main's inline predicate treated it as resolved (`blocker && blocker.status !== 'closed'` → doesn't block).

Unreachable today: blocker IDs are always drawn from the same complete snapshot that builds the map, so the lookup always resolves — and the docstring already states the "one complete issue snapshot" precondition. Flagging so that stays a documented invariant rather than an accidental one; no change made, since adding an unreachable branch is its own cost.

---

### Design assessment

**The architecture holds up, and the sequencing decision was the right one.** Landing watch first as a working increment and treating the bridge as safe additions on top (§6/§6a) is what let this review find real bugs in shipped code instead of arguing about a design doc.

Specific strengths worth keeping:

- **The pure/impure split is the reason this is testable.** `issue-changes.ts` is a pure snapshot differ with no I/O; `bead-watch.ts` injects every boundary (`now`, `sleep`, `getRemoteTip`, `fetchRemoteTip`, `createReport`). Polling and deadline behavior are unit-testable without a clock or a network, and the 1425-test suite shows it.
- **Ref-injection is closed properly.** `resolveCommit` runs `rev-parse --verify --end-of-options` and resolves to a full SHA *before* anything downstream interpolates it into `git show <ref>:<path>`, so a hostile `--since` can't reach git as an option. This is the right shape, not an escaping hack.
- **Base-snapshot 3-way diffing is the correct echo-suppression primitive** for the bridge, and it works with plain API-key auth — no dependence on actor filtering, which is what makes the pilot viable without OAuth `actor=app`. The single-source-per-bead invariant genuinely eliminates multi-master reconciliation rather than deferring it, and the provider-generic seam (`TrackerAdapter`, `tbd bridge`, per-provider state layout) is cheap to ship now and expensive to retrofit.
- **Polling-first is honest**, not a shortcut: Linear webhooks need a public HTTPS endpoint, and polling is the reconciliation path a webhook daemon needs anyway.

**The one scaling limit I'd flag** (already tracked as `tbd-293h`, correctly deferred): `readSnapshot` spawns one `git show` per issue, for both endpoints — ~2N subprocesses per report, so ~390 for this repo's 193 issues. That's per *remote movement*, not per poll, so it's fine today. But the bridge's whole point is to increase how often beads move, and `tbd-atv5` folds bridge sync into `tbd sync`. I'd move `git cat-file --batch` ahead of the bridge phases rather than after, so the cost doesn't land at the same time as the traffic that exposes it.

**On the plans themselves:** the two specs are now genuinely reconciled rather than merely cross-referenced — §6a resolves naming (`mirror` → `bridge`), bindings (`linked` field over interim labels), echo suppression (base tuples *and* content hashes, complementary), and the writer model, and it supersedes its own earlier streaming-JSONL sketch in favor of the shipped blocking contract. That is the right kind of plan update: it records what reality did to the design. The `linked` field's array-with-invariant shape is a good call — §8.7 compatibility without paying for multi-master today.

### Documentation

Exit codes were stated in five places and are now consistent across all of them: the manual (`tbd-docs.md`), the `watch-beads` shortcut (prose **and** the loop), both spec statements (`plan-2026-07-19` §Exit codes and §JSON output), the pilot spec's §6 summary, and the CHANGELOG. The CHANGELOG's security note was also tightened, since "removes its private ref on normal exit" is now explicitly best-effort.

The pilot spec cited `packages/tbd/src/file/git.ts:407` twice for the `extensions: 'lww'` line that Phase 0 has to change; main's lint overhaul moved it to **413**. Both references updated — a line-number citation that silently drifts is worse than none, since the next agent trusts it.

No dependencies were added or upgraded by this branch or by these fixes, so the supply-chain rules in `SUPPLY-CHAIN-SECURITY.md` are satisfied trivially.

### Tracking

Beads are synced to `tbd-sync` (`efab8df`) and level with origin. Epic `tbd-g305` now reads:

```
tbd-le2l ┐
tbd-80wy ├─→ tbd-7czg ──→ tbd-atv5 ──→ tbd-ii8p
tbd-fjr0 ┘                             ↑
                            tbd-klb6 ──┘   (see R6)
tbd-bbv0  deferred (Phase 4)
tbd-fuiw  closed (delivered as §6a)
```

---

### What landed on this branch

| Commit | |
| --- | --- |
| `60e451b` | Merge `origin/main` — 7 conflicts resolved, incl. the R1 semantic conflict |
| `94fd2eb` | R3 — cleanup can no longer discard the change report |
| `87ecab0` | R2 — timeout exit code split from usage errors |
| `06511a5` | Stale `git.ts` line citations in the pilot spec |

Bead changes are on `tbd-sync` (`efab8df`), not in this diff: `tbd-le2l → tbd-7czg` wired (R4), `tbd-fuiw` closed (R5).

**Open for you:** R6 (the `tbd-klb6` edge — move it to Phase 1, or correct the PR body).

---

jlevy:
## Follow-up: PR #196 consolidated into this PR

While verifying that the plans and beads were in sync, I found the thing that mattered most, and it wasn't in the diff: **PR #196 was still open against `main` and had moved 6 commits ahead of the copy of its work living here.**

Both PRs had been running review rounds on the same feature, in parallel:

| | this PR (#197) | PR #196 |
| --- | --- | --- |
| Fix round | `52867bd` "R1–R6" | `69145dc` "harden" + `f71b1cf` "N1–N5" |
| Base | main @ PR #203 | main @ PR #199 |

That produced two independently evolved forks of one feature — ~627 diverged lines, with the same behavior implemented under different names (`sweepStaleWatchRefs` here vs `removeStaleWatchRefs` there). Merging them conflicted in 14 files. Whichever PR landed second would have silently reverted the other's fixes.

Per your call, **#197 is now the superset** (`150c542`).

### How the two lineages were reconciled

**#196's implementation wins wholesale** for every overlapping source file, test, and command doc — it is the further-along lineage:

- **N1 — timeout exit code.** #196 had already reached the *identical* conclusion I did independently (R2 in my review above): a timeout sharing exit 2 with usage errors makes the documented wake loop spin on a typo. #196 puts the constant in a dedicated `exit-codes.ts` (`EXIT_NO_MATCHING_CHANGE`), which is a better home than mine was. Its version stands; I deleted my `NO_CHANGES_EXIT_CODE` rather than leave two spellings of one contract.
- **N2** — an established watch now rides out a bounded run of failed remote polls, so a brief outage doesn't kill an unattended watch.
- **N5 + batched blob reads** — snapshots load through one `git cat-file --batch` per endpoint instead of one `git show` per issue, and ready sets are computed only for `--ready`. **This supersedes the O(2N) subprocess concern I raised in the design assessment above** — it's already fixed on this branch now, not deferred.
- **Static `--bead` validation before polling**, replacing this branch's startup-fetch approach to the same fail-fast goal.
- **R1 (`--label` accumulator)** — both sides fixed it identically, which is a good independent confirmation that the diagnosis was right.

**Carried forward from this branch**, because #196 doesn't have them:

- **The cleanup fix (R3).** #196's `cleanup` still throws from inside a `finally`, so a failed `update-ref -d` would discard the change report the watch had just produced. Re-applied on top of its version — this was the one code finding of mine that survived the consolidation as genuinely new.
- The Linear pilot spec, the §6a reconciliation, the research consolidation, and the 07-19 Addendum — all unique here.

Conflicting prose resolved toward #196's wording where both sides said the same thing (spec exit codes); the CHANGELOG security note takes #196's wording plus the best-effort clause the cleanup fix requires.

### State

- `git merge-base --is-ancestor` confirms **#196 is now fully contained in #197** — 0 commits outstanding. #196 can be closed as superseded once this lands.
- Verified locally: `format:check` → `lint:check` → `build` → **98 test files / 1429 tests** (up from 97/1425; #196's tests came along), plus the pre-push gate.
- PR is `MERGEABLE`; GitHub Actions running.

### Revised standing findings

R1 and R2 from my review above are resolved by #196's equivalents; R3 stands and is applied. **R4/R5** (bead wiring — `tbd-le2l → tbd-7czg` wired, `tbd-fuiw` closed) are unaffected and remain on `tbd-sync`. **R6** is still open for you: the PR body lists `tbd-klb6` as gating Phase 1, the store gates Phase 3.

One process note worth more than any single finding here: this divergence was invisible from inside either PR. Both looked healthy on their own. If watch work continues across more than one branch, landing the shared foundation to `main` first — rather than folding a copy into a downstream branch — is what prevents a second round of this.

---

packages/tbd/src/cli/lib/exit-codes.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Shared CLI exit codes for the bead change commands.
+ *
+ * Exit 2 is reserved repo-wide for usage errors (see ValidationError in
```
jlevy, line 4: If other codes have meaning, why aren't they all here?

jlevy:
# Senior Engineering Review, round 2 — design focus

**Scope:** fresh review of PR #197 at `563030b` (`tbd shortcut review-github-pr` → `review-code`, GitHub PR scope), after the #196 consolidation. Round 1 covered correctness and merge semantics; per maintainer request this round examines the **broad design**: whether the plan could be simpler, built on smaller primitives with more flexibility, and how to keep the risk to existing tbd workflows as low as possible while making tracker integrations easy to experiment with.

**CI:** all checks green at `563030b`; PR is `CLEAN`/`MERGEABLE`. No unresolved findings from earlier rounds except R6 (noted below).

**Verdict: the shipped half of this PR is the architectural model; the planned half should be resequenced to match it.** `tbd changes`/`tbd watch` landed as read-only, additive, zero-schema-impact primitives composed from userland — that is exactly "simple things simple." The bridge plan, as spec'd, couples an *unvalidated* integration into the four highest-blast-radius parts of core — issue schema, config schema, sync layout, and the format version — before any Linear code has run. Nearly all of that coupling can be deferred until after the pilot proves the model, without giving up any of the pilot. The findings below are decision points, not defects; the plan is buildable exactly as written.

| ID | Severity | Finding |
| --- | --- | --- |
| D1 | High | Top-level `linked` field forces a repo-wide format gate; `extensions` gives the same pilot with zero forced upgrades |
| D2 | High | The `bridges:` config block *also* triggers the format-bump policy — the spec's rollout doesn't account for it |
| D3 | High | The bridge can run as userland over three small core primitives; in-core `tbd bridge` can wait for evidence |
| D4 | Medium | Pilot syncs all fields bidirectionally from the start; import + status-writeback covers the stated loop at a fraction of the machinery |
| D5 | Medium | Folding bridge sync into `tbd sync` couples external APIs into the one workflow every agent runs — keep it out for the pilot |
| D6 | Low | Provider-generic internals (ref inference, registry) are speculative before provider #2; the generic *command name* is worth keeping |

---

## The coupling inventory

What the pilot spec asks core to absorb **before** the first Linear round-trip has ever run:

1. **Issue schema** — new top-level `linked` and `last_actor` fields (spec §1, Phase 0)
2. **Format version** — a `tbd_format` bump so pre-pilot CLIs refuse to write (spec Rollout §1)
3. **Config schema** — a `bridges:` block in `.tbd/config.yml` (spec §5)
4. **Sync layout** — a new `bridge/` directory on the sync branch beside `issues/` and `mappings/` (spec §2; `paths.ts:22` enumerates the current normative layout)
5. **Sync path** — external phases inside `tbd sync` ordering (spec §4)
6. **CLI surface** — a five-subcommand `tbd bridge` group with provider inference (spec §5)

Each is individually defensible. Together they put the whole experiment on core's release train and force every user of a piloting repo to upgrade in lockstep. The findings below take them in turn.

## D1 — `linked` vs `extensions`: the format gate is the real cost · High

The spec presents the top-level `linked` field as following design doc §8.7. But §8.7 is explicitly neutral — *"Issues could have a `linked` field **(or use `extensions`)**"* — and the design doc's `extensions` documentation (tbd-design.md:1458) actively designates that namespace for this exact use, with a worked example that is literally external-tracker linkage:

```yaml
extensions:
  github:
    issue_number: 123
    synced_at: 2025-01-07T10:00:00Z
```

The compatibility math differs sharply between the two options, and it's verifiable in the schema today:

- `IssueSchema` uses Zod strip mode (`schemas.ts:279` states the policy): an older CLI rewriting a bead **silently destroys** any unknown top-level field. That is why the spec's Rollout must gate the repo — every agent and human must upgrade before the repo can link a single bead.
- `extensions` is a **known** field to every existing CLI (`schemas.ts:76`, `z.record(z.string(), z.unknown()).optional()`): data inside it round-trips through old CLIs on rewrite, no gate needed.

The one real hazard for extensions-based linkage is the merge: the implementation does whole-object LWW (`git.ts:413`, `extensions: 'lww'`), so concurrent writers to *different* namespaces can drop one side. But design §3.5 (tbd-design.md:2343) already specifies `deep_merge_by_key` — the implementation diverged from the design. **Phase 0's extensions-merge fix is therefore an independently-justified bug fix**, worth landing regardless of any bridge, and once it lands, `extensions.bridge.linked` is merge-safe.

**Recommendation:** pilot with linkage under `extensions` (single entry, single-source invariant enforced by the linking CLI plus a `doctor` check). Promote to a first-class `linked` field — with one format bump and a mechanical migration — only after the pilot validates the model. The trade-off honestly stated: a top-level field is cleaner long-term, displays without a join, and the gate machinery exists and works. But sequencing standardization *after* validation converts "all users of this repo must upgrade now for an experiment" into "nobody notices until we've decided this is real." If the pilot dies, nothing was standardized; if it succeeds, the migration is one rename with evidence behind it.

Same reasoning applies to `last_actor`: as spec'd it is a second gated schema field. Under extensions it's `extensions.bridge.actor` or simply a documented label convention until the coordination work needs more.

## D2 — the `bridges:` config block also triggers the gate policy · High

`ConfigSchema` is also strip-mode, and the policy comment above it (`schemas.ts:279`) is explicit: *"When changing config schema (adding, removing, or modifying fields), **ALWAYS bump the format version**."* An old CLI doesn't crash on an unknown `bridges:` block — worse, it strips it and **destroys the block on its next config rewrite** (`tbd setup`, migrations).

The spec's Rollout §1 gates only for the issue-schema fields and describes Phases 1–2 as "enabled-by-config-only… no behavior change for repos without the block." Under the repo's own documented policy, that's incomplete: the config block *itself* requires the bump. Two options:

- Accept it and fold the config change into the same single format bump (if D1's route is rejected), or
- Keep bridge configuration **out of `.tbd/config.yml`** for the pilot — its own file (e.g. `.tbd/bridge.yml`) that core never parses, plus env vars for secrets. Old CLIs can't destroy what they never read. This falls out for free under D3.

Either way the spec needs the gap closed; right now a piloting repo with an old-CLI agent would silently lose its bridge config.

## D3 — the broad-design alternative: userland bridge over small core primitives · High

This is the heart of the "simpler primitives, more flexibility" question. The shipped watch foundation plus the existing CLI already provide almost everything a bridge needs — from the *outside*:

| Bridge need | Already available |
| --- | --- |
| Detect bead changes | `tbd changes --since` / `tbd watch --json` (this PR, shipped) |
| Read beads | `tbd show --json`, `tbd list --json` |
| Write beads | `tbd create` / `update` / `close` + `tbd sync` |
| Detect Linear changes | Linear `updatedAt`-filtered polling (spec §5 research, verified) |
| Echo suppression | The bridge's **own** base snapshots in its **own** state file |
| Serialization | Data-sync lock already serializes writers via `tbd sync` |

What is genuinely missing from core is small, and each piece is independently justified:

1. **The §3.5 extensions merge fix** — a bug fix against the written design (D1).
2. **Generic extensions read/write on the CLI** — today extensions is a dead letter: nothing in `create.ts`/`update.ts` can set it and nothing in `show.ts`/`issue-format.ts` displays it (verified — zero references). Something like `tbd update <id> --set-extension bridge='{...}'` plus surfacing it in `show`/`show --json`. This is the *actual* missing primitive, and it serves every future tool, not just Linear — it's what tbd-design.md:1458 promised extensions would be for.
3. **(Optional) actor attribution** — per D1, an extensions field or label convention during the pilot.

With those, the entire Linear pilot — link, import, 3-way sync with base snapshots, status/priority mapping, orphan detection — runs as an **external script or separate package** (`tbd-bridge-linear`, or a script in this repo run by cron/CI), keeping its state in its own committed file. Core's release cadence, schema, sync path, and format version are untouched. A Jira or GitHub experiment becomes something anyone can write *tomorrow* without a tbd release. The project's own research points here: the coordination-kernel brief's "bridges as thin, stateless adapters" over the watch surface, and OpenAI Symphony — cited in the spec as validating polling-first — is precisely an external orchestrator polling Linear with no tracker code inside the tool it orchestrates.

Two real costs, stated fairly: **(a)** UX — no single `tbd bridge sync`; the pilot invocation is `node bridge-linear.mjs` or a cron entry. For a pilot run by the maintainer, that seems acceptable; `tbd bridge` becomes the *productization* step once shape is proven. **(b)** Bridge state doesn't ride the sync branch — spec §2's `bridge/` dir needs core sync-layout changes anyway (`paths.ts` enumerates `issues/`, `mappings/`, `attic/` only), so during the pilot, state-as-committed-file (or a branch the script manages with plain git) is simpler than teaching core a new directory. Shared-state-on-sync-branch is another thing to standardize *from evidence*.

This also dissolves the Phase 0 → Phase 1 gating problem from round 1 (R4/R6): with no schema change, the only hard prerequisite left is the merge fix.

## D4 — pilot field scope: full bidirectional LWW vs import + status writeback · Medium

The spec's own success loop (§6, Goals) is: *PM files in Linear → import creates linked bead → agent works it → close → Linear shows completed.* That loop needs **import (pull-create) plus status push** — nothing else. Yet Phase 1 already implements per-field 3-way diffing, and the hard open questions all live in the fields the loop doesn't need: description LWW granularity and markdown-flavor normalization (Open Question 4), label collision naming (OQ 5), title conflicts, attic entries for text fields.

Status is the *least* conflict-prone field — a five-value enum with a bijective-ish mapping table — and one-way-per-direction flows (create inbound, status outbound-and-inbound) need only a trivial last-sync marker, not full base-tuple snapshots. Base snapshots are the right mechanism **for the fields that need them**; let them arrive with those fields in a later phase. This roughly halves the pilot's mapping tables, conflict machinery, and convergence-test surface while still delivering the demo that decides whether the whole direction is right.

## D5 — keep bridge sync out of `tbd sync` for the pilot · Medium

`bridges.sync_on_tbd_sync` folds external network calls into the single most-executed workflow in every agent session. Even opt-in and failure-contained, it means routine syncs acquire new latency modes, rate-limit coupling, and "degraded" report states that agents must learn to interpret. The spec already treats CI/cron as the steady state (§6a writer model); sharpen that: for the pilot, bridge sync is **only** a separate invocation, and folding into `tbd sync` is a post-pilot decision with its own risk review. Non-goal it explicitly.

## D6 — provider-generic internals can wait for provider #2 · Low

The four-method `TrackerAdapter` is small and reasonable to write from day one. But ref inference across providers, the shared `bridges:` multi-provider config shape, and per-provider state registries are generalizations with exactly one consumer. Classic rule-of-three: extract the seam when GitHub actually lands — the spec itself argues GitHub is mechanical, which cuts both ways: if it's mechanical later, nothing is lost by not pre-building for it now. The user-facing *name* `tbd bridge` (vs `tbd linear`) is the one generic choice worth locking early, since command renames break users — and that choice is free.

---

## Code-level pass

Rounds 1–2 covered the shipped code in detail; the consolidated `563030b` state was re-checked against the general/TS guidelines this round. Nothing new material. Notes for the record: constants are named and documented where the rules require (`EXIT_NO_MATCHING_CHANGE` with its rationale in `exit-codes.ts`, `MAX_CONSECUTIVE_POLL_FAILURES`, `TEXT_HUNK_CONTEXT_LINES`, the Windows test timeout); the exit-code contract is written down at its definition site and pinned by tests, which is exactly what error-handling-rules asks of exit codes as API contracts. `show --json` in bulk mode spreads the full issue object, so extensions data would already surface there — single-ID `show` projects a display shape and would need the D3(2) work to expose it.

**Still open from round 1:** R6 — the PR body says `tbd-klb6` (last_actor) gates Phase 1; the bead store gates Phase 3. Under D1/D3 this resolves itself (the field moves out of the gated path entirely), which is one more small argument for the resequencing.

## Documentation

- If any of D1–D5 is adopted, the pilot spec's §1/§2/§5 and Rollout need revising; if **rejected**, the spec should still record the extensions-vs-linked decision as a considered trade-off — right now it cites §8.7 as settling a question §8.7 explicitly leaves open, and D2's config-gate gap needs closing regardless.
- The 07-19 spec's Status line still reads "PR #196 pending"; after the consolidation it should point at this PR.

## Summary

The watch foundation shipped in this PR demonstrates the project's best pattern: additive read-only primitives, stable JSON + exit-code contracts, composition in userland. Applied to the bridge, the same pattern gives a concrete resequencing: **land the extensions merge fix and generic extensions CLI access (small, independently right), run the Linear pilot as an external bridge with its own state, and only then standardize** — the `linked` field, the `bridge/` sync layout, the `tbd bridge` command, and one format bump, all justified by a working pilot instead of a plan. Simple things stay simple, complex things stay possible, and no existing workflow can tell the experiment is happening until it has earned its place in core.

---

jlevy:
## Watch infrastructure extracted to #205

Following the round-2 design review above, the provider-neutral watch infrastructure has been extracted from this PR onto a clean base from `main`: **#205** — the same reviewed, hardened code (content-extracted from `563030b`, all three review rounds included), with the spec rescoped to an Integration Layer architecture and zero schema/config/format coupling.

What remains unique to this PR after #205 lands: the Linear pilot spec (`plan-2026-07-20-linear-bead-sync-pilot.md`), the Linear API research (§5 of the bridge-integrations brief), and the monitors-research addendum. Those are tracked for rework under the layering rules — extensions-first bindings, separable module, no format gate — as `tbd-vm5s`, before any implementation.

Suggested sequencing: land #205 → close #196 (already fully contained) → rework the pilot spec per `tbd-vm5s`, at which point this PR either shrinks to the reworked spec or closes in favor of a fresh spec-only PR.

---

jlevy:
replaced by #206

PR #205, opened 2026-08-06
feat: bead watch infrastructure (tbd changes + tbd watch), extracted as a clean provider-neutral landing

jlevy:
## Summary

This PR lands provider-neutral, opt-in infrastructure for watching committed tbd issue state. It stays independently mergeable from any Linear integration while providing the stable primitive needed to experiment with one end to end.

### Core contract

- **`tbd changes --since <commit>`** is a pure, deterministic diff over committed sync-branch snapshots. It supports bead, label, spec, status, ready, and all selectors using the same predicates as existing list and ready workflows.
- **`tbd watch`** polls the remote sync tip, fetches movement into a collision-resistant private ref, emits one matching report, and exits. It does not mutate the caller worktree, hidden sync worktree, configured sync refs, remote-tracking refs, or `FETCH_HEAD`.
- JSON reports include `since`, `tip`, and deterministic per-bead/per-field changes. The report is command output, not repository state, so it carries no format marker: like every other tbd `--json` surface it evolves by addition only, and `tbd_format` (`fNN`) stays the single format code for on-disk state.
- Stable CLI exit codes are centralized: success 0, operational error 1, usage error 2, no matching change or timeout 3, and SIGINT 130.
- The **`watch-beads`** shortcut provides a durable watch-then-worker recipe plus in-session guidance for Claude Code and Codex.

## Review and release hardening through dc0ba559

All findings from the senior review and inline threads are addressed:

- Final and partial timeout intervals perform an inclusive remote observation, so a boundary change is not dropped.
- Every network-facing Git process has a positive wall-time limit. Operations started before the watch boundary use the remaining watch budget; the explicit final observation gets one bounded poll budget, capped at 30 seconds. A fetch and its private-ref resolution share one deadline. Startup, established poll, fetch, resolution, and cleanup paths have regression coverage.
- Text diffing has a deterministic edit-distance cutoff; larger rewrites retain complete before/after JSON and set `hunks_omitted` to `complexity_limit`. Static selection runs before expensive field deltas.
- Git object reads use bounded 128-object `cat-file` batches instead of one graph-sized output buffer.
- The substantive issue-field order is compile-time exhaustive.
- The unattended worker example persists a pending report before execution, pulls and revalidates current state, preserves pending work on worker or sync failure, advances an atomic checkpoint only after success, and exits correctly on signals. Delivery is explicitly at least once and workers must be idempotent. The executable harness is pinned to system Bash, passes macOS Bash 3.2, and normalizes Windows CRLF checkout text before extracting the Markdown script.
- Documentation states that `--notes` replaces the complete body. Notes are single-writer replaceable state, not an append-only conversation log; child beads or external comment IDs are recommended for durable multi-writer history.
- The Cursor missing-blocker report was confirmed as a false positive and covered by a blocker-deletion readiness regression. Blocker IDs and issue lookup entries come from the same snapshot collection, and deleting a blocker deletes its outbound edge.
- All runtime CLI exit-code literals were audited and moved to the shared exit-code module.
- The built-in docs cover the feature: the CLI manual (`tbd docs`) is authoritative for selectors, `--since` baseline commits, the report shape, and the repo-wide exit codes; the design doc gains §3.7 (read-only remote observation) and §4.14 (change and watch commands), with §4.10, §5.6, §7.2, §8.2, and §8.7 reconciled. The `watch-beads` shortcut keeps the recipes and platform notes instead of restating the contract, and `development.md` documents the release smoke.
- The release smoke found a real isolation bug that unit tests had not exposed: a configured remote fetch refmap could advance `origin/tbd-sync` even when watch supplied an explicit private destination. Watch fetches now pass `--refmap=`. The regression test starts with a stale `origin/tbd-sync`, and the end-to-end harness snapshots the caller worktree, local sync ref, remote-tracking sync ref, `FETCH_HEAD`, hidden worktree, and private refs before and after a real blocking watch.

## Linear and external-integration boundary

Linear is deliberately not a merge dependency for this PR. The change report, dynamic selectors, resume tip, and durable worker protocol are enough to run disposable one-way experiments on this branch while keeping provider code and operational state outside core.

The remaining promotion work stays separately tracked:

- `tbd-le2l` — merge extension namespaces independently rather than whole-object LWW
- `tbd-z95g` — generic extension read/write/display support
- `tbd-vm5s` — rework the Linear pilot under the Integration Layer rules

A pilot should remain isolated and single-writer until those foundations exist, with external-ID bindings under extensions, module-owned checkpoints, idempotency markers, and failure-injection coverage. This PR does not add provider dependencies, provider schema fields, or provider imports to core.

## Stability

This remains additive and opt-in. There is no `IssueSchema` change, `ConfigSchema`
change, `tbd_format` bump, dependency addition, or lockfile change. Repositories that
do not invoke `changes`, `watch`, or the worker recipe get no new runtime behavior.

The release harness now proves coexistence rather than only post-watch recovery. Two
watchers run concurrently in one checkout while `show`, `list`, `ready`, a full
`sync --pull`, and `sync --status` execute. The harness distinguishes the documented
effects of normal sync from watcher effects, then proves the watchers do not alter the
caller worktree, local or remote-tracking sync refs, `FETCH_HEAD`, hidden worktree,
sync-lock presence, or private refs.

## Validation plan

The authoritative artifacts are the [release validation plan](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/docs/project/specs/active/valid-2026-08-09-bead-watch-release.md), the [manual QA playbook](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/tests/qa/watch-infrastructure-release.qa.md), and the [executable release smoke](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/packages/tbd/scripts/validate-watch-release.ts). Run the source candidate with `pnpm qa:watch-release`; set `TBD_QA_BIN` to validate a packed or installed candidate.

### Automated evidence on dc0ba559

- [x] Focused bead-watch suite: **18/18 tests**.
- [x] Real-Git topology: temporary bare remote, independent writer/watcher clones,
  selected and unrelated movement, durable resume checkpoint, and a real blocked wake.
- [x] Concurrent coordination: bead and `--ready` watchers run simultaneously in one
  checkout and wake from one commit without ref, worktree, or lock contention.
- [x] Every selector family over committed history: one/multiple bead IDs, label,
  status, spec, ready, and all.
- [x] Stable output/exit contract: human, JSON, and quiet output; success 0,
  operational failure 1, usage failure 2, and no-match/timeout 3.
- [x] Existing workflows during and after watching: `show`, `list`, `ready`, full
  `sync --pull`, `sync --status`, and bundled `watch-beads` shortcut discovery.
- [x] Git isolation snapshots cover the caller worktree, local sync ref,
  `origin/tbd-sync`, `FETCH_HEAD`, hidden worktree head/status, sync-lock presence, and
  private watcher refs.
- [x] Source-tree and freshly packed candidates both pass the complete smoke. The
  tarball was installed into an isolated prefix with lifecycle scripts disabled and
  without changing the global CLI.
- [x] `pnpm precommit` and exact pre-push gates: formatting, Markdown, lint, typecheck,
  build, **100/100 files and 1,451/1,451 tests**.
- [x] `pnpm --filter get-tbd test:tryscript`: **1,068/1,068 cases**. The validation run
  also found and fixed a randomized-ID-adjacent missing sentinel (`tbd-dbyj`) rather
  than weakening the fail-closed assertion.
- [x] `pnpm publint`; `pnpm check:package-age` with **0 violations across 31 pins**;
  documentation references and formatting.
- [x] [Exact-head CI run](https://github.com/jlevy/tbd/actions/runs/31351016944): the
  built-candidate smoke passed after unit tests on Ubuntu, macOS, and Windows; coverage
  and lint, benchmark, Cursor Bugbot, and DeepSource also passed.

### Human/release validation remaining

- [ ] **Release gate (`tbd-t750`):** build from the exact release-tag SHA, rerun the
  isolated packed-artifact smoke, and record its SHA-256 checksum plus Node/pnpm/Git
  versions.
- [ ] **Release gate (`tbd-t750`):** run that exact artifact under the supported Node.js
  20 floor; PR CI exercises the built candidate under Node.js 24.
- [ ] **Release gate (`tbd-t750`):** execute the two-terminal playbook against a
  credentialed disposable private GitHub remote and capture the before/after protected
  Git-state diff, auth class, latency, and command transcript.
- [ ] **Before unattended production promotion:** exercise brief and persistent
  transport failures, worker failure/restart/idempotency, intended runner permissions,
  a 30–60 minute idle soak, and packed-artifact shell sampling on macOS, Linux, and
  Windows (including the `.cmd` shim and CRLF shortcut extraction).
- [ ] **Optional, non-gating:** run a disposable one-way Linear sandbox adapter consuming
  the change report. Keep external IDs and checkpoints outside core; neither PR merge
  nor release depends on this experiment.

Merge of this provider-neutral primitive is not coupled to Linear. Release promotion
requires the exact-artifact, Node 20, and credentialed-remote checks; unattended
promotion additionally requires resilience, worker, permission, soak, and platform
evidence.

### Tracking

- `tbd-tp6n` — release-validation epic (in progress until release evidence is complete)
- `tbd-961h` — initial executable smoke and validation artifacts (closed)
- `tbd-3x5y` — cross-platform/concurrent smoke expansion (closed)
- `tbd-dbyj` — deterministic missing-ID transcript hardening (closed)
- `tbd-t750` — remaining manual release-promotion QA (open)
- `tbd-g305` / `tbd-vm5s` — Linear pilot design rework, explicitly non-gating

## Relationship to earlier PRs

- #196 is code-wise contained here and superseded.
- #197 contains the earlier Linear research/pilot material; that work remains separate and is tracked for rework under the Integration Layer rules.

---

> [!NOTE]
> **Medium Risk**
> New remote polling and private-ref fetch paths touch Git isolation and agent wake contracts, but the design is read-only, heavily regression-tested (including refmap isolation), and opt-in with no storage schema changes.
> 
> **Overview**
> Adds **opt-in, read-only** infrastructure so agents can diff committed sync-branch bead state and block until remote changes match a selector.
> 
> **`tbd changes --since <commit>`** diffs the local sync-branch tip against a baseline without fetch or hidden worktree access. **`tbd watch`** polls the remote tip, fetches only after movement into a **private ref** (with `--refmap=` so `origin/tbd-sync` is not updated), emits one **`format_version: 1`** JSON/human report, and exits. Selectors mirror **`list`/`ready`** via shared **`issue-selection`** helpers (bead, label, spec, status, **`--ready`**, **`--all`**).
> 
> Exit codes are centralized (**0** success, **1** operational, **2** usage, **3** no match/timeout, **130** SIGINT). Git network calls gain **timeouts**; snapshot reads use bounded **`cat-file`** batches and capped text hunks.
> 
> Ships the **`watch-beads`** shortcut (durable watch-then-worker recipe), README/CLI docs, active specs/validation plans, **`pnpm qa:watch-release`**, and a **two-clone CI smoke** on Ubuntu/macOS/Windows. No schema, config, dependency, or default-command behavior changes unless these commands are invoked.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc0ba5599ff624517e0905c595a1ede388593dc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy:
## Senior engineering review — PR #205

**Scope.** I reviewed PR #205 at `4a2c0318896dacf17fb0151bffa8957fab90490e` against `main`: 28 files, +3,705/-60. I checked the complete diff, the active spec and validation transcript, the user/agent documentation, prior review context, GitHub Actions history, and the implementation/tests on the checked-out branch. I also ran full local validation and targeted deadline, hung-I/O, diff-scale, and shell-signal probes.

**Verdict: request changes before merge.** The core architecture is a good additive foundation, and I would keep Linear implementation out of this PR. Before this becomes stable coordination infrastructure, however, the timeout contract, change-engine resource bounds, and advertised unattended workflow need hardening. Finding count: **4 High, 1 Medium, 1 Low**.

## Design assessment

The main architectural choices are strong:

- `tbd changes` is a pure, deterministic primitive and `tbd watch` is a thin blocking wrapper.
- Remote observation uses collision-resistant private refs, does not write `FETCH_HEAD`, and avoids the shared sync worktree and its lock. That is the right isolation boundary for an optional watcher.
- Strict committed-snapshot parsing, ancestry validation, explicit created/deleted changes, shared list/ready predicates, stable exit code 3, and injectable polling boundaries are all good design decisions.
- Existing workflows remain opt-in and effectively untouched; the behavior-bearing list/ready extraction is covered by the existing suite.
- The Integration Layer rules are directionally right: provider code remains separable and is never imported by core; experiments do not alter `IssueSchema`, `ConfigSchema`, or `tbd_format`; bindings live under `extensions`; operational watermarks/base snapshots stay module-owned; promotion follows a working pilot.

Relative to webhooks, a resident daemon in core, or folding polling into `tbd sync`, this remote-tip primitive is simpler, portable, provider-neutral, and much easier to test. I would retain it. The weak point is not the primitive boundary but the delivery/timeout semantics around it and the shell workflow currently presented as production-ready.

For Linear, this PR should remain independently mergeable after the findings below are addressed. The current branch is enough for disposable one-way experiments, but not yet for a safe concurrent two-way bridge: `extensions` still merges whole-object LWW and there is no generic extension read/write CLI. The follow-up order in the spec is appropriate: first fix per-namespace extension merging and generic extension access; then build an isolated, single-writer adapter with external-ID bindings, module-owned state, idempotency markers, failure injection, and a narrow import/status-writeback pilot. Do not add provider schema/config fields or import that adapter from core.

## Findings

### R1 — High — A change arriving in the final polling interval is discarded

`packages/tbd/src/file/bead-watch.ts:194-200`

After sleeping to the deadline, the loop returns `timeout` before performing the poll that sleep was waiting for. With `--interval 10 --timeout 10`, the initial baseline is the only remote read; an A→B change during the sleep returns exit 3. If the caller starts a fresh bounded watch without a persisted `--since`, B becomes the new baseline and the wake is permanently lost.

I reproduced this with the injected clock and A→B tips: result `timeout`, one remote-tip read, logical clock 10,000 ms.

**Fix:** Make every completed sleep lead to a poll, including the final partial interval, then decide whether the deadline elapsed. Define the deadline boundary explicitly (an observation at the deadline should count), and add tests for `interval === timeout`, a final partial interval, and a caller resuming from the last acknowledged tip.

### R2 — High — `--timeout` does not bound the Git/network operations

`packages/tbd/src/file/bead-watch.ts:110-145`

The deadline only controls sleeps. Both `git ls-remote` and `git fetch` are awaited without an AbortSignal or subprocess timeout. `gitNoPrompt` prevents credential prompts, but it does not prevent a stalled transport, helper, or DNS operation from holding the watcher forever. This breaks the documented portable bounded-loop contract for agent sessions.

I injected a never-resolving second remote-tip read: a watch with a 5 ms logical timeout was still pending after 50 ms and never reached cleanup.

**Fix:** Propagate the remaining deadline into each Git operation, terminate the child safely when it expires, and distinguish deadline expiry from the existing consecutive transient-poll failures. Cover a hung startup read, a hung established poll/fetch, and private-ref cleanup after cancellation.

### R3 — High — The change engine is not resource-bounded for valid issue data

`packages/tbd/src/lib/issue-changes.ts:152-179`, `packages/tbd/src/lib/issue-changes.ts:401-407`, `packages/tbd/src/file/sync-branch-changes.ts:42-43`, `packages/tbd/src/file/sync-branch-changes.ts:82-105`

The Myers implementation stores a full `Map` trace at every edit distance, giving worst-case quadratic memory. A valid body may contain 50,000 characters. In a probe with only 1,000 completely different short lines (roughly 8 KiB per side), one report took about 357 ms and grew the heap to roughly 166 MiB. Larger valid notes/descriptions can exhaust the process.

This cost is paid for every candidate before the selector is checked, so `--bead one-id` can diff unrelated changed bodies. Separately, each full snapshot is buffered behind a hard 50 MiB aggregate `cat-file` limit; a valid larger graph fails even for a static one-bead selection. “Three context lines” bounds displayed context, not diff computation, full `before`/`after` JSON values, or snapshot memory.

**Fix:** Add a deterministic complexity cutoff with a replacement/summary fallback (or use a linear-space diff), apply static selection before expensive field deltas, and stream or chunk `cat-file` reads instead of imposing one graph-wide buffer ceiling. Add regression cases at the schema’s maximum body size and a large graph for both `--bead` and `--all`, with explicit time/memory ceilings.

### R4 — High — The documented default daemon is at-most-once and reads stale state

`packages/tbd/docs/shortcuts/standard/watch-beads.md:45-71`

The sample advances `since_args` before the worker succeeds and then suppresses every worker failure with `|| true`. A crash, auth failure, sandbox failure, or failed `tbd sync` therefore acknowledges the report and advances past it with no retry or durable failure record.

The watcher intentionally fetches only into a private ref; it does not update the local sync branch/worktree. Yet the worker prompt tells the agent to act before requiring `tbd sync --pull`. That is particularly unsafe for `--ready`: a bead can enter the ready set because another bead changed, yielding an empty field delta for the selected bead while local `tbd ready/show/update` still sees the old graph. Acting on that stale store can select the wrong work or later overwrite remote state.

Finally, `trap 'rm -f "$wake_file"' ... TERM` handles TERM without exiting; a Bash loop with this trap continues after SIGTERM. I reproduced that behavior.

**Fix:** Define an explicit at-least-once acknowledgement protocol: retain the old checkpoint/report until the worker and its final sync succeed; retry with bounded backoff or stop nonzero while preserving the report; pull and revalidate current bead state before mutation; and use signal handlers that clean up and exit with the correct status. Add end-to-end tests for worker failure/restart, a change arriving while the worker runs, stale local state, and SIGTERM.

### R5 — Medium — The advertised notes-based message protocol overwrites history

`packages/tbd/docs/shortcuts/standard/watch-beads.md:80-84`, `docs/project/specs/active/plan-2026-07-19-bead-watch-and-external-sync.md:289-295`, `packages/tbd/src/cli/commands/update.ts:147-152`

The docs tell agents to send messages with `tbd update <id> --notes <message>`, but `--notes` is a setter: it assigns `issue.notes = updates.notes`. Even without concurrency, a second up-to-date writer replaces the first message and creates no conflict/attic artifact. LWW-with-attic only protects a losing concurrent branch; it is not append semantics.

**Fix:** For Phase 1, either describe notes as replaceable state and document an exact pull/read/preserve/write/sync protocol, or stop promising conversation history. If durable conversation is required, add a future atomic append or union-by-ID comments primitive. Pin the chosen protocol with a sequential two-agent transcript, not only a concurrent merge case.

### R6 — Low — A generated development-version stamp leaked into the PR

`.tbd/config.yml:2`, `.tbd/config.yml:17-18`

The diff changes `tbd_version` from the released `0.4.2` to a dirty development build and records it in `tbd_upgrades`. This is local setup churn and contradicts the PR description’s claim that only the watch shortcut mapping changes in this file.

**Fix:** Restore the released version and remove the development upgrade entry; retain only the new `docs_cache` mapping.

## Non-blocking suggestions

1. Make the substantive-field contract compile-time exhaustive. The current `ISSUE_CHANGE_FIELDS satisfies readonly (keyof Issue)[]` proves that listed names are valid, but not that a newly added substantive schema field was included. Derive it from the canonical order with explicit metadata exclusions, or add an exhaustiveness assertion/test.
2. Give the stable JSON report an explicit format/version contract before external agents and adapters depend on it.
3. Keep the Linear E2E work as a separate pilot artifact. A useful thin slice on this branch is: mocked GraphQL contract → external-ID-bound one-way import → one status writeback → duplicate/retry/crash/rate-limit tests. Treat whole-object `extensions` LWW as a single-writer limitation until the planned merge fix lands.

- The existing Cursor/Bugbot comment about an absent blocker being treated as unresolved is benign. `blockerIdsByTarget` is built by iterating the same `allIssues` used to create `issueById`; every pushed blocker ID is therefore present in that map. Also, a `blocks` edge is stored on the blocker and points to the blocked target, so deleting the blocker removes the edge with it.
- I found no existing `list`/`ready` behavior regression from extracting the shared predicates. The full legacy suite passed.
- The GitHub check failures are not evidence of source failures: on attempt 3, Benchmark, Coverage & Lint, Ubuntu, and macOS were cancelled before acquiring runners and contain no steps; Windows completed build/tests successfully. An earlier macOS attempt failed while GitHub’s action-download service returned unavailable.

## Validation and CI status

Local validation on the PR head:

- `pnpm format:check` — pass
- `pnpm lint:check` — pass
- `pnpm build` — pass
- `pnpm test` — **98 files / 1,429 tests passed**
- `pnpm --filter get-tbd test:tryscript` — **1,068 transcript cases passed**
- `pnpm publint` — pass
- Focused watch/change suites and 5,000 randomized repeated-line diff reconstructions — pass
- Targeted deadline, hung-I/O, resource, and SIGTERM probes — reproduced R1–R4

GitHub is **not green at review time**: Windows and DeepSource pass, but four jobs are recorded as failed/cancelled because the hosted jobs were not acquired (after three attempts). Those checks still need a successful rerun before merge even though the branch passes locally: [workflow run 31120466453](https://github.com/jlevy/tbd/actions/runs/31120466453).

packages/tbd/src/cli/lib/exit-codes.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Shared CLI exit codes for the bead change commands.
+ *
+ * Exit 2 is reserved repo-wide for usage errors (see ValidationError in
```
jlevy: Exit code 2 should be part of this file if it's a defined constant. All exit codes should be in one file. 

Check if there are any other code inconsistencies of this type and fix them all.

jlevy:
## Review follow-up complete

I addressed every senior-review finding and all three inline review threads. The final head is [`5b23f7f8`](https://github.com/jlevy/tbd/commit/5b23f7f88f67965ca72df4c5a92f6d9269137a6e).

### Blocking findings

- **R1 — deadline boundary:** completed and partial sleeps now lead to an inclusive final observation, with resume-tip coverage.
- **R2 — bounded Git operations:** remote reads and fetches use positive subprocess timeouts; fetch and private-ref `rev-parse` now share one deadline, so resolution receives only the remaining poll budget. Startup, established poll/fetch, boundary failure, and cleanup paths are covered.
- **R3 — bounded change computation:** text diffing has a deterministic complexity cutoff, selection precedes expensive deltas, object reads are batched, and maximum-body / 1,000-selected / 2,000-all graph cases are covered.
- **R4 — durable unattended delivery:** the `watch-beads` worker persists pending work before execution, pulls and revalidates, advances the checkpoint only after worker plus final sync success, preserves failures for retry, and exits on signals. The executable failure/restart harness passes macOS Bash 3.2 and Windows CRLF checkout handling.
- **R5 — notes semantics:** docs now describe `--notes` as whole-body replacement and direct durable multi-writer history to child beads or external comment IDs.
- **R6 — config hygiene:** the generated development stamp is gone; the only intended config change is the `watch-beads` docs-cache registration.

The non-blocking suggestions are also implemented: report `format_version: 1`, compile-time exhaustive substantive fields, bounded human rendering, and centralized success/operational/usage/no-match/SIGINT exit codes. The missing-blocker Cursor finding was confirmed as a false positive and pinned with a blocker-deletion readiness regression.

### Design and Linear boundary

The PR remains additive, provider-neutral, and independently mergeable: no issue/config schema change, format bump, dependency, lockfile, or provider import. It is sufficient for disposable one-way Linear experiments while promotion work remains separately tracked in `tbd-le2l`, `tbd-z95g`, and `tbd-vm5s`.

### Validation

- Full local commit/push gates: formatting, lint, typecheck, build, **100 test files / 1,451 tests**
- CLI transcripts: **1,068 cases**
- `publint`: pass
- package-age policy: **0 violations across 31 pins**
- [Exact-head GitHub run](https://github.com/jlevy/tbd/actions/runs/31332349329): Ubuntu, macOS, Windows, Coverage & Lint, Benchmark, Cursor Bugbot, and DeepSource all pass
- Final thread audit: **0 unresolved review threads**

PR #205 is ready for merge from the engineering-review and CI perspective.

jlevy:
Release validation update for `ec0424f4`

The PR body now carries the complete validation matrix. The durable artifacts are:

- [Release validation plan](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/docs/project/specs/active/valid-2026-08-09-bead-watch-release.md)
- [Manual QA playbook](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/tests/qa/watch-infrastructure-release.qa.md)
- [Executable two-clone smoke](https://github.com/jlevy/tbd/blob/claude/watch-infrastructure/packages/tbd/scripts/validate-watch-release.ts)

Automated evidence is green: the source and isolated packed-candidate smokes; focused watch tests (18/18); precommit/pre-push formatting, lint, typecheck, build, and unit/integration tests (100 files / 1,451 tests); Tryscript (1,068 cases); publint; package-age; and documentation checks.

The smoke found and drove a real fix: an explicit private fetch destination could still honor the configured remote refmap and advance `origin/tbd-sync`. Watch fetches now pass `--refmap=`, with both a focused regression and end-to-end non-mutation assertions for the caller worktree, local/remote sync refs, `FETCH_HEAD`, hidden worktree, and private refs.

Remaining release gates are tracked by `tbd-t750`: rerun the exact tagged package and record its checksum/transcript, then execute the credentialed hosted-remote two-terminal scenario. Resilience, worker restart/idempotency, runner permissions, and the platform matrix gate unattended operational promotion. A disposable Linear sandbox experiment is explicitly optional and non-gating for both merge and release.

jlevy:
## Validation plan update — `dc0ba559`

The validation increment is committed and pushed. The PR body and durable artifacts now
cover the complete automated/manual split:

- [Release validation plan](https://github.com/jlevy/tbd/blob/dc0ba5599ff624517e0905c595a1ede388593dc0/docs/project/specs/active/valid-2026-08-09-bead-watch-release.md)
- [Manual QA playbook](https://github.com/jlevy/tbd/blob/dc0ba5599ff624517e0905c595a1ede388593dc0/tests/qa/watch-infrastructure-release.qa.md)
- [Executable release smoke](https://github.com/jlevy/tbd/blob/dc0ba5599ff624517e0905c595a1ede388593dc0/packages/tbd/scripts/validate-watch-release.ts)

What changed:

- The built-candidate smoke now runs in the Ubuntu, macOS, and Windows CI jobs.
- The real-Git topology covers concurrent bead/ready watchers, every selector family,
  human/JSON/quiet output, exit classes 0/1/2/3, and a matching `--all` report.
- While both watchers are blocked, the harness runs `show`, `list`, `ready`, a full
  `sync --pull`, and `sync --status`. It then proves the watchers did not alter the
  caller worktree, sync refs, `FETCH_HEAD`, hidden worktree, lock state, or private refs.
- The source candidate and a freshly packed candidate installed into an isolated prefix
  both pass the expanded smoke.
- Full transcript validation found a randomized-ID-adjacent fixture (`test-zzzz`) and
  replaced it with a deterministic distant sentinel while preserving explicit
  fail-closed assertions (`tbd-dbyj`).

Automated evidence is green:

- precommit and pre-push: formatting, Markdown, lint, typecheck, build, 100/100 files,
  1,451/1,451 tests;
- Tryscript: 1,068/1,068 cases;
- focused watcher tests: 18/18;
- publint and dependency-age policy: pass, 0 violations across 31 pins;
- [exact-head GitHub run](https://github.com/jlevy/tbd/actions/runs/31351016944): Ubuntu,
  macOS, Windows, coverage/lint, benchmark, Cursor Bugbot, and DeepSource all pass.

The remaining work is intentionally a release-promotion gate on `tbd-t750`, not a PR
merge gate: exact release-tag SHA/checksum, Node.js 20 floor, credentialed disposable
GitHub remote, transport failure/recovery, durable-worker replay/idempotency, intended
runner permissions, idle soak, and packed-artifact platform/shell sampling. Linear
experimentation remains optional and non-gating.

jlevy:
## Technical review: merge safety and additivity of the watch infrastructure

Scope of this review: is this branch a stable, safe merge that is strictly additive to current tbd behavior, given that web-UI and Linear-integration work is proceeding in parallel and must not block on it?

**Verdict: safe to merge, and additive with one narrow exception (detailed below, not a blocker).**

### Verification performed

All against head `dc0ba55`:

- `typecheck` and `lint`: clean
- Full vitest suite: 101 files, 1458 tests passed
- Full tryscript CLI suite: 1068 assertions passed
- `qa:watch-release:built` smoke: all 16 steps passed in ~31s locally
- Behavioral A/B of refactored commands against a `main` worktree (details below)
- PR CI: green on ubuntu, macos, windows plus Coverage & Lint and Benchmark

### Why the "strictly additive" claim holds

The new surface is genuinely new: `tbd changes`, `tbd watch`, and six new modules (`bead-watch.ts`, `sync-branch-changes.ts`, `issue-changes.ts`, `issue-selection.ts`, `change-selection.ts`, `issue-changes-output.ts`, `exit-codes.ts`). Only two existing code paths were touched, and both are pure extractions:

1. **`ready.ts`** now delegates to `readyIssueIds()` in `src/lib/issue-selection.ts`. Behaviorally identical: the old inline `blocker && blocker.status !== 'closed'` and the new `issueById.get(id)?.status !== 'closed'` differ only for a dangling blocker id, and by construction every blocker id in the reverse map comes from a real issue in the same snapshot.

2. **`list.ts`** now delegates label/spec/status filtering to `issueMatchesSharedFilters()`. Label (ANDed) and spec (gradual matching) semantics are preserved; the empty-string spec is explicitly mapped back to "no filter". One edge regressed; see below.

The exit-code extraction is a true no-op: `EXIT_SUCCESS/OPERATIONAL_ERROR/USAGE_ERROR/INTERRUPTED` are `0/1/2/130`, exactly the literals they replaced, and `exit-codes.test.ts` pins those values so drift would fail CI. The only new code is `EXIT_NO_MATCHING_CHANGE = 3`, used solely by the two new commands.

`git.ts` changes are additive: `GitError` gains `timedOut`/`timeoutMs` fields defaulting to `false`/`null`, and the timeout message branch is unreachable unless a caller passes a timeout, which only the new `gitNoPromptWithTimeout` does.

### Non-intrusiveness of `tbd watch` is machine-enforced, not asserted

Watch fetches to a collision-resistant private ref `refs/tbd/watch/<pid>-<uuid>` with `--no-write-fetch-head --no-tags --refmap=`, and reads snapshots via `ls-tree` / `cat-file --batch` with no checkout. The smoke script's `snapshotSafety()` diffs, across every watch run: caller working-tree status, local sync tip, remote-tracking tip, `FETCH_HEAD`, hidden data-sync worktree HEAD and status, data-sync lock presence, and leftover private watch refs. It also exercises `sync`, `list`, `ready`, and bead mutation while two watches are concurrently blocked. This is the strongest available form of the "does not disturb existing workflows" guarantee, and it now gates CI.

### No coupling with the parallel web/Linear work

The in-flight web spike (`bead-web.ts` etc.) imports nothing from the watch modules, and this branch never touches `stats.ts`. The only shared file is `.tbd/config.yml`, which gains one `docs_cache` row. Merging this PR cannot block or be blocked by that work.

### One real regression (small, fix in a follow-up or before merge)

`tbd list --status ""` changed from "ignore the filter" (main returns all 224 issues) to "match nothing" (branch returns `No issues found`). Verified by running both trees side by side. Cause: `list.ts` passes `options.status ?? null` into the shared filter, where the old code's `if (options.status && ...)` treated `''` as falsy. The same commit explicitly handled this case for `--spec` (`options.spec === '' ? null : ...`) but not for `--status`, so it reads as an oversight rather than a decision. Either mirror the spec handling or reject the empty value as a usage error the way `parseChangeSelection` already does for `watch`/`changes`. Contrived input; not a merge blocker.

### Two awareness items, not blockers

1. **New CI gating surface.** The watch smoke runs on all three OS matrix legs at ~31s each. Four of the nine commits on this branch are cross-platform fixes for exactly this script (macOS Bash 3.2, CRLF, newline handling), so it has been shaken out, but if it ever flakes it blocks unrelated PRs. If insulation is wanted, restricting the smoke to ubuntu-latest keeps the coverage while removing two-thirds of the flake exposure.

2. **Bundled unrelated cleanup.** The `test-zzzz` → `test-zzzzzzz` rename across `cli-bulk-mutation.tryscript.md` is an improvement (the 4-char id was being swallowed by the `[SHORTID]` normalizer, so `[..]` wildcards became exact error assertions), but it is unrelated to watch. Naming it here so it doesn't read as scope creep.

### Self-documentation for agents: reference surface complete, one gap

Complete: `tbd watch --help` / `tbd changes --help`, a full `tbd-docs.md` section with an explicit exit-code table, the 202-line `watch-beads` shortcut with working daemon, Claude Code, and Codex recipes (recipe semantics are themselves tested in `watch-beads-shortcut.test.ts`), README coverage, and a `docs_cache` entry so `tbd shortcut watch-beads` resolves.

Gap: `skills/tbd/SKILL.md` (the document agents load first) never mentions `tbd watch` or `tbd changes`. Its "Dependencies & Sync" command table lists `sync`/`stats`/`doctor` but no watch, and the "User Request → Agent Action" table has no row for "wait until X changes" or "wake on new work". The only trace is one `watch-beads` line inside the ~40-row auto-generated shortcut directory. Suggested docs-only follow-up:

- Add to Dependencies & Sync: `tbd watch --ready --json` (block until a bead newly becomes ready; exit 3 on `--timeout`) and `tbd changes --since <commit>` (committed bead deltas on the local sync branch)
- Add a request-to-action row: "wait for someone else's work / wake on new work" → `tbd watch --ready` or `tbd shortcut watch-beads`
- Add a Quick Reference line noting exit 3 means "no matching change", since that convention is new and agents will otherwise treat any nonzero exit as failure

### Summary for the final reviewer

- Merge-safe: yes. All suites green locally and in CI; mergeable and clean against main.
- Strictly additive: yes, except the `--status ""` edge above.
- Parallel web/Linear work: fully decoupled; nothing here blocks it or depends on it.
- Recommended follow-ups (either pre-merge or immediately after): the `--status ""` fix and the SKILL.md watch/changes additions.

PR #212, opened 2026-08-13
fix(integrations): migrate Linear managed blocks

jlevy:
## Summary

- replace visible Linear HTML-comment markers with the stable plain-text pair `⟦tbd⟧` / `⟦/tbd⟧`
- dual-read the legacy format, write only the current format, preserve human prose, and fail closed on malformed regions
- project managed metadata during both one-way mirror and full bidirectional sync, including values reconciled in the same run
- keep already-linked deep beads synchronizable while retaining `max_nesting` as a creation limit and preserving provider hierarchy when an ancestor is skipped
- strengthen the repeatable live QA for marker verification and ambient real-project discovery
- harden scenario-owned live-QA failure checks, no-op push reporting, and absent `.env` ignore preflight; document prose-push block relocation
- update the integration plan, design docs, user docs, QA playbook, and generated tbd skills

## Existing-data migration

- backfilled all 163 currently linked Linear issues
- independently audited all 163 exact Linear UUIDs through GraphQL
- result: 163 current-format descriptions; zero legacy, malformed, or missing managed regions

## Validation

- `pnpm run ci`: 134 test files and 1,982 tests passed, plus format, docs format, typecheck, lint, and build
- `pnpm --filter get-tbd qa:linear-live -- --team TBD --project tbd`: all 11 API-driven scenarios passed, including fields, comments, assignee, hierarchy, discovery scope, conflict recovery, exact-once settling, orphan detection, current-marker verification, and fixture cleanup
- mandatory pre-commit and pre-push hooks passed

## Beads

Closes `tbd-bldq`, `tbd-r3d4`, `tbd-t5vq`, `tbd-mlu0`, and `tbd-363x`.

---

> [!NOTE]
> **Medium Risk**
> Changes the Linear sync and description-splice path where mistakes could affect external issue text, but behavior is fail-closed on bad markers and covered by expanded unit and live QA scenarios.
> 
> **Overview**
> **Linear managed regions** now use the plain-text `⟦tbd⟧` / `⟦/tbd⟧` pair instead of HTML comments. Readers still accept the legacy comment delimiters and upgrade them on the next outbound write; mixed or broken markers fail closed without rewriting provider descriptions.
> 
> **Full sync** re-splices the managed block from reconciled bead values in the same run (backfill, refresh after pulls, skip pushes when already satisfied). **`max_nesting`** only blocks *new* outbound creates—already-linked beads deeper in the tree still update, without forcing parent changes when an ancestor is skipped.
> 
> **Onboarding:** new **`tbd shortcut setup-linear`** (first-time repo vs teammate-with-key-only), wired through README, welcome, setup/prime output, skills, and integration `status` hints. **`.env` preflight** warns when `.env` is not gitignored even before a key exists; live QA asserts current delimiters on Linear and tolerates ambient project noise in scope scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f588b5e96f32bc20ce05e0fee689981df349b24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy:
Final implementation and validation disposition:

- **Delimiter design — fixed.** The sole writer constants are `MANAGED_BLOCK_MARKERS.begin = ⟦tbd⟧` and `.end = ⟦/tbd⟧` in `core/managed-block.ts`. These are visible, compact, and inert under Markdown/rich-text rendering.
- **Backward compatibility — fixed.** Readers accept both current and legacy HTML-comment pairs. The next outbound write atomically replaces a legacy pair with the current pair while preserving human prose on both sides.
- **Malformed content safety — fixed.** Mixed, incomplete, reversed, and duplicate marker forms fail closed. Full sync quarantines the affected pair before field or comment writes while unrelated pairs continue.
- **Full bidirectional sync — fixed.** Managed metadata is now an explicit journaled provider projection, not a merge-owned description field. It is backfilled/refreshed after prose changes and uses the same-run reconciled status and priority.
- **Deep linked beads — fixed.** `max_nesting` remains a limit on new external creation; already-linked descendants remain synchronizable.
- **Existing data — migrated and audited.** All 163 linked Linear UUIDs were updated. An independent GraphQL audit found 163 current-format descriptions and zero legacy, malformed, or missing regions.
- **Live QA isolation — fixed.** Project discovery now validates its owned sentinels from structured output without assuming the real Linear project has no unrelated hierarchy failures.
- **Documentation and skills — complete.** The plan/spec, architecture/design, user docs, development guide, QA matrix, source skill baseline, and all generated skill mirrors describe the same contract.

Validation on commit `0ca74c0d`:

- `pnpm run ci`: format, docs format, typecheck, lint, build, 134 test files / 1,973 tests passed.
- Linear API QA: all 11 scenarios passed and fixtures cleaned.
- GitHub Actions: benchmark, coverage/lint, Linux, macOS, and Windows all passed.
- DeepSource: grade A; no blocking security or static-analysis findings.

Cursor Bugbot has remained externally stalled after gathering PR context and has posted no finding; it is not a repository-owned or required check.

jlevy (commented):
## Senior engineering review — PR #212 and the Linear integration

**Scope.** PR #212 at `0ca74c0` (21 files, +449/−88), plus a full read of the
`src/integrations/` subsystem it lives in (21 files, 5,403 lines) and the CLI surface
that drives it. Findings below carry stable IDs; every one has a bead.

**Verdict: approve the change, but do not tag a release until R1 is fixed.** R1 is a
regression introduced by this PR that breaks `tbd sync --push` permanently for a
reachable repository shape. Everything else is pre-existing debt this review surfaced
while checking whether the integration is "consolidated and mapped out" — the honest
answer to that question is R2.

The core of the PR — the marker migration and treating the managed block as a
projection of reconciled values — is well built. Dual-read/single-write with
fail-closed on ambiguity is the right shape, the migration is idempotent, and the test
coverage added is real coverage, not ceremony.

---

### R1 — Blocker for release · `mirror.ts:202`

**The `max_nesting` exemption converts a clean skip into a permanent hard failure.**

An already-linked bead deeper than `max_nesting` now enters `plan.updates`. If a
*selected* ancestor between it and the mirror root is itself unlinked and therefore
skipped, `applyMirror` (`mirror.ts:334`) throws `parent <id> was not mirrored` for that
bead — on every run, with no path to recovery. `sync.ts`
`assertIntegrationReportsHealthy` promotes any integration failure to a command
failure, so routine `tbd sync --push` fails outright.

Reproduced with a probe test (`root → mid1 → mid2 → deep`, `max_nesting: 2`, `mid2`
unlinked, `deep` linked):

```
skipped:  [{ beadId: tbd-mid2, reason: "nested 3 levels, past max_nesting 2" }]
failures: [{ beadId: tbd-deep,  error: "parent tbd-mid2 was not mirrored" }]
```

Reachable by linking a deep bead directly (`integration link`), by unlinking an
intermediate, or by lowering `max_nesting` after linking.

**Fix:** when a bead is included *only* because it is already linked, omit `parentId`
from the patch rather than emitting `null`, and drop the "parent was not mirrored"
requirement for it. The sync engine already does exactly this at
`sync-engine.ts:636-645` — `expectedParentId === undefined` leaves the provider parent
alone. Mirror that rule in `planMirror`. → `tbd-8raq`

---

### R2 — High · the consolidation answer

**`--push` and full `sync` are two different engines behind one flag vocabulary.**

- `tbd integration sync --push` and `tbd sync --push` → `runEnabledIntegrationPushes` →
  `planMirror`/`applyMirror` (the Phase 1 one-way mirror).
- Bare `sync` and `--pull` → `runSync`.
- `runSync` has **no outbound-only mode**: `direction?: 'both' | 'inbound'`
  (`sync-engine.ts:143`).

Taking the `--push` path means no three-way reconcile (a Linear-side edit is
overwritten with no conflict comment and no archived artifact), no intent journal and
so no crash replay, no comment push, and `parentId: null` can un-parent a provider
issue.

It also writes no bridge base: `writeLinkRecord` is reachable only from `sync-engine.ts`
and the `link` command, while the push path's `onLinked` writes the link onto the bead
alone. That degrades gracefully rather than breaking — the first full sync afterwards
seeds the base from the remote snapshot (`sync-engine.ts:598-611`) — but the
consequence is that any tracker-side edit made during a `--push`-only period is
silently adopted as the baseline instead of surfacing as a pull or a conflict.

Both `skill-baseline.md` and `tbd-docs.md` state that direction flags "mean the same
thing everywhere in tbd" and present `--push` as *outbound only* — the same operation,
half the directions. That is true of the flag and false of the engine, and the manual
recommended `--push` as **step 3 of first-time setup**, routing every new user through
the weaker path. R1 is a symptom of the same split: the fix was applied to the mirror
planner to match behavior the sync engine already had, instead of removing the
divergence.

**Fix (pick one):** (a) add `direction: 'outbound'` to `runSync`, route `--push`
through it, retire `planMirror`/`applyMirror`; or (b) keep both and document the
divergence honestly. (a) is the real fix. I did the doc half of (b) in the companion
branch as an interim measure. → `tbd-dqiq`

---

### R3 — Medium · `sync-engine.ts:775-801`

**Malformed markers freeze the whole pair, including comments and inbound-only runs.**

The malformed scan runs for every pair regardless of direction; the pair is then
dropped from `synchronizablePairs`, which also drops its bead patch, comment pull, and
comment push. On a `--pull` run the message still reads *"skipped all writes for this
pair"* although a pull performs no provider writes.

Failing closed on the **description** is correct — `stripManagedBlock` returns the raw
body on malformed input, so a pull would otherwise write machine text into the bead.
Extending the freeze to comments and unrelated fields is broader than necessary.

There is also no way to *find* a stuck pair: `integration status` reports config,
credential, and reachability only, never a count of quarantined pairs. One can sit
frozen indefinitely with a single line in a sync report as the only signal.

**Fix:** scope the skip to description-bearing operations, reword the message for
inbound-only runs, and surface a quarantined-pair count in `integration status`.
→ `tbd-bexc`

---

### R4 — Medium · `validate-linear-integration-live.ts:196-202`

**The live-QA relaxation discards the signal it was meant to preserve.**

`expectReportOutput` accepts exit `0` *or* `1` for `automatic-inbound-scope`. The
rationale in the comment is legitimate — ambient real-project items can fail unrelated
imports. But nothing asserts *which* failures occurred, so a genuine regression that
breaks the scenario's own work now passes.

**Fix:** keep tolerating exit 1, but parse `report.failures` and require that none name
either scenario-owned sentinel. Ambient tolerance without discarding the signal.
→ `tbd-fqf4`

---

### R5–R9 — Low

| ID | Finding | Fix | Bead |
| --- | --- | --- | --- |
| R5 | `sync-engine.ts:1109-1118` reports a push when `spliceDescription` returns `null` (remote already correct). Dry-run and live paths disagree. | Push the id only when `result` is non-null or `hasExternalPatch`. | `tbd-bhv5` |
| R6 | A prose push wipes the block (`applyChanges` writes stripped prose) then re-appends it, so a mid-description block **relocates to the bottom**; costs 3 API calls per description change. The design doc's "byte-for-byte" claim covers the prose, not the block position. | Have reconcile emit the already-spliced description, or document the relocation. | `tbd-nfjq` |
| R7 | `status.ts:76-84` reports `.env: not present` as `ok` without checking whether one *would* be gitignored — the warning arrives only after a key is on disk. | Run `git check-ignore` even when absent. | `tbd-sjta` |
| R8 | Dual-format reading has no removal plan, so `LEGACY_MANAGED_BLOCK_MARKERS` is load-bearing forever — despite all 163 linked issues already being migrated and audited. | Drop the legacy branch after a release or two. | `tbd-dhfm` |
| R9 | `repoUrl` is threaded through `MirrorContext` and `MirrorLinks` but `mirrorExtrasFor` always passes `undefined` and the runner never supplies it, so the `Bead: [id](url)` line and bead-source attachment are unreachable. | Wire it (`core/permalink.ts` already builds blob URLs) or delete the branches. | `tbd-dj3t` |

---

### Checked and confirmed benign — do not "fix" these

- **`locateManagedBlock` ordering.** Current-format-first; a body carrying both pairs
  correctly returns `malformed`. Verified against the added unit tests.
- **Migration idempotency.** `spliced.result !== pair.remote.description` holds for a
  legacy region, so migration happens on the next outbound splice and then settles.
  `descriptionHash` strips either format, so no oscillation.
- **`splice_description` intent replay.** Already in the op schema (`intents.ts:69`),
  journaled before provider writes, idempotent (`spliceDescription` returns `null` on a
  no-op).
- **Base advance after a splice.** `postWriteUpdatedAt` correctly takes the splice
  timestamp, so the splice does not read as a remote edit on the next run.
- **`report.pushed` double-count** on the `hasExternalPatch` + `managedBlock` path — is
  guarded. (R5 is the *opposite* case.)
- **Marker choice `⟦tbd⟧` / `⟦/tbd⟧`.** Plain text, no Markdown or HTML semantics,
  survives Windows CI. Users now see the delimiters in Linear, which is a real UX cost,
  but it is the right trade against depending on Linear's rich-text sanitizer to
  round-trip HTML comments.
- **`stripManagedBlock` returning the full body on malformed input** — deliberate, and
  correct given R3's quarantine.

---

### Design assessment

The `core/` versus provider split is clean and the provider boundary is honest: policy,
reconcile, intents, and orchestration are all provider-neutral, and `LinearAdapter`
holds transport and mapping only. The write-ahead intent journal with client-ids gives
genuine exactly-once semantics rather than best-effort retry, three-way reconcile
against a durable base is the right model for bidirectional sync, and fail-closed on
ambiguity is applied consistently rather than selectively.

The structural debt is R2. Two engines behind one flag vocabulary is the kind of split
that keeps generating bugs shaped like R1 — a divergence gets patched on one side
instead of removed. Consolidating `--push` onto `runSync` would delete
`planMirror`/`applyMirror` (368 lines), remove the un-parenting hazard, and make the
documented flag semantics true at the implementation level.

### Documentation

The manual routed new users through the weaker engine and had no path for the most
common team case — joining a repository that already syncs, where only a personal API
key is needed. Addressed in the companion branch
`claude/linear-integration-review-docs-4sv9xi`: a new `setup-linear` shortcut, a
config-is-shared/credentials-are-personal split made explicit in `tbd-docs.md` and the
skill baseline, and skill routing so "add my Linear key" reaches the shortcut.

### CI

All 7 checks pass on `0ca74c0`. Locally: `pnpm test` — 134 files, 1,973 tests, all
pass; `pnpm run ci` clean.

---

jlevy:
Addressed review `4931891999` in `dcc136dd`.

### Disposition

- **R1 — fixed** (`tbd-8raq`, closed): linked beads beyond `max_nesting` now omit the parent patch when their selected ancestor is skipped, preserving the provider hierarchy instead of hard-failing. Added plan/apply regressions for the exact deep-linked topology.
- **R2 — deferred** (`tbd-dqiq`): consolidating outbound-only sync onto the bidirectional engine crosses reconcile, journal, comments, and durable-base semantics. Keeping this as a P1 design follow-up avoids a partial engine migration in this marker-focused PR.
- **R3 — deferred** (`tbd-bexc`): description-only quarantine plus status-level quarantine counts needs a cohesive adapter/status design. The current fail-closed behavior remains intact until that follow-up lands.
- **R4 — fixed** (`tbd-fqf4`, closed): live QA still tolerates unrelated ambient failures, but now rejects any failure naming either scenario-owned sentinel.
- **R5 — fixed** (`tbd-bhv5`, closed): a `null` description splice no longer reports a push unless an external patch actually occurred; regression coverage added.
- **R6 — addressed by documentation** (`tbd-nfjq`, closed): the integration plan and design doc now explicitly state that a prose push relocates the managed region to the end, while delimiter-only refresh preserves its position. This uses the review's documented alternative without adding another write path in this PR.
- **R7 — fixed** (`tbd-sjta`, closed): an absent `.env` is checked for future ignore behavior and warns when it would not be gitignored; covered for ignored and unignored cases.
- **R8 — deferred** (`tbd-dhfm`): retain legacy-marker reading through the compatibility window, then remove it after one or two shipped releases and a fresh audit.
- **R9 — deferred** (`tbd-dj3t`): `repoUrl` wiring/dead-path cleanup is independent permalink work and remains tracked separately.

### Validation

- `pnpm run ci`: format, docs format, typecheck, lint, build, **134 test files / 1,978 tests** passed.
- Pre-commit and pre-push gates passed.
- Hosted PR checks passed: benchmark, coverage/lint, macOS, Ubuntu, Windows, and secret scanning.
- PR summary and validation count were updated for the review follow-ups.

There were no inline review threads to resolve; the review was submitted as a formal review body, so this top-level disposition is the matching response channel.

jlevy:
Completed the follow-up Linear onboarding and self-documentation audit in `6f588b5e`.

### What changed

- Added `tbd shortcut setup-linear` with two explicit journeys:
  - first-time tbd + Linear repository setup, including prefix, team/project, safe staged projection, and the shared config boundary;
  - joining an already-configured team repository, where the user supplies only a personal key, previews reconciliation, and uses plain `tbd sync` so current bead state is pulled first.
- Updated credential guidance to Linear's current **Settings > Account > Security & Access** flow, required restricted-key permissions/team scope, and the admin control for member API keys.
- Made the safety boundary explicit: never request a raw key in chat, never echo it, require an ignored `.env` before writing it, and keep credentials out of `.tbd/config.yml`.
- Wired natural “Set up Linear” and “Add my Linear key” requests through setup output, `integration status`, welcome guidance, full/brief/minimal skills, generated Claude/portable skill mirrors, README, manual, changelog, and the implementation spec.
- Corrected team setup docs to include `.agents/`, `.claude/`, `.codex/`, and `AGENTS.md` as generated shared surfaces.

### Validation

- `pnpm run ci`: format, Markdown format, typecheck, lint, build, **134 test files / 1,982 tests passed**.
- Built-CLI smoke checks: shortcut listing/rendering, brief skill route, brief prime route, and setup-generated skill content.
- Hosted checks on `6f588b5e`: Ubuntu, macOS, Windows, Coverage & Lint, Benchmark, and DeepSource secret scan all passed.
- Closed and synced `tbd-vnv9`.

The companion documentation branch was incorporated selectively; its generated version stamp was excluded, and its stale `.env`, key-location, and joiner-sync guidance was corrected before landing.

jlevy:
## PR #213 documentation reconciliation

I compared PR #213 commit `f7607469` with this branch at `6f588b5e`, using their shared parent `0ca74c0d` so the now-conflicting stacked-PR base did not obscure the intended changes.

Every relevant documentation improvement from #213 is incorporated here:

| #213 surface | Disposition in #212 |
| --- | --- |
| `setup-linear` shortcut | Incorporated and expanded: tbd initialization, four-case routing, current Linear key location and permissions, never-paste-secret boundary, joiner dry run plus plain `tbd sync`, and the `sync_on_tbd_sync: false` exception |
| `welcome-user` | Incorporated with a separate read-only `integration status --offline` check instead of implying `tbd status` reports integration state |
| `skill-baseline` | Incorporated with corrected offline/live status roles and safe joiner sequencing |
| `tbd-docs.md` | Incorporated with first-time and joiner paths, current credential guidance, and the shared-config/personal-key model |
| Generated `.agents`, `.claude`, and portable skills | Regenerated from the source and extended with Linear/API-key activation metadata |
| `.tbd/config.yml` docs manifest | `setup-linear` is registered |

This branch also covers surfaces #213 omitted: root README onboarding, brief and minimal skill tiers, setup and prime output, integration-status remedies, changelog/spec updates, and drift/onboarding tests.

I did not cherry-pick or merge #213 because that would replace corrected guidance with older variants, including the stale Linear key path, the inaccurate `.env` status description, and a joiner sequence that does not pull current team bead state first.

Validation on `6f588b5e`:

- focused onboarding/setup/status suites: 75 tests passed
- relevant Markdown and generated skill surfaces pass Prettier
- the locally built CLI lists and renders `tbd shortcut setup-linear`
- hosted CI remains green across Ubuntu, macOS, Windows, coverage/lint, benchmark, and secret scanning

PR #212 is the definitive implementation and review target; PR #213 is superseded.

PR #219, opened 2026-08-14
docs: close remaining gaps in setup-github-cli (proxied sessions)

jlevy:
Adds three verified gaps to `setup-github-cli`, and **drops the standalone guideline this
PR originally proposed**.

## Why the pivot

The first draft added a `github-access-layers` guideline documenting the layer model. That
was written while this session was running **tbd 0.3.0**, whose `setup-github-cli` is 85
lines. On **0.6.1** it is 229 lines and already covers the layer model authoritatively —
the `x-github-request-id` test, the credential-substitution behavior, the GraphQL false
negative in `gh auth status`, the ref-scoped git broker, and the verified `NO_PROXY`
recipe.

Two overlapping documents would drift and disagree. Worse, the draft was wrong in ways
0.6.1 already had right: it told readers never to add GitHub hosts to `NO_PROXY` (the
opposite of the verified recipe), and it attributed the *proxy's substituted credential* —
empty `X-Oauth-Scopes`, 8-hour expiry — to the user's `GH_TOKEN`. On the direct channel
that token is a classic PAT with `project`, `read:org`, `repo`, `workflow`. So the
guideline is deleted and only genuinely additive material lands, in the doc that already
owns this topic.

## The three gaps

Each was measured in a live Claude Code Cloud session, and each misled an agent in
practice.

**1. The git broker has two implementations.** The doc described only the remote rewrite to
`http://local_proxy@127.0.0.1:<port>/git/owner/repo`. The other leaves origin as a plain
`https://github.com/owner/repo` and injects credentials through `GIT_ASKPASS` plus
`GIT_CONFIG_KEY_*` / `GIT_CONFIG_VALUE_*` pairs, with `GIT_SSL_CAINFO` pointing at the CA
bundle. An agent checking `git remote -v` for a local endpoint concludes there is no broker
and mis-models the session. Now documents both, with an env-based detection that works for
either.

**2. TLS interception is per-host, and the signing CA names the tier.** In one session
`api.github.com` and `github.com` were intercepted while `raw.githubusercontent.com`
returned GitHub's genuine Let's Encrypt certificate — so generalizing across GitHub
hostnames is unsafe. Provider bundles carry staging *and* production variants of several
roles (local agent proxy, egress gateway, TLS inspection), so a staging CA is not anomalous
by itself; but two sessions signed by different tiers can enforce different policy.

**3. The agent harness is a refusal layer above the network.** A harness permission check
can decline to execute the very `NO_PROXY`-prefixed `gh` invocations this shortcut
recommends. It produces no HTTP status, no `x-github-request-id`, and no response body,
arriving instead as a harness-level denial naming a permission rule. An agent that reads
this as a policy 403 concludes the operation is impossible when it is merely *unapproved* —
opposite responses required. Documented with explicit guidance to report the blocked
invocation rather than rephrase around the check.

Plus three Quick Reference rows, including one pointing at `tbd setup --auto` when a repo's
generated scripts predate its docs.

## The meta-finding

The single most expensive problem was **version drift**, not any of the above. The repo was
pinned to tbd 0.3.0, so every session regenerated a 0.3.0-era `ensure-gh-cli.sh` (123 lines)
that reported `Token may be invalid or expired` — a false negative — while the current
shortcut described entirely different behavior. `tbd doctor` reported "Repository is
healthy" throughout.

After `tbd setup --auto`, the 0.6.1 script (174 lines) reports on the same session:

```
[gh] GH_TOKEN is VALID, but this session's proxy intercepts GitHub API calls
```

That one line would have prevented hours of misdiagnosis. **Worth considering as a
follow-up:** a `tbd doctor` check comparing the generated scripts' originating tbd version
against the running CLI, so stale hooks surface as drift rather than as silent health.
Happy to open that separately.

## Validation

- Verified against **tbd 0.6.1** installed globally (`npm install -g get-tbd@latest`), not
  by reading source — the ensure script was executed and its output captured.
- `flowmark-rs@0.3.1 --auto --check` clean on the modified shortcut.
- Edits are additive within existing sections; no restructuring of the authoritative
  content.

https://claude.ai/code/session_012HZ3HcEYX6fkabniU7ZE8Y


jlevy:
## Feedback on the upgrade process

Ran the full 0.3.0 → 0.6.1 upgrade on a real repo (jlevy/softschema, PR #35) rather than reading the code. Notes below for folding into a future release.

### The upgrade itself is genuinely streamlined

`tbd setup --auto` did everything in one idempotent command: migrated the config `f06` → `f07`, cleaned up 2 legacy hooks, regenerated all scripts and skill mirrors for both Claude and Codex surfaces, updated the AGENTS.md block, and re-synced 45 docs. No manual steps, no prompts, no breakage — 176 downstream tests and lint passed immediately after. It also correctly flagged the config change as needing a commit. **No changes needed to the mechanics.**

### The gap is discovery, not execution

Nothing in tbd tells you an upgrade is available or needed. Verified on 0.6.1:

| Check | Result |
| --- | --- |
| `tbd doctor` version-drift checks | **0** — reported "Repository is healthy" while pinned to 0.3.0 |
| `tbd status` update notices | **0** |
| `tbd upgrade` command | none (`update` is for issues) |

The repo sat three minor versions stale, shipping a `ensure-gh-cli.sh` with a **known false negative** (`Token may be invalid or expired` for a perfectly valid token), and every health signal said fine.

### Why it can't self-correct

`tbd-session.sh` runs `npx --yes get-tbd@<pinned> prime`. The pin is a deliberate supply-chain and consistency control and shouldn't be removed — but combined with the absence of any drift signal, a repo set up on 0.3.0 runs 0.3.0 forever and cannot notice otherwise. The npx fallback compounds it: because `tbd` never lands on `PATH`, an agent can't even run `tbd --version` to spot the mismatch without installing first.

### Six concrete suggestions

1. **`doctor` compares `.tbd/config.yml:tbd_version` against the running CLI** and reports drift as a finding. This alone would have collapsed a multi-hour investigation into one line. Highest value of anything here.
2. **The session hook prints a one-line notice when the pinned version trails latest** — keep executing the pin, just say so. e.g. `[tbd] pinned 0.3.0, latest 0.6.1 — run 'tbd setup --auto' to upgrade`.
3. **`setup --auto` reports the version transition, not just the format transition.** It printed `tbd_format f06 → f07` but never `tbd 0.3.0 → 0.6.1`, which is the more consequential fact.
4. **Point at the changelog after a multi-minor jump.** Crossing 0.3.0 → 0.6.1 materially changed generated *behavior* (the gh diagnosis rewrite). Nothing prompted me to re-read the changed guidance — I found it by diffing.
5. **Distinguish tracked from cached changes in the output.** The run touched 11 committed files plus 45 gitignored cache docs. It flagged the config for committing but not the scripts; an agent has to run `git status` to learn what needs a commit.
6. **Stamp generated scripts with the tbd version that wrote them.** Makes drift visible in the file itself and gives (1) something cheap to compare against.

### One unrelated gap that persists in 0.6.1

`tbd-session.sh` still leaves no binary on `PATH` — the npx fallback executes from a private cache and discards it, while the banner reads `✓ tbd installed (v0.6.1)`. Suggest the fallback `npm install -g get-tbd@<pinned>` (pin preserved, so the supply-chain property holds) and fall back to npx priming only when the install is refused, with a warning that the CLI won't persist.

---

jlevy:
Superseded by #224, which preserves the verified proxy/session evidence while moving implementation to the focused GitHub CLI session-readiness plan tracked by tbd-mslv.

PR #220, opened 2026-08-14
docs(guidelines): add a decision procedure to backward-compatibility-rules

jlevy:
## Summary

This turns `backward-compatibility-rules` into a concise decision procedure:

> Name the real consumer or released data that cannot move in a coordinated change.
> If none exists, choose **DO NOT MAINTAIN**.

Unjustified compatibility is not free insurance. Each retained alias, shim, fallback,
reader, and migration path becomes state every later agent must discover, preserve,
test, and document. Later changes must account for every retained path, so the cost
compounds; because those paths look like real constraints, agents also propagate them.

The decision stays in the always-loaded General engineering guidance, but the final
document is 4.72 kB (~1.3k tokens), down from the earlier 8.9 kB (~2.5k-token) draft.

## What Changed

- Defines compatibility boundaries from deployment topology, release history, adoption,
  and persisted data
- Separates real consumers from hypothetical integrations, co-updated tests, unreleased
  formats, and impossible client skew
- Defines the smallest valid responses: **DO NOT MAINTAIN**, **VERSION + FAIL FAST**,
  **UPGRADE + GATE**, **KEEP DEPRECATED**, **SUPPORT BOTH**, and **MIGRATE**
- Adds plugin and extension APIs plus persisted client state to the specification
  template without repeating option lists for every area
- Requires removal when a boundary ends and treats unsupported compatibility paths as
  review findings

## Validation

- Markdown formatting check
- Full build and bundled-guideline rendering
- CLI documentation golden: 9/9
- Documentation references: 2/2
- Full pre-push gate: formatting, lint, typecheck, build, and 2,005/2,005 tests

### DIFF
--- a/packages/tbd/docs/guidelines/backward-compatibility-rules.md
+++ b/packages/tbd/docs/guidelines/backward-compatibility-rules.md
@@ -1,10 +1,18 @@
 ---
 title: Backward Compatibility Rules
-description: Guidelines for maintaining backward compatibility across code, APIs, file formats, and database schemas
+description: Guidelines for maintaining backward compatibility across code, APIs, file formats, plugin surfaces, and database schemas, including how to decide whether compatibility is needed at all
 author: Joshua Levy (github.com/jlevy) with LLM assistance
 category: general
 ---
 ## Backward Compatibility Guidelines
+
+Backward compatibility is a cost paid to protect a consumer that cannot be updated in
+the same change. When such a consumer exists, that cost buys something real.
+When one does not, the same code buys nothing and cannot even be tested honestly,
+because no input can reach it.
+
+These guidelines cover both halves: how to decide whether compatibility is needed, and
+what to do once you know.
 
 ### Types of Backward Compatibility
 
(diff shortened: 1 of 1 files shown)


jlevy (commented):
## Scope

Reviewed PR #220 (`31e37f1...c0ccaf9`), one changed Markdown guideline, against
`common-doc-guidelines` and the repository's backward-compatibility guidance. Existing
PR discussion contains no unresolved human or inline review findings. CI is green.

## Summary and Verdict

The PR supplies the missing decision rule and should remain in the always-loaded
guideline. It needs a concise editorial pass before merge: the rule is repeated across
several sections, the cumulative cost to later agent work is not explicit, and a few
absolute claims are broader than the evidence supports.

**Verdict: revise, then merge.** No runtime or architectural blockers.

## Findings

### R1 — Medium — State the compounding agent-work tax

`packages/tbd/docs/guidelines/backward-compatibility-rules.md:9`

The introduction calls compatibility a cost, but does not explain its main operational
cost: every alias, fallback, reader, and migration path expands the state that every
later agent must discover, preserve, test, and document. Those paths compose across
changes and can signal false constraints that agents propagate.

**Fix:** Add one short paragraph that names this ongoing, compounding tax and makes it
part of the reason to decide before implementation.

### R2 — Medium — Consolidate repeated rules

`packages/tbd/docs/guidelines/backward-compatibility-rules.md:36`

The deciding rule reappears in “Always Clarify,” “When Backward Compatibility Is
Important,” “Single Application Code,” and “Removing Compatibility Code.” This grows an
always-loaded guideline from about 80 to 203 lines and weakens the punch of the central
rule.

**Fix:** Organize the document once around the compatibility boundary, the smallest
valid response, a copyable requirements template, and removal. Define each idea once.

### R3 — Medium — Use the shipping boundary, not the commit boundary

`packages/tbd/docs/guidelines/backward-compatibility-rules.md:40`

“Cannot be updated in the same commit” is not the decisive boundary when components in
one commit deploy independently. Likewise, “anything that leaves the process” does not
always require versioning, and an unneeded path can be exercised by a synthetic test
even when no production consumer needs it.

**Fix:** Ask whether a real consumer must keep using the old contract after the change
ships. Require evidence from deployment topology, release history, adoption, and
persisted data; recommend versioning only where mismatches can occur and need diagnosis.

### R4 — Low — Describe supported state, not implementation history

`packages/tbd/docs/guidelines/backward-compatibility-rules.md:108`

`KEEP DEPRECATED` requires “comments about past changes,” which conflicts with
`common-doc-guidelines`: docs and comments should describe the present state unless
history serves the reader. Requiring a changelog entry for every internal contract edit
is similarly overbroad.

**Fix:** Define the current deprecated surface, notice, and removal policy; remove the
blanket history/changelog requirements.

## Design Assessment

Keeping the decision procedure in this guideline is the right design because agents
need it whenever compatibility is considered. Splitting it into an on-demand companion
would hide the most important part. Compression is the better way to control context
cost.

## Documentation

No spec, architecture, or development-doc update is needed. This managed source file is
the authoritative bundled guideline.

## False Positives / Do Not Fix

- The added plugin, extension, and persisted-state areas are useful and should remain.
- `VERSION + FAIL FAST` and `UPGRADE + GATE` are useful outcomes if clearly identified
  as alternatives to backward compatibility rather than compatibility modes.

## CI Status

All required checks pass at review time; Cursor Bugbot is neutral/skipped.

jlevy:
All four findings addressed in `ed284be`. `pnpm build`, `pnpm test` (136 files, 2005 tests), and `pnpm ci:quality` pass; DeepSource is grade A on the new range.

**R1 — compounding agent-work tax.** Added to the opening, framed as the operational cost rather than a one-time one: every alias, fallback, reader, and migration path becomes state each later change must discover, preserve, test, and describe; the paths compose, and each one *looks like evidence of a constraint*, so later work preserves it and adds more. That is now the stated reason to decide before implementing.

**R2 — consolidate.** Eight sections to five, 205 lines to 178 (7,780 bytes, down from 8,904). "Always Clarify", "When Backward Compatibility Is Important", and "Single Application Code" were all restating the deciding rule; their distinct content moved into the boundary section, the template, and removal respectively, and the rule is now stated once. The remaining growth over the original 84 lines is the decision procedure itself plus two added areas and two added answers in the template — the parts you asked to keep.

**R3 — shipping boundary.** This was the most substantive correction and it changed the rule, not just the wording. The question is now *"After this change ships, must some real consumer keep using the old contract?"*, with an explicit note that two components changed in one commit still need compatibility if they reach production independently, and that separate repositories may need none if they always ship together. Evidence is now enumerated as deployment topology, release history, adoption, and persisted data. Versioning is recommended where a mismatch *can occur and would be hard to diagnose*, not for "anything that leaves the process."

You were also right that the testability claim was too strong. I had written that unneeded compatibility code "cannot be tested honestly"; a synthetic test can exercise it. That now reads: coverage is not evidence of need, because it shows the branch runs and entrenches it.

**R4 — supported state, not history.** `KEEP DEPRECATED` now asks for the deprecated surface, how consumers are notified, and when it will be removed, instead of "comments about past changes". The blanket changelog requirement is gone; removals belong in release notes when they affect consumers.

**Do-not-fix respected.** The plugin/extension and persisted-client-state areas stay. `VERSION + FAIL FAST` and `UPGRADE + GATE` stay and are now labelled explicitly as alternatives to backward compatibility rather than compatibility modes — they sit under "Choosing the Smallest Valid Response", which opens by saying compatibility is the most expensive answer and these two often serve better.

On the design assessment: agreed, and it closes the open question from the PR description. I had offered to split the decision procedure into an on-demand companion; keeping it here and compressing instead is the better call, since the procedure is needed exactly when compatibility is being considered.

I have also applied R3 and R4 downstream to the repository that motivated this, [jlevy/metabrowser#43](https://github.com/jlevy/metabrowser/pull/43), so its standing answers use the shipping boundary and drop the blanket changelog requirement rather than diverging from these rules.

---

jlevy (commented):
Addressed the review findings in `c9074298`:

- **R1: fixed** — added the explicit ongoing, compounding agent tax and false-constraint propagation.
- **R2: fixed** — consolidated 204 lines to 107, with five focused sections and about 1.3k tokens.
- **R3: fixed** — based the boundary on coordinated shipping and deployment evidence, release history, ownership and adoption, and persistence; versioning now applies only when mismatches can occur.
- **R4: fixed** — deprecations now describe and announce current support and its removal horizon; consumer-visible removals go in release notes, not historical comments.

All required local and GitHub CI checks pass.

PR #222, opened 2026-08-14
External sync and traceability: research, plan, and Phases 1–2 (f08)

jlevy:
Research brief, the four-phase plan it feeds, **the Phase 1–2 implementation, and the live-Linear hardening that followed it**. This PR started docs-only and grew the code; the sections below are ordered accordingly.

- **Research:** `docs/project/research/current/research-2026-08-14-agent-sync-protocol-and-hooks.md`
- **Identity research:** `docs/project/research/current/research-2026-08-14-agent-and-session-identity.md`
- **Plan:** `docs/project/specs/active/plan-2026-08-14-external-sync-and-traceability.md`
- **Release plan:** `docs/project/specs/active/plan-2026-08-15-f08-release-rollout.md`
- **Tracked as:** epic `tbd-dzme`, 32 children labelled `phase-1` … `phase-4`

**Status:** CI green on `6dedc989` across the full matrix including Coverage & Lint; `MERGEABLE`. Local gate: 143 test files, **2,197 tests**, **1,101 golden assertions**. Live-Linear testing is **done** — the runbook ran end to end against a real workspace, found five defects, and three further stability beads have since closed. The mirror settles to nothing on consecutive syncs. Reports: [live testing](https://github.com/jlevy/tbd/pull/222#issuecomment-5305256855) · [follow-up fixes](https://github.com/jlevy/tbd/pull/222#issuecomment-5306360576) · [original runbook](https://github.com/jlevy/tbd/pull/222#issuecomment-5303659936).

## The goal

Open Linear at any moment, see what every agent is doing, and click through — to the governing spec, the research behind it, the GitHub issue that requested it, the PR that implements it, or into the bead browser for the dependency graph. Kept current automatically.

## What the audit found

The sync **mechanism** was essentially complete: the fold is systematic, the engine three-way reconciles against a recorded base with write-ahead intents. Three other things were missing.

**1. A quiet sync is not quiet.** Measured against the bundled mock Linear server:

```
3 beads,  comments=two_way:  5 requests, 3 of 3 bridge records rewritten
10 beads, comments=two_way: 12 requests, 10 of 10 rewritten
10 beads, comments=off:      2 requests, 10 of 10 rewritten
```

A no-op sync rewrote every bridge record (only `synced_at` differs), so it committed, pushed, and ran the full pre-push suite — while reporting `nothingToDo: true` (**F9**). Cost is `2 + N` requests (**F10**).

**2. A bead had nowhere to put external metadata.** Round-trip probe: a bead carrying `refs:` and `docs:` written by a newer tbd came back from an older client's parse-and-write with **both fields silently deleted**. `IssueSchema` parsed in Zod strip mode, and `tbd sync` rewrites beads during ordinary merges — a data-loss vector, not an inconvenience.

**3. The links a reader would follow did not exist.** `repoUrl` and `prUrls` were rendering code with no data behind them (**F11**, **F12**); no bead field could hold a PR (**F14**).

Also: **0 of 1,681 beads had ever carried an assignee** (**F4**).

## The plan

| Phase | Theme | Ships |
|---|---|---|
| **1** | Operational: cheap, correct, habitual sync | Syncing constantly is free; agents claim their work |
| **2** | Schema: room for external metadata (`f08`) | The bead fields everything later renders |
| **3** | Projection: Linear as the entry point | The click-throughs that make review possible |
| **4** | Enforcement and reach | Gates, more agent surfaces, the GitHub decision |

---

# What this PR implements (Phases 1–2)

## `f08`: the last bead-format bump additive metadata should need

`IssueSchema` gains `.passthrough()`, and — the part that is easy to miss — **all four layers** had to change, not just parsing. Serialize was already correct; parse stripped; `mergeIssues` built from `{...base}` plus a fixed key table, so a field present on only one side was lost even once parsing preserved it; and change-detection had to consider unknown keys or a preserved field would never sync. Unknown keys now merge under an explicit `lww` default.

New bead fields on top of it:

- `docs` — supporting repo-relative paths, `role` an open string
- `refs` — external URLs including GitHub issues and PRs, `kind` an open string

Kept separate because a path resolves against a branch and a URL does not. Both union-merge **by identity field** (`path` / `url`) via a new `union_by_key` strategy — plain `union` would list the same PR twice after an edit. Both optional rather than defaulted, so the migration does not write `docs: []` into ~1,740 beads.

CLI: `tbd ref add|rm`, `tbd doc add|rm`.

## Config reshape, and three booleans that should have been enums

The provider block had grown one flat key at a time into five mixed concerns. f08 groups them into `target` / `policy` / `labels` / `identity`. Both spellings are read for one release via `resolveProviderSettings`, because the migration rewrites a **committed** file and a teammate may be on an older branch.

Three fields changed type, each fixing a defect rather than only widening:

- **`labels.create`** (was `create_labels: boolean`) → `none | tbd | all`. `false` meant "never create a label", which silently included the origin labels tbd itself needs — producing a permanent write loop: assert, drop for want of permission, re-assert, forever. `tbd` (the default) creates tbd's own markers and leaves the workspace vocabulary alone.
- **`labels.mirror`** (was `mirror_labels: boolean`) → `none | prefixed | verbatim`. `verbatim` is the mode a boolean could not express.
- **`integrations.on_tbd_sync`** (was `sync_on_tbd_sync: boolean`) → `auto | guarded | report | off`. The boolean welded two decisions together: the folded run passes `assumeYes`, so `true` also **waived** the 20-create/40-update bulk guard. "Fold it in, but stop if it is about to rewrite 200 issues" was unsayable; that is `guarded`. `report` plans and writes nothing.

The migration **translates** these values rather than moving them — carrying `false` verbatim into an enum field would write a config the next release cannot parse.

## Linear label groups

Linear models a group as a label with `isGroup: true`, and a label inside it stores **only the leaf** in `name`, with the group in `parent`. Every label query here selected `{ id name }` and every lookup keyed off the bare name, so:

- `repo/tbd` matched nothing and the adapter created a **flat label literally named `repo/tbd`** — no group, so none of the one-label-per-issue guarantee the group existed to provide
- a root `tbd` and a grouped leaf `tbd` under `repo` collapsed into one entry — this repository's exact shape
- where a real group existed, issue labels read back as bare leaves, so the engine re-asserted `repo/tbd` on every sync forever

`label-groups.ts` introduces the qualified name (`group/leaf`) and the id index, the create path (group first, then child), and the read-back all speak it. Verified against Linear's published schema: `isGroup` and `parentId` are both settable on `IssueLabelCreateInput`.

## `tbd integration setup`

Provisions the tracker-side scaffolding — the origin label, the `repo` group, and the per-repository child inside it. Idempotent; `--dry-run` reports without writing. It derives names through the same `originLabelPlan` the sync uses, so it provisions exactly what sync will later assert rather than a parallel guess.

Sync deliberately does **not** auto-provision: creating label groups restructures a shared workspace and should be something you ran on purpose.

## Origin labels, without breaking quiet sync

Every mirrored item carries the `tbd:sync` marker and its per-repository label so a human can filter agent traffic out (`label is not tbd:sync`) and many repositories can report into one surface. The engine asserts only the **missing** ones, diffed against the remote — putting them in every patch would make every pair look changed, since a non-empty `externalPatch` is what "changed" means, and a settled mirror would write on every sync.

## Agent identity, and `tbd start`

`agid-{ulid}` reusing tbd's internal-ID tier: 80 random bits make uncoordinated minting safe, so no registry round-trip is needed (the display tier's registry check cannot be borrowed for this). `tbd start` claims a bead and refuses to steal a live claim; `tbd whoami --ensure-id`.

## Release

`supportedFormatForVersion` plus a doctor check for the trap found while planning: a format bump does not refresh `tbd_fallback_version`, so the launcher would deterministically install a CLI that refuses the repository. It fails closed with the remedy named, which makes `tbd setup --auto` a **required** upgrade step rather than an optional one.

---

# Live-Linear hardening (added after Phases 1–2)

Pointing the code at a real workspace found five defects CI could not see. Full detail in the [live-testing report](https://github.com/jlevy/tbd/pull/222#issuecomment-5305256855); the short version:

**The root cause of two of them was the mock server being more permissive than Linear.** It allowed duplicate label names and round-tripped markdown verbatim, so the suite validated shapes the API rejects. Both rules are now modelled in `tests/helpers/linear-mock-server.ts`, with a negative test asserting the illegal label shape is refused.

| | Defect | Fix |
|---|---|---|
| 1 | Linear enforces label-name uniqueness **team-wide** — a group does not scope it — so a root `tbd` and a grouped `repo/tbd` cannot coexist. Provisioning was impossible for this repo. | Origin marker moves to `tbd:sync`. `sanitizeRepoLabel` can only emit `[a-z0-9._-]`, so a repo leaf can never contain a colon: the namespaces are disjoint by construction. |
| 2 | The dry run promised 3 creates for an operation Linear would refuse. | Collision detected at plan time, reported `blocked`. |
| 3 | Setup created 2 labels, threw on the third, and printed only `Error: duplicate label name` — a partial write to a shared workspace with nothing on screen saying so. | Failures contained per item; the report survives. |
| 4 | The client discarded Linear's `userPresentableMessage`, which names the label, team and remedy. | Prefers it; classifies on both strings. |
| 5 | **Perpetual write loop.** tbd renders `[name](url)`, Linear stores `[name](<url>)`, and the managed block was compared as a raw string — rewriting 109 of 205 issues on every sync, forever. | Compare normalized. The base hash could never have caught it: hashing strips the managed block, so that region had no check at all. |

Fixing #5 changed hash output (`DESCRIPTION_HASH_PREFIX` → `v7`) and, with the base erased, surfaced three more tracker rewrites plus a genuine defect: **an empty tracker description beat real bead prose**, because tbd's own managed-block splice bumps the remote clock and `tie_break: newest` then picks remote — tbd discarding content by losing to its own write. A side holding nothing is no longer treated as a competing edit.

Convergence, run by run: **109 pushes → 2 → 0.**

## The tracker identifier moved off the bead

Renaming the team `TBD` → `OS` mid-session was an accidental but useful test. Links survived intact — 163 pairs were recognized as updates, not creates — because `extensions.linear.id` is an immutable UUID and nothing reads the human identifier for behaviour. But every bead still carried a stale `key: TBD-nnn` and URL in a file committed to git.

Beads now store `id` and `linked_at` only. `external_key` / `external_url` live on the bridge record, which every sync rewrites and therefore repairs itself. `PERSISTED_LINK_KEYS` still lists the retired keys so beads from an older tbd have them **stripped** rather than frozen. This also deleted the per-update refetch in `mirror.ts`: it cost one request per updated bead (~163 per full sync) and never repaired a settled mirror — exactly the state a rename leaves behind.

## Known gaps, all tracked

Three of the original nine beads have since closed, each verified live:

- **`tbd-cuwr`** — spec permalinks named the branch that ran the sync, so every link would 404 after merge and any cross-branch sync rewrote the whole mirror. Now resolves durable-trunk-first; 75 issues flipped to `main` URLs.
- **`tbd-majw`** — proven that creating a comment advances the issue's `updatedAt`, which was the plan's precondition for the delta-gated comment fetch. Comment sync verified end to end (`commentsPulled: 1`).
- **`tbd-ktji`** — mirror-only `--push` now refreshes the identifier from data the write already returned, at zero extra requests.

Still open, all children of `tbd-dzme`:

| Bead | |
|---|---|
| `tbd-7q6v` (P2) | Tests are load-sensitive: at load average 34, 7 tests across 4 files failed and all passed in isolation. Worse, `test:coverage` is `vitest run --coverage && tryscript run …`, so any vitest flake **silently skips all 1,101 goldens** — which happened during this verification. |
| `tbd-62a5` (P2) | Two goldens invert when 0.7.0 ships, so they "break" on success. |
| `tbd-1emr` (P3) | Sync's duplicate-link failure names a UUID where doctor names the issue key. |
| `tbd-klgh` (P3) | `identity.user_map` is empty, so assignee sync is skipped for `OS-162`/`OS-166` — the two warnings on every run. |
| `tbd-t9hi` (P3) | Research and active-plan docs still say `label is not tbd`; in-code references were corrected. |
| `tbd-b7cy` (P3) | `customViewCreate` and workspace-scoped labels — decisions to make, not gaps to close. |

**At merge time:** run a sync afterwards so the 42 spec links that exist only on this branch move to `main`. Until then they point at a branch that will no longer exist.

## Review notes

- Bugs found and fixed while building, each with a regression test: the `create: false` write loop; the migration moving booleans verbatim into enum fields; `--dry-run` on `integration setup` reading a subcommand option shadowed by the global flag, so the dry run **created all three labels**; the mock server's `IssueCreate` ignoring `labelIds` while `IssueUpdate` honored them; the engine's create path building its own patch without origin labels, guaranteeing a second write per create.
- A separate defect found during the research — duplicate keys in `ids.yml` dropping a bead's mapping — was fixed on its own branch in #232, not here.
- The research proposes and recommends; it deliberately does not decide. Open questions are carried into the spec.
- Caveat, now partly settled: the mock-server probe numbers stood up live, and the 2,500/hr rate limit was confirmed directly from response headers. Whether Linear bumps `updatedAt` on comment creation is **still** unproven — the live cycle pushed and pulled zero comments, so it was never exercised (`tbd-majw`).

---

> [!NOTE]
> **Low Risk**
> Changes are documentation and committed config reshape; Linear sync remains off via `on_tbd_sync: off`, so no runtime sync behavior change from this diff alone.
> 
> **Overview**
> **Repo config (`f08`)** moves this checkout to `tbd_format: f08` and replaces the flat Linear integration block with the grouped shape: `on_tbd_sync: off` (replacing `sync_on_tbd_sync: false`), `target` (`team_key: OS`, `project: tbd`), `labels` (`origin`, `repo: auto`, `mirror: none`, `create: tbd`), and `policy` with explicit `outbound` / `inbound` / `field_sync` clauses (epic-focused outbound mirror, inbound `report`, two-way comments).
> 
> **Tooling** adds `.claude/worktrees/` to `.gitignore` and `.prettierignore` so transient background-agent worktrees are not committed or reformatted.
> 
> **Research docs** add two large briefs—agent/session identity (`agid-{ulid}`, hook stdin, storage model) and keeping agents synchronized with Linear (audit findings, claim/sync protocol, traceability)—and extend the Linear task-surfaces brief with ecosystem/delegation context and cross-links.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6dedc98921ea0d51bda64c725b3a90ba24f0d9ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy:
## Handoff: next phase is local testing against a real Linear workspace

Everything up to the network boundary is built and green. What remains cannot be done without a live API key, so this comment is the runbook for whoever picks it up.

**Branch state:** merged with `main` as of `0054dc0`, no conflicts. `pnpm precommit` green — 142 test files, 2172 tests. The only CI check configured on this PR is Cursor Bugbot (`neutral`); there is no test workflow running here, so local `pnpm precommit` is the gate.

---

### The one thing tbd cannot do for you

A **personal API key**. It is the credential that authorizes everything else, so it cannot authorize its own creation. Linear → Settings → Security & access → Personal API keys.

```bash
echo 'LINEAR_API_KEY=lin_api_...' >> .env    # .env is already gitignored; verify before writing
tbd integration status                       # must print: credential: ✓, team TBD reachable
```

Everything else — labels, label groups, teams, projects, even saved views — is API-creatable. Verified against Linear's published GraphQL schema: `IssueLabelCreateInput` accepts both `isGroup` and `parentId`.

---

### What the first real sync will actually do

Measured against this repository's current beads and the committed policy (`kinds: [epic]`, `statuses: [open, in_progress, blocked]`, `specs: active`, `linked: true`, `max_nesting: 2`):

| | count |
|---|---|
| beads total | 1,745 |
| in the mirror set | 198 |
| already linked to Linear → **updates** | 163 |
| unlinked → **creates** | 35 (all `open`) |

Two consequences worth knowing before running anything:

1. **Both bulk thresholds trip.** 35 creates > 20, and 163 updates > 40. Under `on_tbd_sync: guarded` the run refuses and reports; under `auto` it proceeds. That refusal is the intended first signal, not a failure.
2. **The 163 already-linked beads have no origin labels yet**, because the labels are new in this PR. The first sync backfills `tbd` and `repo/tbd` onto all 163 — a genuine one-time bulk write against a real workspace. Do not run it blind.

Of the 198, 89 are `closed`. That is deliberate, not a selection bug: the `linked` clause keeps an explicitly-linked bead in the set even once it stops matching, so closing a bead does not strand its Linear issue.

---

### Testing phases, in order

Each phase is safe to stop at. Do not skip 2 or 3.

**Phase 1 — provision the labels.** New command in this PR.

```bash
tbd integration setup --dry-run   # writes nothing
tbd integration setup             # creates what is missing
tbd integration setup             # must print "Already provisioned"; proves idempotence
```

Then confirm **in the Linear UI**, because this is the part the mock server cannot prove:
- `repo` exists as a **label group**, not a plain label
- `tbd` is a separate root label
- the repo label is a **child of the group**, and there is no flat label with a slash in its name
- applying a second `repo/*` label to one issue is refused by Linear (the one-per-issue property the group exists for)

**Phase 2 — read-only sync.** Set `integrations.on_tbd_sync: report` in `.tbd/config.yml`, then:

```bash
tbd sync
```

`report` plans and prints, writing nothing to Linear or to beads. Output lines say `would push` / `would create`. Confirm the counts match the table above. **If anything is surprising here, stop** — this is the last checkpoint before real writes.

**Phase 3 — one issue, end to end.** Scoped outbound push of a single bead:

```bash
tbd integration sync --push --bead <one-open-epic-id>
```

Verify in Linear that the issue carries both `tbd` and `repo/tbd`, that `repo/tbd` shows as grouped, and that the description's managed block rendered.

**Phase 4 — the full run.** Expect the bulk guard. `-y` affirms it:

```bash
tbd integration sync -y
```

**Phase 5 — the property that matters most.** Run a full sync twice more:

```bash
tbd integration sync && tbd integration sync
```

The second must report `nothingToDo`. If it reports pushes on a settled mirror, that is the write loop this PR spent most of its effort eliminating, and it is a blocker — capture the report and the diff of one affected bridge record before doing anything else. Two distinct causes were already found and fixed (`labels.create: false` dropping labels forever; grouped labels reading back as bare leaves), so a third would be a new one.

**Phase 6 — turn it on.** Only after 5 is clean: set `on_tbd_sync` to `auto` or `guarded`. Prefer `guarded` here first — this repo's mirror is large enough that a mis-set selector is expensive, and `guarded` refuses rather than writing.

---

### Rollback

- `tbd integration unlink <beads...>` severs pairs and cancels pending writes.
- Labels and the group can be deleted in the Linear UI; nothing in tbd depends on their ids.
- `on_tbd_sync: off` returns to today's state.

---

### Watch for these

- **`tbd_fallback_version` is `0.6.5`, which cannot read `f08`.** The doctor warns and names the remedy (`tbd setup --auto`); it fails closed rather than corrupting anything. This is expected until the 0.7.0 release, and `tbd setup --auto` is a **required** upgrade step — see `docs/project/specs/active/plan-2026-08-15-f08-release-rollout.md`.
- **Comment sync assumes Linear bumps `updatedAt` when a comment is created.** Still unproven against live Linear; the plan makes proving it a precondition for shipping the delta-gated comment fetch. Phase 5 is a good moment to check it.
- **Rate limits.** A quiet sync costs `2 + N` requests. With 198 mirrored beads that is ~200 requests per full sync, so do not loop Phase 5.

---

### Deliberately not built

- **`customViewCreate`** for the "label is not tbd" filter view. It is API-creatable and would be a small addition, but creating shared views in someone's workspace is more invasive than labels and wanted a human decision first.
- **Workspace-scoped labels.** Labels are created team-scoped, matching prior behavior. Cross-team consolidation would want workspace scope (omit `teamId`); that is a config decision, not a code gap.

---

jlevy:
### Correction to the runbook above, and a CI fix

Two updates to the handoff comment.

**1. I was wrong about CI.** The comment said "there is no test workflow running here, so local `pnpm precommit` is the gate." That is false — there is a full matrix (ubuntu/macos/windows × Node 22.12.0/24, plus Coverage & Lint, Benchmark, DeepSource Secrets, Bugbot). It had simply not run on this PR yet when I looked. Treat CI as the gate, not just local precommit.

**2. That matrix immediately caught a real failure the format bump missed.** `scripts/validate-upgrade-package.mjs` asserted the candidate produces `f07`, so every packed-upgrade scenario failed at the first config check. Fixed in `b516b7f`.

The mechanical half was a `CANDIDATE_FORMAT` constant. Deliberately **not** imported from `src` — this script validates a *packed tarball*, so reading the constant out of the working tree would assert the candidate against itself and pass no matter what it produced.

The substantive half is worth knowing, because it is a real property of this release: **the 0.6.3 baseline changed meaning.** It used to be the same-format case — an f07 client that had to keep working against an f07 candidate. Against an f08 candidate it is an older-format client and must now **fail closed**, exactly like the f06 baselines. That is what the bump is for: a pre-f08 client parses beads in Zod strip mode and would delete fields it does not know. The scenario is renamed `latest-published` with `expectOldClientToWork: false`.

**There is deliberately no same-format baseline until 0.7.0 ships.** The first published f08 release restores one.

Verified locally before pushing, all four scenarios:

```
Packed upgrade proof passed: 0.6.3 (f07) -> 0.6.5 (f08)
Packed upgrade proof passed: 0.4.2 (f06) -> 0.6.5 (f08)
Packed upgrade proof passed: 0.5.0 (f06) -> 0.6.5 (f08)
Packed legacy-remote proof passed: 0.4.2 (f06) -> 0.6.5 (f08)
```

Everything else in the runbook stands. The live-Linear phases still need the API key.

---

jlevy:
### CI green, conflicts resolved — and the golden tests caught a live bug

`70a73ad` is `mergeable: clean`, every check green (Bugbot is `neutral`, advisory):

```
success  Benchmark
success  Coverage & Lint
success  Test (macos-latest, Node 24)
success  Test (ubuntu-latest, Node 22.12.0)
success  Test (ubuntu-latest, Node 24)
success  Test (windows-latest, Node 24)
```

Merged `main` (`a209f1d`) in at `0054dc0`. One conflict, `.tbd/config.yml`: this branch had `f08` with `0.6.3` pins, main had `f07` with `0.6.5`. Resolved to **f08 with main's newer pins** — f08 is this branch's contribution, `0.6.5` is the real package version.

### A verification gap worth knowing about

**`pnpm precommit` does not run the golden CLI tests.** `precommit` → `pnpm test` → vitest only. The `tryscript` goldens run solely under `test:coverage`, which is the Coverage & Lint job. Every local "green" I reported was true and incomplete: ten golden cases across six files were failing throughout.

Run `pnpm --filter get-tbd test:coverage` before pushing anything that changes CLI output. Added to the release checklist with the reason.

### One of the ten was a real bug

`on_tbd_sync` was declared `SyncFoldMode.default('auto')`. Zod materializes defaults on every read, so the key was **never** `undefined` and the legacy `sync_on_tbd_sync` branch in `resolveSyncFoldMode` was unreachable dead code.

**Any repository that had set `sync_on_tbd_sync: false` would have silently started syncing on upgrade.** For this repository that is 35 issue creations and 163 label backfills against a live Linear workspace, unasked — the exact scenario the runbook above exists to prevent.

Same trap the flat label keys hit, same fix: optional on the schema, default in the resolver. The unit tests missed it because they called the resolver with hand-built objects, which never carry Zod's defaults; the golden CLI test was the only thing on the real path. Added two tests that go through `IntegrationsConfigSchema.parse`.

### The other nine were genuine golden updates

`f07`→`f08` constants in four files; `--help` gaining `start` / `doc` / `ref` / `whoami`; and `tbd doctor` gaining the Launcher fallback warning, which fires even on a fresh init because `tbd_fallback_version` names a released build that cannot read f08.

Edited by hand, **not** with `tryscript --update`. I tested `--update` against a backup first: it writes observed output verbatim, replacing the `[VERSION]`, `[PATH]`, and `[TIMESTAMP]` redaction patterns with concrete values — which would have made the goldens machine-specific and quietly worthless.

### Two goldens invert at 0.7.0

They will "break" on success, so they are now checklist items rather than surprises:

1. The doctor's `Launcher fallback` warning disappears once the tagged version can read f08.
2. `validate-upgrade-package.mjs` regains a genuine same-format baseline — add 0.7.0 as the `expectOldClientToWork: true` scenario, which no published version can be today.

Full CI-equivalent run before pushing: **142 test files, 2,174 tests, 1,101 golden assertions, 0 failures.** The live-Linear runbook is unchanged and still needs the API key.

---

jlevy:
## Live-Linear testing done: five defects found, all fixed, mirror now settles

The runbook's remaining phases ran end to end against the real workspace (Finterm, team `OS`, project `tbd`). Everything below was found by pointing the code at Linear — none of it was visible to CI, and the reason why is the most important finding here.

### The mock server was more permissive than Linear

Two of the five defects existed because the bundled mock accepted things Linear rejects, so the suite validated shapes that cannot work. Both are now modelled in `tests/helpers/linear-mock-server.ts`:

- **Label names are unique across an entire team**, and a group does not scope them. The mock allowed duplicates.
- **Linear rewrites markdown on store**, wrapping every link destination in angle brackets. The mock round-tripped prose verbatim.

A test suite that is kinder than the API cannot catch this class at all. Both rules are enforced in the mock now, with a negative test asserting the illegal label shape is refused.

### What broke, in severity order

**1. `tbd integration setup` could never succeed for this repository.** Linear enforces label-name uniqueness team-wide, so a root `tbd` and a grouped `repo/tbd` cannot coexist — Linear's own words:

> Label "tbd" already exists in team opensource. Please pick a different name and try again.

This PR anticipated the shape as a *read* problem and fixed the indexing (that half genuinely works — the group-aware lookup correctly distinguishes `tbd` from `repo/tbd`). What it missed is that the shape cannot be *created*. Not specific to this repo: a repository named `docs` would collide with an existing `docs` label just as hard.

**Fix:** the origin marker moves to `tbd:sync`. That namespace was already tbd's — `prefixLabels` mints `tbd:` names and `isTbdOwnedLabel` recognizes them — and since `sanitizeRepoLabel` can only emit `[a-z0-9._-]`, a repo leaf can never contain a colon. The two namespaces are now disjoint by construction rather than by luck.

**2. The dry run gave a green light for an impossible operation.** It reported "3 to create" with no warning, which is precisely the job of the documented safety checkpoint before writing to a shared workspace. Now detects the collision at plan time and reports it `blocked` with the remedy.

**3. Setup half-applied and said nothing.** It created two labels, threw on the third, and printed one line: `Error: duplicate label name`. The throw escaped `provision()` and discarded the report of what had already been written. A partial write you cannot see is the worst failure mode on a shared workspace. Creates are now contained per item.

**4. The error text was gutted.** `client.ts` used the terse `message` and discarded `extensions.userPresentableMessage`, which names the label, the team, and the fix. The duplicate-classification regex also only tested the terse form, so this was not even classified as a duplicate error — though the detailed message would have matched.

**5. A perpetual write loop — the third distinct one in this feature.** tbd renders `Spec: [name](url)`; Linear stores `[name](<url>)`. The managed-block check compared **raw strings**, so it differed on every sync and rewrote **109 of 205 issues forever**. The base hash could never have caught it: hashing strips the managed block before comparing, so that region had no check at all.

Fixing it required normalizing both sides, which changed hash output and so bumped `DESCRIPTION_HASH_PREFIX` to `v7`. With the base erased, chasing the fallout surfaced three more tracker rewrites and one genuine defect:

- angle-bracketed link destinations (any labelled link, not just self-titled ones)
- insignificant leading indentation — CommonMark ignores 1–3 spaces; 4+ still open a code block, and list markers keep their indent since that carries nesting
- a bare email auto-linkified to `[a@b.com](mailto:a@b.com)` **inside a fenced YAML block**
- **an empty tracker description beat real bead prose.** tbd's own managed-block splice bumps the remote `updatedAt`, so under `tie_break: newest` the remote won — tbd discarding its content by losing to its own write. A side holding nothing is no longer treated as a competing edit; that reuses the reasoning `priorityFromLinear` already applies to Linear priority 0.

### Verified end to end

```
C1  provision      2 created; re-run prints "Already provisioned"
C2  single push    1 updated
C3  full sync      push 163, pull 1, create 42, 0 failures
C4  settle x2      nothing but 2 assignee warnings, and no files written
```

The convergence, run by run: **109 pushes → 2 → 0.**

Confirmed in the Linear UI and via the API: `repo` is a real label group, `tbd` is its child (leaf-only, as Linear stores it), `tbd:sync` is a separate root marker, no flat slash-named label exists, and all 205 issues carry both labels. A human-applied `Feature` label survived untouched. The bulk guard refused correctly before `-y` (`42 creates (threshold 20) and 164 updates (threshold 40)`).

Local gate: **142 test files, 2,191 tests, 1,101 golden assertions.** For the two most important fixes I verified the regression test *fails* without its fix rather than trusting a green run — `expected ['ma21'] to deeply equal []` for the write loop, `expected 'FIN-1' to be 'OS-1'` for the identifier refresh.

### Also in this branch: the tracker identifier moved off the bead

Renaming the team from `TBD` to `OS` mid-session was an accidental but useful test. The links survived intact — 163 pairs were recognized as updates, not creates — because `extensions.linear.id` is an immutable UUID and nothing reads the human identifier for behaviour. But every bead still carried a stale `key: TBD-nnn` and a stale URL, in a file committed to git.

Beads now store `id` and `linked_at` only. `external_key` / `external_url` live on the bridge record, which every sync rewrites and which therefore repairs itself; `PERSISTED_LINK_KEYS` still lists the retired keys so beads written by an older tbd have them stripped rather than frozen. This also deleted the per-update refetch in `mirror.ts`, which cost one request per updated bead (~163 on a full sync) and never repaired a settled mirror — exactly the state a rename leaves behind.

### What is deliberately not fixed here

Nine beads, all children of `tbd-dzme`. The two that most affect Linear stability:

| Bead | |
|---|---|
| `tbd-cuwr` (P1) | **Managed block embeds the current git branch.** Spec links render as `blob/claude/linear-sync-agent-hooks-vqatny/…`, so syncing from another branch rewrites every block, and all 205 links 404 once this branch is deleted after merge. Largest remaining source of avoidable writes. |
| `tbd-ktji` (P2) | **Mirror-only push maintains no bridge record**, so `--push` never refreshes identifiers. Self-heals on the next full sync. Cheap: `ISSUE_UPDATE_MUTATION` already returns `identifier` and `url`, but `applyChanges` narrows them away. |

Plus: `tbd-7q6v` (wall-clock perf assertions flake under load, and a flake aborts `test:coverage` before the goldens run), `tbd-majw` (whether Linear bumps `updatedAt` on comment creation — still unproven, no comment traffic occurred), `tbd-1emr`, `tbd-klgh`, `tbd-t9hi`, `tbd-62a5`, `tbd-b7cy`.

---

jlevy:
# Senior engineering review: External sync and traceability (Phases 1–2, f08)

**Scope:** the full PR at `4048997e` (60 files, ~9,000 additions), reviewed as a structural, principal-level pass with emphasis on: design soundness, migration safety, multi-repo workflow flexibility, label clarity in shared Linear workspaces, and groundwork for delegating work to agents.

**Method:** branch checkout and direct code reading, plus four parallel subsystem reviews (migration/config, sync engine/bridge state, Linear adapter/labels, CLI/schema/identity). Every blocker and most majors below were re-verified by hand against the code; findings cite `file:line` on this branch. Paths are relative to `packages/tbd/` unless noted. Tests were not executed (read-only review); all claims are from code reading.

**Bottom line:** the architecture is right, the craftsmanship is unusually high, and the PR should ship. But it should ship after a short hardening pass: two independently-discovered defects will corrupt real data on the first post-upgrade sync in specific states (blocked/deferred beads; workspace labels), the release sequencing as constituted strands agent sessions until 0.7.0 publishes, and two one-line schema omissions will re-create f08's own data-loss class in a place no future format gate can catch. All of these are small fixes now and expensive later.

---

## 1. What is structurally right

Worth stating explicitly, because these decisions will compound and none should be relitigated:

1. **The f08 unknown-key work is done at the right depth.** Preserving unknown keys through all four layers (parse, merge, change detection, serialization) rather than just the schema is the difference between a checkbox and a guarantee, and the PR's own discovery of the fourth layer (`issuesSubstantivelyEqual`) is evidence it was tested from outside. Pairing it with a fail-closed format gate converts silent data loss into a visible upgrade step. This is the most valuable change in the PR.
2. **The config regroup is the right shape, executed carefully.** `target`/`policy`/`labels`/`identity` map cleanly onto the five concerns that were interleaved as flat siblings. Two details show real care: defaults live in `resolveProviderSettings` rather than the schema so the migration's moves stick (`provider-settings.ts:84-95`), and the migration *translates* booleans into enums rather than moving them (`tbd-format.ts:448-451`), each onto the mode that preserves exact prior behavior. Verified: translations are injective and idempotent, `select` folds only when no explicit `policy` contradicts it, and no live code path reads the flat keys around the resolver.
3. **Each new enum fixes a real defect, not a hypothetical.** `labels.create: none|tbd|all` kills the `create: false` permanent write loop; `on_tbd_sync: guarded` makes "fold in, but keep the bulk guard" sayable; `labels.mirror: verbatim` is the mode a boolean could not express. Verified that the four fold postures genuinely differ in behavior (`provider-settings.ts:183-194`, engine-wide dry-run for `report`, bulk guard active for `guarded`).
4. **The label namespace is disjoint by construction.** `tbd:sync` plus `repo/<name>`, with `sanitizeRepoLabel` restricted to `[a-z0-9._-]`, means a repo leaf can never collide with the `tbd:` namespace. The one-label-per-group invariant matching one-repo-per-bead is a structural fit. Filterability (`label is not tbd:sync`) directly serves the "don't disrupt existing Linear workflows" requirement.
5. **Moving `external_key`/`external_url` off the bead onto the bridge record** is correct, and the accidental team-rename test proved it: identity is the immutable UUID; the human-readable key lives in state that rewrites itself. It also deleted a per-update refetch. The quiet-sync guarantee genuinely holds on the default configuration: reconcile produces an empty patch, no intent file, no adapter call, `writeLinkRecordIfChanged` skips, no commit, no push (verified end to end, and pinned by the settled-mirror test).
6. **`tbd integration setup` as an explicit provisioning step** rather than sync auto-restructuring a shared workspace is the right ownership boundary, and the per-item provision report with named holders and remedies is good operational design.
7. **The live-testing discipline is real.** Hardening the mock to Linear's actual behaviors (team-wide label uniqueness, markdown canonicalization) with negative tests means the next contributor cannot silently reintroduce shapes Linear rejects.

The consistent weakness, across all four subsystem reviews: **the engine's invariants stop at the adapter boundary and at the edge of the default configuration.** Labels are where this bites; every engine test pins `mirrorLabels: 'none'`, and the compositions of adapter label behavior (wholesale replace, carrier filter, silent drop of uncreatable names) with engine behavior (assert-only-missing, compare-then-prefix) are where the serious defects below live.

---

## 2. Blockers: fix before this ships

### B1. The origin-label backfill strips status carriers, then corrupts bead status (verified end to end; found independently by two reviews)

The engine asserts missing origin labels via an `ensureLabels`-only patch on otherwise-settled pairs (`sync-engine.ts:673-678`). That is exactly what the first post-upgrade sync emits for **every previously-mirrored item**, since none carry `tbd:sync` yet. In `applyChanges`, an asserting patch triggers a read-modify-write of labels that filters out `tbd:blocked`/`tbd:deferred` (`adapter.ts:293-298`), and `toInput` re-derives carriers only when `patch.status !== undefined` (`adapter.ts:690-698`), which it is not. The write removes the carrier from every blocked or deferred mirrored issue.

Consequences in order: Linear immediately misrepresents those items (a `started` state without `tbd:blocked` reads back as `in_progress`); the bead-side flip is deferred (the recorded post-write `updatedAt` suppresses the echo) but lands the next time anything touches the pair, when reconcile sees remote `in_progress` against base `blocked` and pulls the corruption into the bead as an ordinary no-conflict pull. `deferred` becomes `open` the same way. It also recurs whenever a human deletes an origin label in Linear.

**Fix (surgical):** filter carriers from `preservedLabels` only when `patch.status !== undefined` (they are about to be re-derived; otherwise preserve them). Add the missing regression test: a blocked pair receiving an origin backfill retains `tbd:blocked`. The only current carrier test exercises them alongside a status write (`tests/linear-adapter.test.ts:335`), which is why this was invisible.

### B2. Release sequencing: merging this PR strands agent sessions until 0.7.0 publishes (verified)

The PR commits this repository's `.tbd/config.yml` already migrated to f08 with `tbd_fallback_version: 0.6.5`. `FORMAT_HISTORY.f08.introduced` is `0.7.0`; `package.json` is `0.6.5`. Between merge and the npm publish, every published CLI refuses the repository and the launcher fallback deterministically installs 0.6.5, which also refuses. The named remedy (`tbd setup --auto`) cannot help because it requires a CLI that can read f08, and none is installable yet. The rollout plan states the right order ("publish 0.7.0 first, upgrade repositories second, this repository last"); the merge as constituted violates its own plan.

Two adjacent facts sharpen this:

- **The doctor golden will break on the release's critical path.** `cli-orientation-golden.tryscript.md:110-112` pins the "Launcher fallback" warning that exists only while `package.json` < 0.7.0. The version bump flips the check to ok and the golden mismatches, failing CI on the release PR. The rollout checklist anticipates updating it, but nothing automated ties the three numbers together. Add a tripwire test: `supportedFormatForVersion(PINNED_NPM_VERSION) === CURRENT_FORMAT` (or a golden variant keyed on it), so the coupling cannot regress silently.
- **The mixed-version window is not just a refusal, it re-opens a fixed bug.** Until a teammate pulls the migrated config commit, their 0.6.x client happily co-syncs the shared sync branch: it lacks the empty-description guard, so it can re-enter the prose-wipe this PR fixes (v6/v7 hash prefixes mutually read as "no base"), and it churns `key`/`url` back into `extensions.linear` while new clients strip them. Nothing here is silent (attic entries), but the wipe is destructive remotely.

**Fix (process plus one test):** treat merge and release as one train: merge, tag v0.7.0, publish, run `tbd setup --auto`, commit the refreshed pin, same day. Or split the `.tbd/config.yml` migration into a commit applied after publish. State in the release notes that mixed-version co-syncing during rollout can reintroduce the description-wipe until everyone upgrades. Add the tripwire test above.

### B3. `IssueDoc`/`IssueRef` (and `field_sync.fields`) still strip unknown keys: f08's own bug, one level down, uncatchable later (verified)

`IssueDoc` and `IssueRef` are plain `z.object` (`schemas.ts:177-183, 196-204`) inside the passthrough `IssueSchema`. A future client adding a field inside a ref item (say `refs[].commit`) has it silently deleted by every f08 client on parse-and-write. This is exactly the failure class f08 exists to eliminate, except **no format gate can ever catch it**: both clients are f08-compatible, so the deletion is invisible and permanent. The config side has the same pocket: `FieldSyncClauseSchema`'s inner `fields` object is strip mode (`schemas.ts:571-580`), so a future `field_sync.fields.refs: merge` written without a bump is deleted by this release's config rewrite, and `FORMAT_HISTORY`'s claim that "the nested integration clauses preserve unknown keys too (select, inbound, fields)" (`tbd-format.ts:165`) currently overclaims.

**Fix (three `.passthrough()` calls plus a round-trip test), and only cheap now:** after 0.7.0 ships, fixing the bead-side ones requires another format ceremony. These are the highest-leverage lines in the review. (`PolicyDefinitionSchema`'s strip mode is documented as a deliberate TS7056 trade and can stand, but note it in the format rules: a new *sibling clause* under `policy` still needs a bump.)

---

## 3. Workflow fit: the topology you care about, and the ones nearby

**Mode 3 (one team, several repos, each with its own Linear project) is structurally supported today, verified.** Delta and inbound scans are project-filtered when `target.project` is set (`adapter.ts:479-487`), so repos cannot see each other's items; provisioning composes (the `repo` group and `tbd:sync` are shared team scaffolding created once; per-repo children created idempotently by each repo's `setup`). Mode 1 (team per repo) is structurally isolated and works. This matches the research's table, and the primary workflow needs no further changes this release.

Three qualifications, in decreasing order of importance:

1. **Mode 2 (shared team and project) is not in this PR, but the PR half-claims it.** `isForeignRepoLabel`, the guard that would scope the inbound scan by repo label, is implemented, documented as "used by the inbound scan", tested, and **called by nothing** (`origin-labels.ts:141-154`; the inbound filter at `sync-engine.ts:762-770` has no foreign-repo exclusion). Under default `inbound: report`, a shared scope means each repo advertises the sibling's mirrors as importable on every sync (and `nothingToDo` never settles); under `inbound: auto`, every sibling item fails the claim check and the folded sync exits non-zero, every run. Deferring to Phase 3 is fine and planned; shipping a documented-but-dead function is not. Either wire it (the filter site is a ~5-line change) or remove it, and state plainly in setup docs that shared-project topologies require per-repo projects this release.
2. **A repo rename quietly breaks label convergence in every mode.** `labels.repo: auto` derives from the origin remote (`integration-runner.ts:128-133`). After a rename, every mirrored issue holds `repo/oldname` while sync asserts `repo/newname`; Linear's one-label-per-group rule means the update is either rejected or swaps the label, and the adapter has no group-swap handling, so the outcome is provider-defined and possibly a permanent re-assert loop. Cheap mitigation now: `doctor` flags mirrored items whose `repo/` label disagrees with the resolved name, and the docs say "pin `labels.repo` explicitly before renaming a repo." (Same remap-visibility spirit as the plan's team/project guidance.)
3. **Two lifecycle states never converge to quiet.** Remote archive/trash puts the pair into a permanent `orphaned` report (never `nothingToDo` again, one liveness fetch per sync forever) with no resolution short of manual unlink; and `unlink` leaves the claim attachment, origin labels, and managed block on the tracker item, where the stale claim then blocks re-link and auto-import from any repo without `--force` (`sync-engine.ts:596-615`; `link-guard.ts:17-33`). Both deserve a lifecycle answer eventually (an `orphaned` resolution verb; an opt-in remote cleanup on unlink); for this release a docs note suffices.

**Mapping into existing Linear workflows.** The defaults are right: `field_sync` keeps `labels` and `assignee` local, status maps by state *type* so team state renames cannot break it, and `blocked`/`deferred` ride tbd-owned carrier labels. Two flexibility gaps to plan for, neither blocking:

- **No named-state pinning.** Where a team has several states of one type, the lowest-position one wins (`adapter.ts:530-539`). A team with "In Progress" and "In Review" (both `started`) cannot direct agent work to a specific state, and a team wanting agent traffic in a dedicated "Agent WIP" state has no hook. A post-release `target.states: {in_progress: "Agent WIP"}` override, falling back to type mapping, is the natural extension and needs no format work (target is passthrough).
- **A missing project errors at sync, invisibly at setup.** `resolveProjectId` throws "Linear project not found" (`adapter.ts:150-190`), but `integration setup` neither verifies nor creates the configured project; it provisions labels only. For Mode 3, where someone must hand-create one project per repo, setup should at minimum *report* the project in the provision output (present/missing/blocked), plausibly offer to create it. Cheap, and it converts the most likely first-run failure into a guided step.

**Non-disruption of human workspaces has one real hole: workspace labels (major, fix this release).** The label index is built from team labels only (`ensureMeta` reads `team.labels`), and `applyChanges` preserves current labels *by name*, resolving against that index and silently dropping anything it cannot resolve (`adapter.ts:720-738`). A human-applied workspace (organization-level) label is therefore removed from the issue by any tbd write that touches labels: a status change or the origin backfill itself. This is data loss on human state, in exactly the "don't disrupt their workflow" dimension. Robust fix: preserve unresolvable current labels **by id** (pass through label ids tbd cannot name), which makes tbd writes non-destructive by construction; indexing org labels is a bonus. The mock has no workspace labels, so model one before trusting any fix.

---

## 4. Delegation groundwork: right architecture, one gap that defeats its purpose

The identity design points the right way: `agid-{ulid}` minted locally without coordination, non-person-identifying derived names, `tbd start` as the claim verb symmetric with `close`, `assignee: local` keeping tracker assignees out of beads, and the inbound `when.delegate` design (Phase 3) aligned with how Linear's agent ecosystem actually triggers work. The layering for future assignment flows is sound and needs no format changes, which is what f08 was for.

But the claim protocol as shipped does not deliver its headline guarantee. The guard compares `assignee !== identity.name` (`start.ts:123-135`), and the default name is `<harness>@<host>` (`agent-identity.ts:134-150`). **Two agents in parallel worktrees on one machine, tbd's flagship deployment shape, resolve to the same name:** the second agent is told "already yours" about a claim it does not hold, and the minted agid, which exists precisely to make uncoordinated identities distinct, never lands in a claim. There is also no staleness path: a crashed agent's claim blocks its bead until a human intervenes, which will matter more once Phase 4 gates nag agents about claimed work.

Recommendations in leverage order: (1) make the claim value distinguishable, e.g. suffix the derived name with the agid's short tail (`claude@host-3f9k`); assignee is an open string, so this is release-safe; (2) until then, document `TBD_AGENT` as the parallel-worktree convention; (3) design the staleness answer (even `tbd start --steal <id>` gated on claim age) before Phase 4 enforcement lands. Also worth fixing while in the file: `tbd start` silently flips `blocked`/`deferred` beads to `in_progress` with no dependency check, and silently reassigns an `open` bead that carries someone else's assignee (`start.ts:113-141`); both contradict the care shown by the live-claim guard.

---

## 5. Migration and compatibility assessment

The core is genuinely good; the verified findings are at the seams:

1. **The migration itself: lossless, idempotent, behavior-preserving (verified).** Boolean-to-enum translations are injective; `sync_on_tbd_sync: true → auto` exactly reproduces the old fold behavior including `assumeYes`; idempotence guards are everywhere; conflicting old+new spellings are left in place with new-first read precedence.
2. **The legacy-spelling retirement plan is currently not executable as stated.** Three things contradict the "read both for one release" window: flat `max_nesting` is *deliberately* left in place when `policy` is a preset name (the resolver reads it as a fallback indefinitely; dropping the flat read later silently reverts affected repos to 2); the tool's own inert-status help still prints the flat shape (`integration.ts:100-106`); and `tbd-docs.md` and `setup-linear.md` still teach flat `team_key`/`user_map`. Users will keep seeding legacy spellings into f08 configs from the tool's own guidance. **Fix this release:** update the help text and the two docs, add a `doctor` lint flagging legacy flat keys in f08 configs (also covers the silent both-spellings-disagree case, which today has no warning anywhere), and make an explicit decision about `max_nesting` retirement.
3. **The onboarding shortcut is the biggest doc gap:** `setup-linear.md` teaches a hybrid config (flat `team_key` next to `policy: default`), references retired `mirror_labels: true` (line 113), and never mentions `tbd integration setup`, the provisioning step this PR built specifically so shared workspaces are not restructured by a routine sync. First-run agents following it will write mixed-era configs and skip provisioning, then hit the uncreatable-label loop below. Same rank as a code defect for release stabilization.
4. **Known trade, worth its release-notes line (as the rollout plan already says):** the persisted migration diff is much larger than the migration function's output because `ConfigSchema.parse` freezes defaults into the committed file (a migration-created `labels` group gains `origin: true`, `repo: auto`, `create: tbd`; a select-fold policy expands to the full tree). Reviewers of migration commits will see settings nobody chose; future default changes will not reach migrated repos.
5. Small fail-open edges in `supportedFormatForVersion` (lexicographic comparison breaks at f100; unanchored version regex; prerelease pins count as their base) are fine to leave with a comment.

---

## 6. Remaining verified majors (with fixes)

M1. **Any uncreatable origin label is a permanent per-issue write loop.** If the origin or repo label cannot be created (group name occupied by a plain label; Linear declines; or the custom-name case below), `resolveLabelIds` silently drops it (`adapter.ts:728-729`), the issueUpdate still executes, and the next sync recomputes `missingOriginLabels` non-empty: one wasted write per linked issue per sync, forever, and the mirror never settles. The `create: none` short-circuit (`integration-runner.ts:90-105`) covers only the global off switch, and sync never consults provision's `blocked` knowledge. Fix: when an assert-labels resolution comes back empty for a label, record it (per run or in bridge state) and stop re-asserting; surface it as a warning naming `tbd integration setup`.
M2. **A custom `labels.origin: <name>` can never be created at all.** `mayCreateLabel` calls `isTbdOwnedLabel(name)` without passing the operator's override (`adapter.ts:1008` vs `origin-labels.ts:168`), so under default `create: tbd` a non-prefixed custom marker is uncreatable, guaranteeing the loop above for everyone who uses the documented override. Fix: thread the resolved origin/repo names into the ownership check. (This is the same defect class the enum redesign just fixed; it deserves a test asserting a custom origin label round-trips.)
M3. **Mirror modes (`prefixed`/`verbatim`) oscillate once origin labels exist.** Reconcile compares the bead's unprefixed labels against the remote's prefixed-plus-tbd-owned set, and prefixing happens after comparison (`sync-engine.ts:680-691`); under the default `labels: local` flow rule the pair pushes wholesale label replacements every sync, alternately stripping and re-asserting origin labels (and carriers, B1). Non-default config, but supported and now load-bearing against origin labels. Fix: canonicalize the remote set before compare (strip prefix, exclude tbd-owned), or gate `mirror != none` with a clear config error this release. No engine test runs with `mirrorLabels != 'none'`; add one either way.
M4. **Dry-run misses the leaf collision exactly in the first-run case it was built for.** The group-unavailable branch `continue`s before the team-wide-uniqueness check (`adapter.ts:914-925` vs `929`), so on a fresh team with a root `tbd` label, `setup --dry-run` promises `repo/tbd: would create` and the real run blocks. Contradicts the adapter's own "report it the same way in a dry run and a real one." Fix: hoist the holder check above the group branch.
M5. **Retry backoff stalls up to 60s on any transient error.** `backoffMs` returns `min(untilReset, 60s)` whenever a reset header has ever been seen (`client.ts:127-138`); the exponential is effectively dead code, contradicting its comment. One 500 becomes a minute-plus of silent stalling. Fix: exponential for non-rate-limit retryables; wait-for-reset only on RATELIMITED. (Unit tests inject a no-op sleep, which is why this was invisible.)
M6. **A duplicate-label race fails the whole issue write.** `ensureLabel` never catches the `LinearDuplicateIdError` the client synthesizes for "already exists" (`client.ts:210-212`), so a concurrent provision or stale index fails every asserting bead that run, against the documented drop-one-label intent. Self-heals next run; still noisy and misattributed. Fix: on duplicate, refetch the index once and reuse the existing id.
M7. **`ref rm`/`doc rm` are undone by any concurrent edit of the same list.** `unionArraysByKey(a, b)` never sees the merge base (`git.ts:628-653`): base `[R]`, A removes R, B adds S, merge is `[R, S]`; mutual removals cancel; silent, no attic entry, no test. Since concurrent adds are the advertised workload, resurrection is the common case. For this release: document the semantics (as was done for labels/#155) and pin a test; the real fix is a base-aware union, which is additive and needs no format change.
M8. **Ref/doc identity is the verbatim string.** No URL normalization (trailing slash, case, fragment, query: five spellings of one PR are five entries), no path normalization, and docs accept `../` traversal and absolute paths despite the repo-relative contract (`ref.ts:207-214`; schema-side `url` is only `min(1)`). Normalize at the CLI boundary **now**; after users accumulate variants, tightening the identity key becomes a data migration.

---

## 7. Minors worth tracking (consolidated)

- Journaled `create_issue` intents omit `ensureLabels` (`sync-engine.ts:1017-1031` vs `1312-1318`): crash-replayed creates arrive unlabelled and pay the extra backfill write creates were designed to avoid.
- `nothingToDo` conflates "wrote nothing" with "nothing to report" (`sync-engine.ts:371-386, 1422-1436`): standing mapping warnings (unmapped assignees in the delta), `skippedOutbound`, `importable`, and permanent `orphaned` entries each keep a genuinely quiet mirror from ever reporting quiet. Dry-run and real-run `nothingToDo` also disagree (orphans counted in one, not the other).
- A missing or damaged bridge record silently converts remote edits into overwrites (base seeds from the current remote snapshot, so `remoteChanged` is false by construction; correct for pre-engine links, but the same path swallows remote edits after record corruption).
- Mirror-only `--push` workflows now store no human identifier anywhere (bead stripped by design, no bridge record in that mode): a real, if small, regression vs main for the fold-off crowd; self-heals on first full sync (tbd-ktji is correctly scoped).
- Normalization applies markdown rules inside fenced code blocks (`bridge-state.ts:56-71`): symmetric, so no loops, but a purely cosmetic remote edit inside a fence never pulls. Deliberate trade; document it.
- Managed-block canonicalization coverage is one-of-four: only the angle-bracket rewrite is modelled; Linear escaping (`\_` in link labels) or list renumbering inside the block would loop and is a live-only discovery. Worth one live probe with an underscore-bearing spec filename before trusting settle.
- `git commit` failures in the runner are swallowed (`integration-runner.ts:512-514, 567-569`), including disk-full; fine for retry semantics, bad for diagnosis.
- Legacy flat `repo/tbd`-named labels from the old bug are blessed as `present` by qualified-name lookup and never repaired; an empty `repo` group is created beside them.
- Workflow states are fetched with `first: 50` and no pagination (`queries.ts:29`); labels paginate, states silently truncate.
- A repo name that sanitizes to empty falls back to `repo`, yielding `repo/repo` whose leaf collides with its own group team-wide: permanently blocked with no hint.
- `tbd doc` (bead attachments) and `tbd docs` (docs cache) are one character apart, both with `add` subcommands taking paths.
- Bulk `tbd start` aborts mid-batch on an unresolvable id after earlier claims persisted, skipping the summary.
- Case sensitivity of Linear label uniqueness is unverified live; plan-time collision detection is exact-case, so a case-variant collision degrades to a blocked apply with the raw API message.
- Cross-machine outbound double-create (two clones' first sync racing) is neither prevented nor detected post-hoc; the inbound twin at least hits the duplicate quarantine. Pre-existing, unchanged; note for Phase 4.

---

## 8. Test additions this review argues for

The suite is strong where it looks (2,191 tests; the four-layer f08 round-trip and settled-mirror tests are exactly right). The gaps are systematic, not random: **every engine test pins `mirrorLabels: 'none'`, the mock omits workspace labels and group exclusivity, and no CLI-handler tests exist for the new commands.**

Priority order:

1. Blocked pair receives origin backfill, carrier retained (B1 regression).
2. `supportedFormatForVersion(PINNED_NPM_VERSION) === CURRENT_FORMAT` tripwire (B2).
3. Round-trip: unknown key inside a `ref` item and inside `field_sync.fields` survives parse-write-merge (B3).
4. Engine run with `mirrorLabels: 'prefixed'` plus origin labels: converges, no oscillation (M3).
5. Custom `labels.origin` name: creatable, settles (M2).
6. `ref rm` vs concurrent `ref add` on another clone: pinned expected semantics, whatever is decided (M7).
7. Provision dry-run with root-label collision and missing group: reports `blocked`, not `would create` (M4).
8. Mock: one workspace label (assert preserved through a tbd write), group exclusivity on apply, comment-driven `updatedAt` bump (also unblocks tbd-majw/tbd-iqgm).
9. Handler-level tests for `ref`/`doc`/`start`/`whoami --ensure-id`.
10. E2E: genuinely-old flat config through `validate-upgrade-package` (boolean-to-enum path is unit-tested only).

---

## 9. Recommended path to a stable release

**In this PR (small, surgical, all verified):**
1. B1 carrier-strip condition + regression test.
2. B3 passthrough on `IssueDoc`, `IssueRef`, `field_sync.fields` + round-trip test.
3. M2 thread custom origin/repo names into `mayCreateLabel`; M4 hoist the collision check; M5 backoff split; M6 duplicate-create recovery. Each is a few lines in files this PR already owns.
4. Wire `isForeignRepoLabel` into the inbound filter or delete it; one sentence in setup docs either way.
5. URL/path normalization at the `ref`/`doc` CLI boundary (M8), before variants accumulate.
6. `setup-linear.md` + `tbd-docs.md` + inert-status help refresh to f08 shapes, including the `tbd integration setup` step and the `label is not tbd:sync` filter recipe.

**In the release train (process):**
7. Merge and tag v0.7.0 as one motion; refresh and commit the fallback pin immediately (B2); release notes cover the mixed-version window, the config-verbosity trade, and mirror-mode caveat if M3 is deferred.
8. B2 tripwire test; doctor lint for legacy flat keys in f08 configs.
9. Permalink candidates main-first (`integration-runner.ts:169`): with `main` preferred when the file exists there, the 205 links self-heal after merge instead of 404ing (tbd-cuwr's worst half, one line).

**Post-release (Phase 3 alignment, groundwork already right):**
10. Claim distinguishability (agid tail in claim names) + staleness policy, before Phase 4 gates.
11. Workspace-label preserve-by-id (§3) if not taken in-PR; it is the strongest candidate to pull forward since it touches human data.
12. `target.states` named-state overrides; project check (and optional create) in `integration setup`; base-aware union for `refs`/`docs`; orphan/unlink lifecycle verbs; doctor check for `repo/` label vs renamed remote.

**Deferred items I reviewed and agree with deferring:** tbd-ktji (mirror-only bridge records; self-heals, correctly scoped), tbd-majw/tbd-iqgm (comment `updatedAt` proof before delta-gating), the GitHub adapter decision (refs cover the traceability need), and the no-webhook/no-daemon stance (polling plus cheap quiet sync is the right complexity budget).

---

*Review method note: four parallel subsystem reviews (migration/config, engine/bridge, adapter/labels, CLI/identity) with independent convergence on B1 from two of them; every blocker and majors M1-M8 re-verified by hand against branch `4048997e`. Static review only; no tests were executed.*

jlevy:
## Three more fixes on top of the checkpoint, all verified live

Since the live-testing report above, three beads closed. CI is green on `6dedc989` across the full matrix including Coverage & Lint (the job that runs the 1,101 goldens), and the PR is `MERGEABLE`.

### `tbd-cuwr` — spec permalinks named the branch that ran the sync

The largest remaining source of avoidable Linear writes, and the one with a deadline attached: the managed block rendered spec links as `blob/claude/linear-sync-agent-hooks-vqatny/docs/…`, so

- every one of those URLs would 404 the moment this branch is deleted after merge, and
- because the link is *inside* the managed block, running a sync from any other branch changed the block for every bead — rewriting the whole mirror for no reason at all.

Candidates are now ordered durable-first (`origin/main`, `main`, `master`), with the working branch kept only as a last resort so an unmerged spec stays linkable while in flight. Each trunk is checked against its remote-tracking ref before the local one, since a local `main` that hasn't been pulled may not contain a spec the remote already has; the URL names the plain branch either way, which is what a reader opens.

**Verified live:** 75 issues flipped from a feature-branch URL to a `main` URL, and two consecutive syncs settled to nothing afterwards. The remaining 42 are specs that exist only on this branch — they move to `main` on the first sync after merge, which is the designed fallback rather than a gap.

The regression test drives `resolveSpecLinks` end to end rather than the helper alone, because the ordering that actually matters is built by the caller. It fails with the old ordering (`blob/cla…` where `blob/mai…` was expected).

### `tbd-majw` — the comment-sync assumption is now proven, not assumed

The plan made this a precondition for shipping the delta-gated comment fetch, and it had stayed open because the earlier cycle pushed and pulled zero comments. Settled directly against live Linear:

```
before          updatedAt = 2026-08-16T02:11:54.766Z
comment created             2026-08-16T02:30:58.550Z
after           updatedAt = 2026-08-16T02:30:58.529Z
```

Creating a comment **does** advance the issue's `updatedAt`, so the watermark prefilter will surface an issue whose only change is a new comment. Inbound comments are not missed.

One wrinkle worth recording: the issue's `updatedAt` lands 21ms *earlier* than the comment's `createdAt`, so nothing should assume `comment.createdAt <= issue.updatedAt`. Far inside the existing `WATERMARK_OVERLAP_MS` (10 minutes) re-fetch tolerance, but it is a real inversion.

End-to-end in the same run: `tbd integration sync` reported `commentsPulled: 1`, then settled. That was the last path in the comment sync neither a test nor a live run had exercised.

### `tbd-ktji` — mirror-only push now refreshes the identifier, at no extra cost

`--push` uses the one-way mirror, which keeps no bridge state, so after the identifier moved to the bridge record a push-only workflow noticed a team rename nowhere. Transient (the next full sync repaired it) but a mode behaving differently from its sibling for no reason a user could predict.

`ISSUE_UPDATE_MUTATION` already selected `issue { id identifier url updatedAt }` and `applyChanges` was throwing two of those away — so the fix costs **zero** extra requests. Deliberately a new `onIdentifier` callback rather than reusing `onLinked`, because `onLinked` writes the bead and the entire point of moving the identifier off the bead was that a remote rename must not churn `version` and `updated_at` in git.

**Verified live:** a scoped `integration sync --push --bead tbd-ktji` populated `external_key: OS-245` on the record, bead untouched.

### A test-infrastructure finding worth flagging

`tbd-7q6v` grew teeth. The suite is load-sensitive well beyond the two timing assertions I first noticed — at load average 34 a single run failed **7 tests across 4 files** (common-dir-layout-doctor, git-remote, setup-hooks, worktree-health), every one passing in isolation immediately after, and the whole suite passing at load average 7.

The second-order problem is worse than the flakes: `test:coverage` is `vitest run --coverage && tryscript run …`, so **any** vitest flake skips all 1,101 golden assertions silently. That happened during this verification. Fixing the `&&` is worth doing independently of the budgets.

### Remaining, all tracked under `tbd-dzme`

| Bead | |
|---|---|
| `tbd-7q6v` (P2) | Load-sensitive tests; goldens silently skipped when vitest flakes |
| `tbd-1emr` (P3) | Sync's duplicate-link failure names a UUID where doctor names the issue key |
| `tbd-klgh` (P3) | `identity.user_map` empty, so assignee sync is skipped for `OS-162`/`OS-166` — the two warnings on every run |
| `tbd-t9hi` (P3) | Research and active-plan docs still say `label is not tbd`; in-code references corrected |
| `tbd-62a5` (P2) | Two goldens invert when 0.7.0 ships, so they "break" on success |
| `tbd-b7cy` (P3) | `customViewCreate` and workspace-scoped labels — decisions, not gaps |

**One thing to do at merge time:** run a sync afterwards so the 42 branch-scoped spec links move to `main`. Until then they point at a branch that will no longer exist.

Bugbot has not produced a review on this PR — every run since 2026-08-14 reports `Bugbot couldn't run - usage limit reached`, so it is `NEUTRAL` rather than passing. There is no automated review to address here, and no human review comments either.

---

PR #227, opened 2026-08-14
fix: make a sync with nothing to do actually cost nothing

jlevy:
First implementation PR from the [external sync and traceability plan](https://github.com/jlevy/tbd/pull/222) — the Phase 1 items that are unambiguously safe and need no product decision.

Beads: `tbd-774m`, `tbd-7okw`, `tbd-8ot8`, `tbd-1uep`, `tbd-czhw`.

## Why

Two defects compounded to make frequent syncing unaffordable, and each made the other worse. Measured before this change, a *settled* mirror re-synced with nothing changed anywhere:

```
3 beads,  comments=two_way:  5 requests, 3 of 3 bridge records rewritten
10 beads, comments=two_way: 12 requests, 10 of 10 rewritten
```

Every one of those rewrites differed only in `synced_at`.

## What changed

**A quiet sync now writes nothing.** The apply loop reconciles every linked pair on every run and arrives at the same answer for a settled pair, then wrote the bridge record unconditionally with a fresh `synced_at`. So a sync that changed nothing produced one file write per linked bead → a commit → a push, all while reporting `nothingToDo: true`. Records now write only when something other than `synced_at` differs.

`synced_at` is **kept**, not dropped — the plan originally preferred dropping it, but `pickNewestLinkRecord` uses it as a merge tiebreaker when two machines observe the same item. It is load-bearing; only its role as a write trigger was wrong.

**The sync-branch push no longer runs the parent repo's `pre-push` hook.** tbd already passes `--no-verify` on sync-branch *commits*, deliberately, so bead bookkeeping never triggers lefthook/husky (`git.ts` says so). The push made no such promise, so in this repository every sync paid for the full test suite — and the retry loop paid again on each non-fast-forward. Both sync-branch push sites now pass `--no-verify`. This does not affect pushes of your own branches.

**`tbd sync --issues` no longer excludes the tracker silently.** Easy to hit by accident, since the tracker *is* issues. It now says it skipped, and how to reconcile.

**Docs.** The "~10% of a repository" figure for the default policy was measured against every bead *including closed ones*. Status gates the selection, so closed beads can never be mirrored, and that denominator understates what actually lands in the tracker: here the same policy is ~9% of all 1,726 beads but **52% of the 291 still open**. Reframed against open work, with a pointer to the dry run and a note that `max_nesting` skips make the issue count smaller than the selection count. Also fixed a comment claiming the integration fold is off by default — it is on.

## Tests

Both properties are pinned by behavior, not by asserting on argument strings:

- A settled mirror re-synced with no changes must leave `bridge/<provider>/links/` **byte-identical**. Verified to fail before the fix.
- A sync-branch push must succeed against a repo whose `pre-push` hook always rejects — with a companion test proving that hook is live, so the first cannot pass vacuously.

Full suite: 2,026 tests across 137 files, all passing.

## Deliberately not in this PR

- **Comment delta-gating** (`tbd-iqgm`) — the biggest remaining win (`2 + N` → flat 2 requests), but it assumes Linear bumps `updatedAt` on comment creation. The plan makes proving that in live QA a precondition, and I have no key here.
- **Hook PATH order** (`tbd-fnwc`) — #226 just rewrote these launchers and added a test asserting the current PATH order, so changing it is now a conversation rather than a fix.
- **`tbd sync --push` semantics** (`tbd-71am`) — needs a decision: guard it, or make it a full reconcile.
- **`tbd start` / agent identity** (`tbd-mnci`, `tbd-f39i`) — additive and safe, but a bigger surface with an open question (sync by default?). Better as its own PR.

---

> [!NOTE]
> **Medium Risk**
> Changes core sync, integration bridge persistence, and git push behavior for the sync branch; well-tested but affects every frequent `tbd sync` with Linear enabled.
> 
> **Overview**
> Makes **settled integration syncs** stop rewriting bridge link YAML when only `synced_at` would change, so a quiet reconcile no longer forces a sync-branch commit and push (including through heavy parent `pre-push` hooks).
> 
> **Sync-branch git pushes** now pass **`--no-verify`** in `pushWithRetry` and `pushFreshOrphan`, matching sync-branch commits: bead bookkeeping is not gated by the main repo’s hooks.
> 
> **`tbd sync --issues` / `--docs`** (and dry runs) emit a default-visible **notice** when an enabled tracker would normally run but surface flags excluded it, with optional structured `{ skippedSurfaces: ['integrations'] }` on **stderr** under `--json` so stdout stays a single result document.
> 
> **Docs** reframe default Linear policy sizing against **open** beads (not all history), add dry-run / `max_nesting` guidance, and correct the comment that the integration fold is **on** by default when a provider is enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7fea4ce966042549e29856d3683f6bff9422cee5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tbd/docs/shortcuts/standard/setup-linear.md
+++ b/packages/tbd/docs/shortcuts/standard/setup-linear.md
@@ -85,8 +85,19 @@ integrations:
 
 Leave `policy: default` unless the user asks for something else.
 It selects open epics plus anything whose `spec_path` points into `specs/active/`—the
-right starting point for “track our specs and major work”, and roughly 10% of a typical
-repository’s beads.
+right starting point for “track our specs and major work”.
+
+**Size that selection before the first sync, because it is easy to underestimate.**
+`spec_path` propagates from a parent bead to its descendants, so the spec clause selects
+every descendant of every epic carrying a live spec.
+In a spec-driven repository that can be a large share of open work rather than a small
+one. Two things to check with the user when the preview looks bigger than expected:
+
+- `tbd --dry-run integration sync --push` lists exactly which beads would be mirrored.
+- Beads nested deeper than `max_nesting` (default 2) *within the selected set* are
+  reported as skipped rather than created, so the number of Linear issues is normally
+  smaller than the number of selected beads.
+  Both numbers appear in the dry run.
 
 **Then run `tbd setup --auto` before committing.** Writing the block leaves the
 repository at whatever format it was on; setup stamps `tbd_format: f07`, which is what
--- a/packages/tbd/docs/tbd-docs.md
+++ b/packages/tbd/docs/tbd-docs.md
@@ -1668,9 +1668,10 @@ would ask about in a status meeting, plus whatever carries a live plan spec.
 **Judge that against *open* work, not against every bead you have ever closed.** Status
 gates the selection, so closed beads can never be mirrored and including them in the
 denominator makes any policy look small.
-In this repository the default policy selects 151 beads: about 9% of all 1,726 beads,
-but **52% of the 291 that are still open** — and the second number is the one that
-predicts what lands in the tracker.
+Measured in this repository on 2026-08-14, the default policy selected 151 beads: about
+9% of all 1,726 beads, but **52% of the 291 that were still open**. Both numbers move as
+a repository grows; the second is the one that predicts what lands in the tracker.
+Run `tbd --dry-run integration sync --push` for your own current count.
 
 The default policy is *open epics, or anything whose `spec_path` points into
 `specs/active/`*. Kind and spec are **alternatives**, not requirements, so both “every
--- a/packages/tbd/docs/tbd-docs.md
+++ b/packages/tbd/docs/tbd-docs.md
@@ -1662,10 +1662,15 @@ Both are inert when nothing is enabled.
 
 ### What to sync
 
-**Mirror the shape of the work, not the work itself.** A useful rule of thumb is around
-**10% of your beads**: the epics someone would ask about in a status meeting, plus
-whatever carries a live plan spec.
-In this repository that is 122 of 1,372 beads, or 8%.
+**Mirror the shape of the work, not the work itself.** The target is the epics someone
+would ask about in a status meeting, plus whatever carries a live plan spec.
+
+**Judge that against *open* work, not against every bead you have ever closed.** Status
+gates the selection, so closed beads can never be mirrored and including them in the
+denominator makes any policy look small.
+In this repository the default policy selects 151 beads: about 9% of all 1,726 beads,
+but **52% of the 291 that are still open** — and the second number is the one that
+predicts what lands in the tracker.
 
 The default policy is *open epics, or anything whose `spec_path` points into
 `specs/active/`*. Kind and spec are **alternatives**, not requirements, so both “every
--- a/packages/tbd/src/cli/commands/sync.ts
+++ b/packages/tbd/src/cli/commands/sync.ts
@@ -148,18 +148,18 @@ class SyncHandler extends BaseCommand {
     //
     // This must survive the default invocation: `info()` is verbose-only, so
     // reporting it there would leave the ordinary run exactly as silent as
-    // before. `data()` gives the structured form under --json and the
-    // default-visible `notice()` otherwise. Dry runs report it too — "what
-    // would this do?" has to include "not the tracker".
+    // before. `notice()` gives the default-visible text form and writes its
+    // structured diagnostic to stderr under --json, preserving stdout's
+    // single-result-document contract. Dry runs report it too — "what would
+    // this do?" has to include "not the tracker".
     if (!syncIntegrations && !options.status) {
       const config = await readConfig(tbdRoot);
       if (!integrationsInert(config) && config.integrations?.sync_on_tbd_sync !== false) {
-        this.output.data({ skippedSurfaces: ['integrations'] }, () => {
-          this.output.notice(
-            'Skipping external trackers for this run (surface flag given). ' +
-              'Run `tbd sync` or `tbd sync --integrations` to reconcile them.',
-          );
-        });
+        this.output.notice(
+          'Skipping external trackers for this run (surface flag given). ' +
+            'Run `tbd sync` or `tbd sync --integrations` to reconcile them.',
+          { skippedSurfaces: ['integrations'] },
+        );
       }
     }
 
--- a/packages/tbd/src/cli/commands/sync.ts
+++ b/packages/tbd/src/cli/commands/sync.ts
@@ -145,13 +145,21 @@ class SyncHandler extends BaseCommand {
     // Narrowing to a surface excludes the tracker, which is easy to do by
     // accident (`--issues` reads like "the issue surface", and the tracker IS
     // issues). Say so rather than letting the mirror go quietly stale.
-    if (!syncIntegrations && !options.status && !this.ctx.dryRun) {
+    //
+    // This must survive the default invocation: `info()` is verbose-only, so
+    // reporting it there would leave the ordinary run exactly as silent as
+    // before. `data()` gives the structured form under --json and the
+    // default-visible `notice()` otherwise. Dry runs report it too — "what
+    // would this do?" has to include "not the tracker".
+    if (!syncIntegrations && !options.status) {
       const config = await readConfig(tbdRoot);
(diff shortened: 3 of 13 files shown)


jlevy:
## Code Review — PR #227

### Scope

Reviewed PR #227 at `6529778070855fffd6b695af5aae4b48806307ef` against base
`b036fbe2f0ececcdf1d74134b284e59f19c97ea2`: 10 files, +317/−28. I used the
repository's `review-github-pr` / `review-code` shortcuts, inspected the full diff and
the surrounding CLI output, bridge-state, sync-engine, and git-push paths, and checked
the linked Phase 1 plan in PR #222. I also checked prior comments/reviews; there are no
earlier human or inline findings to carry forward.

Validation at review time: `git diff --check` passes. All substantive GitHub checks
pass (Benchmark, Coverage & Lint, DeepSource Secrets, and the Node test matrix on
Ubuntu, macOS, and Windows). Cursor Bugbot skipped because its usage limit was reached.

### Summary and Verdict

The quiet-write guard is a sound fix for the current `LinkRecord` schema: it preserves
`synced_at` as the merge tiebreaker while preventing timestamp-only rewrites, and the
new behavior test checks byte-for-byte stability after the mirror settles. Applying
`--no-verify` at both hidden sync-branch push sites also matches the linked design and
is covered behaviorally for the normal retry path.

One user-visible behavior remains broken, however: the new tracker-skip message is
routed through a verbose-only output method. The ordinary command still excludes the
tracker silently, which is the exact defect this portion of the PR is intended to fix.

**Verdict: changes requested.** Findings: 0 Blocker, 0 High, 1 Medium, 0 Low.

### Findings

#### R1 — Medium — Make the tracker-skip notice visible in ordinary runs

**Files:**
[`packages/tbd/src/cli/commands/sync.ts:148`](https://github.com/jlevy/tbd/blob/6529778070855fffd6b695af5aae4b48806307ef/packages/tbd/src/cli/commands/sync.ts#L148-L155),
[`packages/tbd/src/cli/lib/output.ts:410`](https://github.com/jlevy/tbd/blob/6529778070855fffd6b695af5aae4b48806307ef/packages/tbd/src/cli/lib/output.ts#L410-L416)

The new branch calls `this.output.info(...)`, but `OutputManager.info()` emits only
when `--verbose` or `--debug` is active. It is suppressed in the default invocation
and in JSON mode. The guard also explicitly skips the message whenever `--dry-run` is
active. Consequently, ordinary `tbd sync --issues` (and `--docs`) still gives no
indication that enabled external trackers were excluded, and the dry-run form remains
silent as well. A user can therefore finish a seemingly successful scoped sync while
the tracker quietly goes stale.

**Fix:** Use a default-visible output path for this noteworthy event (for example,
`notice()` for normal text output), represent the omission in machine-readable output,
and do not suppress it merely because the run is dry. Add the Phase 1 plan's missing
surface-honesty golden coverage for the default, dry-run, and JSON forms, including
negative cases where integrations are inert or explicitly selected.

### Design Assessment

The persistence design is appropriately narrow: compare the prior parsed link record,
ignore only `synced_at`, and retain all other current correctness fields. This is
preferable to dropping `synced_at`, because `pickNewestLinkRecord` uses it when provider
timestamps tie. The hidden-branch `--no-verify` change is also intentionally scoped to
tbd's bookkeeping ref; it does not alter ordinary source-branch pushes or server-side
protections.

One non-blocking hardening opportunity: `recordSubstantivelyDiffers()` manually lists
the current schema fields. If `LinkRecordSchema` grows, TypeScript will not force this
comparator to grow with it, so a future field could be silently treated as
non-substantive. Prefer an exhaustive comparison of the whole parsed record after
normalizing/removing only `synced_at`, or a typed key map that fails compilation when
the schema changes.

### Documentation

The documentation changed in this PR accurately explains the open-work denominator,
inherited `spec_path` expansion, and `max_nesting` skips. No additional in-branch
architecture update appears necessary.

The linked plan in PR #222 has one stale cross-PR sentence in its “API Changes” section
that still says `LinkRecordSchema` drops `synced_at`, while its design and schema-table
sections correctly say the field is retained. That should be reconciled on PR #222 so
the plan has one answer before it merges.

### Suggestions

- Make the bridge-record comparator schema-exhaustive, as described above.
- Date the exact repository selection counts in `tbd-docs.md` (or present them as a
  dated measurement) so the example does not read as perpetually current while bead
  totals continue to change.

### False Positives / Do Not Fix

- Do not remove `synced_at`; retaining it for deterministic multi-machine merge
  tie-breaking is correct.
- Do not re-enable parent-repository pre-push hooks for the hidden `tbd-sync` ref merely
  to satisfy the normal source-branch gate. The scoped bypass is the intended behavior
  under the linked Phase 1 design.
- The direct `writeLinkRecordIfChanged` unit tests complement rather than replace the
  end-to-end settled-mirror byte-identity test; both levels are useful here.

### CI Status

Passing: Benchmark; Coverage & Lint; DeepSource Secrets; tests on Ubuntu Node 22.12 and
Node 24, macOS Node 24, and Windows Node 24. Cursor Bugbot: skipped due to usage limit.

jlevy:
Addressed the review findings in `e263b76`:

- **R1: fixed** — you were right, and the fix as shipped was a no-op for ordinary runs. The notice went through `OutputManager.info()`, which emits only under `--verbose`/`--debug` and is suppressed by `--json`, and the guard also skipped it on dry runs. It now routes through `data()`: `{"skippedSurfaces":["integrations"]}` under `--json`, and the default-visible `notice()` otherwise. The dry-run suppression is gone — "what would this do?" has to include "not the tracker."
- **Suggestion (comparator): fixed** — `recordSubstantivelyDiffers()` is now driven by `Readonly<Record<Exclude<keyof LinkRecord, 'synced_at'>, true>>` plus `isDeepStrictEqual`, so a new schema field fails compilation until someone decides whether it is substantive. Verified by temporarily adding a probe field to `LinkRecordSchema` and confirming `tsc` rejects it.
- **Suggestion (dated counts): fixed** — the `tbd-docs.md` figures are now presented as a dated 2026-08-14 measurement, noting both numbers move as a repository grows, with a pointer to `tbd --dry-run integration sync --push` for the reader's own count.
- **Doc note on #222: fixed** — the plan's API Changes line said `LinkRecordSchema` drops `synced_at` while its design and schema table said it is retained. Reconciled on #222 in `37bf96f`; the plan now gives one answer and names the nested-schema passthrough work that does belong under Schema.

New coverage: `tests/cli-sync-surface-honesty.tryscript.md`, 11 assertions over the default, dry-run, and JSON forms plus the negatives you named — no integration configured, tracker explicitly selected via `--integrations`, a full `tbd sync`, `--status`, and a team that set `sync_on_tbd_sync: false`.

Two notes on the "do not fix" list, both of which I agree with and want to record explicitly:

- `synced_at` stays. Finding that `pickNewestLinkRecord` uses it as a tiebreaker is what changed the approach mid-implementation — the plan had preferred dropping it, and that recommendation is now struck through in #222 with the reason.
- The `--no-verify` scope stays limited to the hidden `tbd-sync` ref.

CI is green on `e263b76` across the Node matrix, Coverage & Lint, Benchmark, and DeepSource (Grade A; `bridge-state.ts` at 100% statements/functions). Bugbot is still skipping on the usage limit.

---

jlevy:
## Final senior review — approved for merge

Reviewed the complete diff at `7fea4ce966042549e29856d3683f6bff9422cee5` against current `main`, including the bridge-state persistence semantics, both hidden sync-branch push paths, scoped-sync output behavior, documentation, and behavioral tests.

The final pass found one additional Medium issue: the revised JSON notice emitted a second top-level document on stdout. That violated the established single-result-document contract even though the grep-based test passed. Commit `7fea4ce9` fixes it by keeping stdout as one parseable result and emitting the structured skip diagnostic on stderr; the regression now parses both channels.

Final result: no remaining Blocker, High, Medium, or Low findings. No unresolved review threads.

Validation on the exact head:
- local format, typecheck, lint, build, and full suite: 137 test files / 2,027 tests passed
- strengthened surface-honesty transcript: 13/13 passed
- source-branch pre-push gate passed in full
- hosted CI passed on Ubuntu Node 22 and 24, macOS Node 24, and Windows Node 24
- Coverage & Lint, package audit, publint, Benchmark, and DeepSource Secrets all passed

Verdict: ready to merge and appropriate to include. This fixes recurring sync cost without weakening ordinary source-branch checks or server-side protections.

PR #232, opened 2026-08-15
fix: Resolve duplicate short IDs in ids.yml to prevent bead data loss

jlevy:
## Summary

Fixes a data-loss bug where `merge=union` of `ids.yml` across two clones that independently allocated the same short ID for different beads caused one bead to become unrenderable (`formatDisplayId` throws) and the other to silently address the wrong bead.

Tracking bead: tbd-ysz0

## Problem

`ids.yml` uses `merge=union` in `.gitattributes`, so when two clones both allocate short ID `00wl` for different ULIDs while offline, the merged file contains:

```yaml
00wl: 01aaaaaaaaaaaaaaaaaaaaaaaa
00wl: 01bbbbbbbbbbbbbbbbbbbbbbbb
```

`parseYamlToleratingDuplicateKeys` resolves this as last-occurrence-wins, silently dropping ULID A from both maps. The consequences:

1. **Bead A becomes unrenderable.** `formatDisplayId` throws: "No short ID mapping found for internal ID" — any command that formats that bead (including `tbd list`) fails outright.
2. **`tbd-00wl` silently addresses the wrong bead** (B instead of A).

Neither the `generateUniqueShortId` retry loop (checks only local disk under local lock), the `saveIdMapping` append-only size check (entry count does not shrink), nor the "auto-fixed on next save" warning (false — the losing ULID is dropped at parse time) caught or corrected this.

## Fix

**Repair at load time** in `loadIdMapping` (and `parseIdMappingFromYaml`, `loadIdMappingRaw`), so no code path ever sees a mapping with missing ULIDs:

1. **Check merge conflict markers first.** All three read paths now call `hasMergeConflictMarkers` before any other parsing — including before the duplicate-key branch. This ensures content with both conflict markers and duplicate keys always throws `MergeConflictError` rather than being silently processed as live data.
2. **Parse via YAML AST.** A new `parseYamlDocumentEntries` function uses the yaml library's `parseDocument` with `uniqueKeys: false`, which retains every key-value pair in the AST — including duplicates. This replaces the former regex-based `parseAllIdEntries`, which could not match quoted YAML keys (`"1234"`, `"0000"`, `"true"`, `"null"`) that `stringifyYaml` produces for numeric-looking short IDs. Roughly 15.8% of live `ids.yml` entries have quoted keys — the regex silently skipped them all.
3. **Validate every entry** against `ShortId` and `Ulid` Zod schemas in `extractValidatedIdEntries`. Invalid entries cause a loud error rather than being silently skipped.
4. **Resolve deterministically** via `resolveDuplicateShortIds`:
   - For each contested short ID, the **lexicographically smallest ULID keeps it** (ULIDs are time-ordered, so the earlier-created bead wins).
   - Each displaced ULID receives a replacement short ID **derived from its own characters** via `deriveShortIdFromUlid` — successive windows extracted from the ULID string (length determined by `calculateOptimalLength`), no randomness.
   - Duplicate keys are processed in sorted order; displaced ULIDs are processed in sorted order — full determinism at every step.
5. **On next save**, the corrected mapping is written (the normal `saveIdMapping` read-merge-write path picks up the repaired in-memory mapping), fulfilling the "auto-fixed on next save" promise that was previously false.

### Why deterministic derivation, not random re-allocation

If the displaced ULID's replacement short ID were generated via `generateShortId` (random), two clones repairing the same merged file would pick different replacements. The next union merge would then produce two different short IDs pointing at the same ULID — trading a hard failure for slow divergence. Deriving the replacement from the ULID's own characters makes the repair a pure function of the file content, so every clone converges to the same result.

Window candidates are taken from the **tail of the ULID first**, so they come from its 80 random bits rather than the shared 48-bit timestamp prefix. Two beads created in the same millisecond share their leading characters, so a front-first scan would collide constantly.

### Known residual: alias creation across clones at different sync states

Two clones at different sync states have different `usedShortIds` sets when deriving replacements for displaced ULIDs. This means they can derive different replacement short IDs for the same displaced ULID, creating **aliases** — two short IDs that both point to the same ULID. This is acceptable: aliases are a cosmetic inconsistency (both short IDs work), not data loss, and the next `saveIdMapping` on either clone writes a clean file. The fundamental guarantee — every ULID is addressable on every clone — holds.

### Design decisions

- **Repair at load, not only at save or in `tbd doctor --fix`**: The crash happens at load time when any command formats the displaced bead. Repairing at load prevents the crash on every code path without requiring the user to run a separate repair command first. The file is still corrected on next save.
- **Warning text updated**: The warning now accurately says "Displaced entries have been assigned deterministic replacement short IDs" instead of the previously-false "The file will be auto-fixed on next save" (the auto-fix promise is now true, but the warning also explains what happened to the displaced entries).
- **`tbd doctor` not modified**: This is a load-time data-integrity repair, not a health check. Adding a doctor check for this condition would be reasonable future work but is out of scope for this fix.

## Tests added

New test file: `packages/tbd/tests/id-mapping-duplicate-short-ids.test.ts` (44 tests)

- **`parseYamlDocumentEntries` unit tests** (10 tests): quoted keys, special YAML scalars (`true`, `null`, `0000`), trailing comments, empty content, comments-only, duplicates with quoted keys, duplicates with trailing comments, keys with dots/dashes/underscores, malformed YAML throws
- **Quoted-key bystander tests** (3 tests): uses `stringifyYaml` to produce a file with numeric short ID alongside a duplicate — the scenario that broke the former regex parser — verifying quoted bystander entries survive repair intact
- **Exact bug report scenario**: both ULIDs stay addressable, neither `formatDisplayId` call throws
- **Tiebreak rule pinned**: lexicographically smallest ULID keeps the contested short ID regardless of file order
- **Convergence tests**: two independent repairs of the same input produce identical output, including with multiple duplicate keys and different file orderings
- **End-to-end union merge simulation**: write merged file, load (repair), save, reload — verifies the file is clean after save and produces no warnings on reload
- **Two-clone convergence**: two separate temp dirs with the same merged content produce identical mappings
- **`parseIdMappingFromYaml`** with conflicting duplicates: verifies the same fix applies to the conflict-resolution parse path
- **Conflict markers take priority**: 4 tests ensuring `MergeConflictError` is thrown when content has both conflict markers and duplicate keys (in both `loadIdMapping` and `parseIdMappingFromYaml`), and when content has conflict markers without duplicate keys
- **Repair path validation** (R5): invalid short ID causes loud error; invalid ULID causes loud error naming both the offending value and its key
- **Trailing comment regression** (R3): duplicate lines with trailing comments are parsed and repaired correctly
- **`deriveShortIdFromUlid` escalation tiers under collision pressure**: 5 tests that saturate `usedShortIds` to drive each fallback tier — the single-offset window loop, the 5-char fallback, and the terminal throw — pinning the exact returned value at each tier, asserting determinism and `ShortId` pattern conformance at every tier
- **Unit tests** for `deriveShortIdFromUlid` and `resolveDuplicateShortIds`

Updated existing test in `ids.test.ts`: the "handles duplicate keys mapping to different ULIDs" test previously asserted the old broken behavior (size 2, last-wins). Updated to assert the correct behavior (size 3, all ULIDs preserved, smallest ULID wins).

## Validation

`pnpm precommit` (format, lint, build, full test suite): **138 test files, 2076 tests passed, 0 failures.**

## Scope

Production changes in `id-mapping.ts` and `yaml-utils.ts`. Doc update in `tbd-design.md`. No changes to the on-disk format, sync engine, integrations, or ID system structure. No new dependencies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GF8v1VycQoBxB1s2C2LQri)_

### DIFF
--- a/packages/tbd/docs/tbd-design.md
+++ b/packages/tbd/docs/tbd-design.md
@@ -674,7 +674,11 @@ serialization:
 > file” as the correct state.
 > The `merge=union` strategy keeps all lines from both sides, which is safe for
 > `ids.yml` since it’s an append-only mapping.
-> Duplicate keys (if any) are tolerated by the YAML parser and auto-fixed on next save.
+> Contested duplicate keys (same short ID mapped to different ULIDs) are resolved
+> deterministically at load time: the lexicographically smallest ULID keeps the
+> contested short ID, and each displaced ULID receives a replacement derived from its
+> own characters (no randomness).
+> The corrected mapping is written on next save.
 > This file is placed inside `.tbd/` so all tbd settings are self-contained in one
 > directory. Git supports `.gitattributes` in subdirectories with paths relative to that
 > directory.
--- a/packages/tbd/src/file/id-mapping.ts
+++ b/packages/tbd/src/file/id-mapping.ts
@@ -12,8 +12,7 @@ import { join, dirname } from 'node:path';
 import { writeFile } from 'atomically';
 
 import {
-  parseYamlToleratingDuplicateKeys,
-  detectDuplicateYamlKeys,
+  parseYamlDocumentEntries,
   MergeConflictError,
   stringifyYaml,
   hasMergeConflictMarkers,
@@ -30,7 +29,7 @@ import {
   type InternalIssueId,
 } from '../lib/ids.js';
 import { naturalSort } from '../lib/sort.js';
-import { IdMappingYamlSchema } from '../lib/schemas.js';
+import { ShortId, Ulid } from '../lib/schemas.js';
 
 /**
  * ID mapping from short ID to ULID.
@@ -60,25 +59,50 @@ function serializeIdMapping(mapping: IdMapping): string {
 }
 
 /**
- * Parse all key-value entries from raw ids.yml content, preserving every
- * occurrence of duplicate keys. The YAML parser's `uniqueKeys: false` mode
- * silently drops all but the last value for each key; this line-level parser
- * captures every entry so that duplicate-key repair can see all competing ULIDs.
+ * Extract validated ID mapping entries from raw YAML content, preserving every
+ * occurrence of duplicate keys.
  *
- * Only matches top-level `shortId: ulid` lines; comments and blank lines are
- * skipped. This is safe because ids.yml is a flat key-value map.
+ * Uses the `yaml` library's AST (via `parseYamlDocumentEntries`) to handle
+ * quoted keys, trailing comments, and all other YAML formatting correctly —
+ * unlike a line-level regex, which cannot match keys that `stringifyYaml`
+ * quotes (e.g., all-digit short IDs like `"1234"`, `"0000"`, or YAML
+ * special scalars like `"true"`, `"null"`).
+ *
+ * Each extracted entry is validated against `ShortId` and `Ulid` schemas.
+ * Entries that fail validation cause a loud error rather than being silently
+ * skipped, preserving the same failure behavior as the Zod-validated parse path.
+ *
+ * @param content - Raw ids.yml YAML content
+ * @param filePath - Optional file path for error messages
+ * @returns Every (shortId, ulid) pair in document order, plus duplicate keys
  */
-export function parseAllIdEntries(content: string): { shortId: string; ulid: string }[] {
-  const entries: { shortId: string; ulid: string }[] = [];
-  for (const line of content.split('\n')) {
-    // Match lines like "a7k2: 01hx5zzkbkactav9wevgemmvrz"
-    // ShortId allows [0-9a-z._-]+, Ulid is exactly 26 [0-9a-z] chars
-    const match = /^([0-9a-z._-]+):\s+([0-9a-z]{26})\s*$/.exec(line);
-    if (match) {
-      entries.push({ shortId: match[1]!, ulid: match[2]! });
+function extractValidatedIdEntries(
+  content: string,
+  filePath?: string,
+): {
+  allEntries: { shortId: string; ulid: string }[];
+  duplicateKeys: string[];
+} {
+  const { entries, duplicateKeys } = parseYamlDocumentEntries(content, filePath);
+
+  const allEntries: { shortId: string; ulid: string }[] = [];
+  for (const { key, value } of entries) {
+    const shortIdResult = ShortId.safeParse(key);
+    if (!shortIdResult.success) {
+      const location = filePath ? ` in ${filePath}` : '';
+      throw new Error(`Invalid short ID "${key}"${location}: ${shortIdResult.error.message}`);
     }
+    const ulidResult = Ulid.safeParse(value);
+    if (!ulidResult.success) {
+      const location = filePath ? ` in ${filePath}` : '';
+      throw new Error(
+        `Invalid ULID "${value}" for key "${key}"${location}: ${ulidResult.error.message}`,
+      );
+    }
+    allEntries.push({ shortId: key, ulid: value });
   }
-  return entries;
+
+  return { allEntries, duplicateKeys };
 }
 
 /**
@@ -185,7 +209,14 @@ export function resolveDuplicateShortIds(
   for (const shortId of sortedDuplicateKeys) {
     const ulids = duplicateUlids.get(shortId)!;
 
-    if (ulids.size <= 1) {
+    // Guard: a reported duplicate key with no matching parsed entries is a no-op.
+    // This should not happen when detection and extraction use the same parser,
+    // but the guard prevents writing undefined into the Map if it ever does.
+    if (ulids.size === 0) {
+      continue;
+    }
+
+    if (ulids.size === 1) {
       // Same ULID repeated (harmless duplicate); just keep one copy
       const ulid = ulids.values().next().value!;
       shortToUlid.set(shortId, ulid);
@@ -212,7 +243,11 @@ export function resolveDuplicateShortIds(
   displacedUlids.sort((a, b) => (a.ulid < b.ulid ? -1 : a.ulid > b.ulid ? 1 : 0));
 
   for (const { ulid } of displacedUlids) {
-    const newShortId = deriveShortIdFromUlid(ulid, usedShortIds);
+    const newShortId = deriveShortIdFromUlid(
+      ulid,
+      usedShortIds,
+      calculateOptimalLength(usedShortIds.size),
+    );
     shortToUlid.set(newShortId, ulid);
     ulidToShort.set(ulid, newShortId);
     usedShortIds.add(newShortId);
@@ -250,8 +285,8 @@ export async function loadIdMapping(baseDir: string): Promise<IdMapping> {
   }
 
   // Check for unresolved merge conflict markers before any other parsing.
-  // This must run before the duplicate-key branch, because parseAllIdEntries
-  // silently skips conflict marker lines and would treat both sides of an
+  // Conflict markers must be caught before the duplicate-key branch, because
+  // the YAML AST parser skips marker lines and would treat both sides of an
   // unresolved conflict as live data.
   if (hasMergeConflictMarkers(content)) {
     throw new MergeConflictError(
@@ -262,13 +297,13 @@ export async function loadIdMapping(baseDir: string): Promise<IdMapping> {
     );
   }
 
-  // Detect duplicate keys from the raw content before YAML parsing.
-  const duplicateKeys = detectDuplicateYamlKeys(content);
+  // Single AST parse: extracts every (key, value) pair in document order,
+  // handles quoted keys and trailing comments, and detects duplicates — all
+  // from one parseDocument call rather than separate regex + YAML passes.
+  const { allEntries, duplicateKeys } = extractValidatedIdEntries(content, filePath);
 
   if (duplicateKeys.length > 0) {
-    // Parse all entries from raw content (preserving every duplicate) and
-    // resolve deterministically so no ULID loses its mapping.
-    const allEntries = parseAllIdEntries(content);
+    // Resolve deterministically so no ULID loses its mapping.
     const mapping = resolveDuplicateShortIds(allEntries, duplicateKeys);
 
     console.warn(
@@ -281,21 +316,11 @@ export async function loadIdMapping(baseDir: string): Promise<IdMapping> {
     return mapping;
   }
 
-  // No duplicates: standard YAML parse path
-  const { data: rawData } = parseYamlToleratingDuplicateKeys(content, filePath);
-  const data = rawData ?? {};
-
-  // Validate with Zod schema - ensures all keys are valid short IDs and values are ULIDs
-  const parseResult = IdMappingYamlSchema.safeParse(data);
-  if (!parseResult.success) {
-    throw new Error(`Invalid ID mapping format in ${filePath}: ${parseResult.error.message}`);
-  }
-  const validData = parseResult.data;
-
+  // No duplicates: build mapping directly from validated entries.
   const shortToUlid = new Map<string, string>();
   const ulidToShort = new Map<string, string>();
 
-  for (const [shortId, ulid] of Object.entries(validData)) {
+  for (const { shortId, ulid } of allEntries) {
     shortToUlid.set(shortId, ulid);
     ulidToShort.set(ulid, shortId);
   }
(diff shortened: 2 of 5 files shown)


jlevy (commented):
## Senior engineering review — PR #232

**Scope:** Full review of `d55790f..dc9801b` (3 commits, 3 files, +914/−13): `packages/tbd/src/file/id-mapping.ts` (+222/−9), `tests/id-mapping-duplicate-short-ids.test.ts` (new, 681 lines), `tests/ids.test.ts`. Reviewed against `general-eng-agent-principles`, `general-coding-rules`, `general-comment-rules`, `error-handling-rules`, `general-testing-rules`, `typescript-rules`, plus `tbd-design.md` §2.5/§merge=union. Claims below were verified by executing the code on this branch, not by reading alone.

**Verdict: request changes.** The diagnosis is correct, the tiebreak/derivation design is sound, and the conflict-marker ordering fix in `735e9ee` is right. But the repair is built on a hand-rolled regex line parser that disagrees with the duplicate detector it is paired with, and that disagreement **silently drops valid, unrelated entries — reintroducing the exact `formatDisplayId` crash this PR fixes**, on roughly 1 in 121 beads. R1–R3 are one root cause with one fix, and that fix makes the change smaller than it is now.

---

### Findings

**R1 — Blocker — `parseAllIdEntries` silently drops every entry whose key YAML quotes**
`packages/tbd/src/file/id-mapping.ts:71-82`

tbd's own serializer quotes short IDs that look like YAML scalars. Measured against `stringifyYaml`, these round-trip as quoted: `1234`, `0123`, `9999`, `0000`, `true`, `null`, `-123`, `.inf`, `0x1f`, `12.5`, `1e10`, `0o07`. The regex `^([0-9a-z._-]+):\s+([0-9a-z]{26})\s*$` cannot match a quoted key, so those lines are skipped entirely.

Reachability, measured over 200,000 `generateShortId(4)` outputs: **1,649 (0.825%) serialize in a form `parseAllIdEntries` cannot read back — about 1 in 121 beads.**

The drop is not limited to the contested key. Once *any* duplicate exists anywhere in the file, the repair branch runs and every quoted entry in the whole file is discarded. Verified end-to-end:

```yaml
"true": 01cccccccccccccccccccccccc     # unrelated bead
a1b2: 01aaaaaaaaaaaaaaaaaaaaaaaa       # the union-merge duplicate
a1b2: 01bbbbbbbbbbbbbbbbbbbbbbbb
```

`loadIdMapping` returns `[["a1b2","01aaaa…"],["bbbb","01bbbb…"]]` — `01cccc…` is gone, so `formatDisplayId('is-01cccc…')` throws and `tbd list` fails. That is the reported bug, reintroduced by its own fix, on a bead that had nothing to do with the merge.

One nuance worth stating precisely: `saveIdMapping` accidentally rescues the dropped entry, because its internal `loadIdMappingRaw` uses the YAML path and `mergeIdMappings` re-adds what the repair lost. So the file self-heals *if a write happens first*. Read-only commands (`list`, `show`, `ready`) never write, so the crash stands for them.

**Fix:** stop re-parsing with a regex — see R2.

---

**R2 — High — the duplicate pairs are already parsed; the regex re-derives what was just thrown away**
`packages/tbd/src/file/id-mapping.ts:71-82`, `src/utils/yaml-utils.ts:210`

`parseYamlToleratingDuplicateKeys` already calls `parseDocument(content, { uniqueKeys: false })`. That AST retains **every** duplicate pair; the function then discards them by calling `doc.toJSON()`. Verified:

```
AST pair: "00wl" -> "01aaaaaaaaaaaaaaaaaaaaaaaa"
AST pair: "00wl" -> "01bbbbbbbbbbbbbbbbbbbbbbbb"
toJSON():  {"00wl":"01bbbbbbbbbbbbbbbbbbbbbbbb"}
```

So the PR re-derives with a regex, one call later, information the YAML parser already had in hand. The cost is that three parsers with three different notions of "a line" must now agree forever:

| | pattern | sees `"1234": …` as |
|---|---|---|
| `detectDuplicateYamlKeys` | `/^([^\s#][^:]*?):\s/` | key `"1234"` *(with quotes)* |
| `parseAllIdEntries` | `/^([0-9a-z._-]+):\s+([0-9a-z]{26})\s*$/` | no match at all |
| `parseDocument` + Zod | full YAML | key `1234` |

Three answers, three spellings. R1, R3, R4 and R5 are all downstream of that.

**Fix:** add one primitive to `yaml-utils.ts` that returns what the existing parse already computed — every pair plus the duplicate-key set — and have `id-mapping.ts` use it for both detection and extraction. Validated against all six shapes in this review (quoted keys, all-digit keys, trailing comments, ordinary files, empty, comments-only); it returns correct key spellings and preserves both sides of every duplicate. `parseAllIdEntries` and its regex then delete entirely, and `detectDuplicateYamlKeys`'s separate scan drops out of this path.

---

**R3 — High — a trailing comment on a duplicated line defeats the repair, silently**
`packages/tbd/src/file/id-mapping.ts:76`

```yaml
a1b2: 01aaaaaaaaaaaaaaaaaaaaaaaa # from clone A
a1b2: 01bbbbbbbbbbbbbbbbbbbbbbbb
```

`detectDuplicateYamlKeys` reports `a1b2`, so the repair branch runs — but `parseAllIdEntries` returns only the second entry, so `01aaaa…` is never re-homed. The warning then tells the user "Displaced entries have been assigned deterministic replacement short IDs," which is false: nothing was displaced, the entry was dropped. This is the failure mode of a hand-resolved conflict, which is exactly the situation this code exists to handle.

**Fix:** same as R2.

---

**R4 — Medium — `undefined` written into a `Map<string, string>`**
`packages/tbd/src/file/id-mapping.ts:188-194`

When a reported duplicate key has no matching parsed entries (reachable today via R1), `ulids` is empty and `ulids.values().next().value!` is `undefined`. The non-null assertion is load-bearing and wrong. Verified output:

```
shortToUlid: [["a1b2","01aaaa…"],["\"true\"",null]]
ulidToShort: [["01aaaa…","a1b2"],[null,"\"true\""]]
```

`stringifyYaml` happens to drop undefined-valued keys, so it does not corrupt the file today — but the type contract is already violated and the next consumer of that map will not be so lucky.

**Fix:** `if (ulids.size === 0) continue;` before the `size <= 1` branch. The `!` should not survive either way. Becomes unreachable once R2 lands, but the guard is still correct.

---

**R5 — Medium — the "schema validation equivalence" claim in the PR description does not hold**
`packages/tbd/src/file/id-mapping.ts:268-282`

The description argues the regex "enforces the same constraints as the Zod schemas, so validation coverage is equivalent." The constraints may match; the *behavior on violation* does not. Zod **throws** `Invalid ID mapping format`; the regex **silently skips**. Verified with an uppercase ULID (`e5f6: 01CCCC…`) alongside a duplicate: pre-PR that file raised a clear error, post-PR the entry vanishes without a word. Trading a loud failure for a silent one is the wrong direction in a data-integrity fix.

**Fix:** validate the extracted entries with `ShortId`/`Ulid` on the duplicate path too, and fail loudly on a non-conforming line rather than skipping it.

---

**R6 — Low — invalid 27-character ULID in a test fixture**
`packages/tbd/tests/id-mapping-duplicate-short-ids.test.ts:317`

`'01zzzzzzzzzzzzzzzzzzzzzzzzz'` is 27 chars; ULIDs are 26. The test passes only because it calls `resolveDuplicateShortIds` directly, bypassing length validation. The tiebreak assertion it makes is still valid, but the fixture would not survive the stricter validation R5 asks for.

**Fix:** drop one `z`.

---

**R7 — Low — `deriveShortIdFromUlid` hardcodes length 4, ignoring `calculateOptimalLength`**
`packages/tbd/src/file/id-mapping.ts:93`, `111`

`length = 4` and `longerLength = length + 1` are bare literals, against `general-coding-rules` ("NEVER hardcode numeric values directly in code"). More concretely: `calculateOptimalLength()` lives in this same file and switches to 5 chars past 50K issues, but a repo over that threshold still derives 4-char replacements. Not a correctness bug — uniqueness is checked — but the two ID-length policies should not disagree.

**Fix:** derive the default from `calculateOptimalLength(mapping.shortToUlid.size)`, or name the constants and comment why the repair path is deliberately fixed at 4.

---

- **S1 — the structural fix is to invert the file.** Keying `ids.yml` by ULID (`<ulid>: <shortid>`) makes `merge=union` collision-*proof*: ULIDs are globally unique by construction, so a union merge can only ever produce the same key twice with the same value — harmless, and no bead is ever unrenderable. The entire tiebreak/derivation/convergence apparatus in this PR exists to compensate for a primary key that two clones can independently mint. Out of scope here (on-disk format change plus migration), but worth a bead: it deletes this problem class rather than repairing it.
- **S2 — `loadIdMappingRaw` (line 379)** still uses the last-wins parse. The PR notes this and correctly argues it is masked by `mergeIdMappings` precedence. Once R2's primitive exists this is a one-line change; making all three read paths agree is worth more than the argument for why one of them can stay wrong.
- **S3 — `tbd doctor` check.** Agreed it is out of scope. Worth a bead so it is not lost.

### False positives / verified benign — do not "fix" these

- **The conflict-marker ordering fix (`735e9ee`) is correct and necessary.** Without it, `parseAllIdEntries` would treat both sides of an unresolved conflict as live data. Keep it.
- **`resolveIdMappingConflicts` is not broken by the new marker check.** Verified: a standard three-way conflict still strips and merges to all three entries.
- **The tiebreak rule and tail-first window scan are sound.** Smallest-ULID-wins is stable and order-independent; scanning from the tail to draw on the ULID's 80 random bits rather than the shared 48-bit timestamp prefix is the right call, and the reasoning in the PR description is correct.
- **Determinism within a single file's repair holds**, and the known alias residual across clones at different sync states is correctly characterised as cosmetic rather than data loss.
- **diff3-style conflicts (`|||||||`) raise a confusing YAML error instead of `MergeConflictError`** — I verified this is **pre-existing on `main`**, not introduced here. Adjacent, worth a bead, not this PR's problem.

### Design assessment

The load-time repair point is the right choice and well argued: the crash happens when a command formats a displaced bead, so repairing at save-time or behind `tbd doctor --fix` would leave `tbd list` broken until the user ran something else. Deterministic derivation over random re-allocation is also right, and the PR's reasoning about slow divergence under random re-allocation is correct.

The weakness is not the algorithm, it is the input layer. The repair operates on a *second, independent, weaker* reading of the file than the one the rest of the module uses, and the two readings disagree on a measurable 0.8% of real short IDs. A data-integrity repair must not have a narrower view of the data than the code it is protecting. Consolidating on the AST the parser already builds removes ~40 lines, deletes the regex, fixes R1/R3/R4/R5 together, and restores schema validation on the repair path — a smaller diff than the current one, with strictly more coverage.

Test quality is otherwise high: the escalation-tier tests pin exact values at each fallback rather than asserting "something valid," convergence is tested across file orderings and two temp dirs, and the end-to-end merge→load→save→reload cycle is the right shape. What is missing is a test whose *input* is a file tbd itself wrote containing an at-risk short ID — every fixture uses hand-picked keys like `a1b2`/`00wl` that the regex happens to handle, which is why the gap survived 32 tests and 94.89% line coverage. Coverage measured the repair code; nothing measured whether the repair could *see* the file.

### Documentation

`packages/tbd/docs/tbd-design.md:675` still states: *"Duplicate keys (if any) are tolerated by the YAML parser and auto-fixed on next save."* That is the same false claim this PR corrects in the warning text, and it never mentions that duplicates can be *contested* (same short ID, different ULIDs) — the entire bug. Update it to describe the resolution semantics: contested keys are resolved at load time, smallest ULID keeps the short ID, displaced ULIDs get deterministic derived replacements.

### CI status

Green at `dc9801b` — Test (ubuntu Node 22.12.0 / 24, macos Node 24, windows Node 24), Benchmark, Coverage & Lint all `success`. Cursor Bugbot `neutral` (usage limit reached, no analysis performed). DeepSource: grade A, static analysis only, no AI review run. **No human review has been posted on this PR** — the local suite also passes (126 tests across the two touched files), so CI being green is not evidence against R1–R5; none of the fixtures exercise them.

---

jlevy:
## Review addressed — `7b5f47e` + `0024a4f`

Every finding is resolved. Verified by execution against this branch, not by reading the diff.

### R1 (blocker) — fixed, and the exposure was worse than measured

The 0.825%-of-random-short-IDs figure understates it badly for this repo. **274 of the 1,739 live entries in `ids.yml` are quoted (15.8%)**, because migrated beads carry numeric short IDs (`"1"`, `"100"`, `"1000"`, `"0622"`). So a single union-merge duplicate would have dropped 274 unrelated beads here, not ~14.

Loading the real file with an injected duplicate, before vs. after:

```
real file: 1743 entries, 274 quoted
loaded:    1744 short IDs, 1744 ULIDs
beads that fail to render: 0
quoted bystander 01krvqqagj5ma44x1c8anjb2z5 renders as: tbd-0622
injected ULID renders as: tbd-zzzz
```

### Disposition

| | Status |
| --- | --- |
| R1 quoted keys dropped | Fixed via R2 |
| R2 consolidate on the AST | `parseYamlDocumentEntries` added; `parseAllIdEntries` and its regex deleted |
| R3 trailing comment | Fixed via R2; regression test added |
| R4 `undefined` into `Map` | `if (ulids.size === 0) continue;` added |
| R5 silent skip | Validation on the repair path; throws now, both branches tested |
| R6 27-char ULID fixture | Fixed |
| R7 hardcoded length 4 | Now `calculateOptimalLength(usedShortIds.size)` |
| S2 `loadIdMappingRaw` | Uses the same path; all three read paths agree |
| Docs | `tbd-design.md` updated to describe contested-key resolution |

Keys now resolve through `String(item.key.value)`, so the YAML parser decides what a key is — which is why `"1234"` works and why a regex could not.

The conflict-marker ordering from `735e9ee` survived the parser swap; re-verified that conflicted-plus-duplicated input still raises `MergeConflictError` at all three read paths. The "schema validation equivalence" claim is removed from the description, since the constraints matched but the behavior on violation did not.

### Two corrections to my own earlier notes

- I attributed the uncovered `yaml-utils.ts:277-278` to the new primitive. Wrong — those lines are inside `parseYamlToleratingDuplicateKeys` (declared at :246), which predates this PR. `parseYamlDocumentEntries` (198-245) is now fully covered.
- Which surfaces something worth deciding: **`parseYamlToleratingDuplicateKeys` now has no production callers**, only its own declaration and tests. This PR removed its last consumer. It is the parser with last-occurrence-wins semantics — the behavior that caused this bug — so leaving it exported means a future caller can still reach for the wrong one. Filed as `tbd-xt7r` rather than folded in here, to keep this PR's scope tight after a review cycle; happy to delete it here instead if you'd prefer.

Also filed: `tbd-64aq` (S1 — invert `ids.yml` to ULID-keyed, which makes union merge collision-proof and deletes this problem class), `tbd-r0fv` (S3 — doctor check), `tbd-g2fd` (diff3 markers, confirmed pre-existing on `main`).

44 tests in the new file, 2,076 across 138 files, `pnpm precommit` green.

---

Issue #244, opened 2026-08-18
State model: tbd cannot express canceled or duplicate, and Linear/GitHub both can


jlevy:
tbd has one terminal status, `closed`. Linear has three terminal state types; GitHub has three terminal close reasons. The collapse is lossy in both directions and silent in both.

Found while reconciling ~105 mirrored epics in a large repository. Two tracked plans whose specs are explicitly `status: Superseded` — one says the approach "was never executed as a standalone effort". Those are **canceled**, not done, and there is no way to say so:

- `tbd close` files them in Linear as **Done** and stamps a completion date (`adapter.ts:134` sends `completedAt` whenever status is `closed`). Linear tracks `completedAt` and `canceledAt` separately and reports on them, so that is wrong data, not just a wrong label.
- The fallback, `deferred` → **Backlog**, understates it. They are not waiting; they are abandoned.

## The three models, as they actually are

All three verified by introspection rather than recollection.

**tbd** (`schemas.ts:90`) — one flat enum:

```ts
z.enum(['open', 'in_progress', 'blocked', 'deferred', 'closed'])
```

**Linear** — team-named workflow states, each carrying a canonical `type`: `triage`, `backlog`, `unstarted`, `started`, `completed`, `canceled`, `duplicate`. Separately, `IssueRelationType` is `['blocks', 'duplicate', 'related', 'similar']`.

**GitHub** — two axes, and importantly *three* enums:

| Enum | Values | Where |
| --- | --- | --- |
| `IssueState` | `OPEN`, `CLOSED` | the state |
| `IssueStateReason` | `REOPENED`, `NOT_PLANNED`, `COMPLETED`, `DUPLICATE` | read side |
| `IssueClosedStateReason` | `COMPLETED`, `NOT_PLANNED`, `DUPLICATE` | **write side**, on close |

`CloseIssueInput` also takes `duplicateIssueId: ID`. `ReopenIssueInput` takes no reason at all — `REOPENED` is derived by GitHub, never set by a client.

### What the comparison actually shows

**1. Both providers model "duplicate" twice.** GitHub: `stateReason: DUPLICATE` *plus* `duplicateIssueId`. Linear: a `duplicate` state type *plus* `IssueRelationType.duplicate`. Two independent designs reached the same conclusion: the reason alone is useless without the pointer. Any tbd design that stores `duplicate` as a bare value repeats a mistake both providers already corrected.

[...]

PR #245, opened 2026-08-18
feat: the state and actor axes, complete and dogfooded, plus the OS-351 fix

jlevy:
Started as a fix to one triage rule in `update-specs-status`. That rule kept running into a model tbd did not have, so this now carries the fix **and** both axes it needed, complete: 30 commits, every phase of both specs implemented, dogfooded against the real Linear workspace, plus a data-loss blocker reported against the branch mid-flight.

45 files, +5139/−102. Three new source modules (`slots.ts`, `actor-binding.ts`, `pause.ts`), five new test files.

---

## 1. What was broken

**tbd could not say why work ended, or that it had stopped.** One terminal status, `closed`. Cancelling a bead filed it in Linear as **Done** with a completion date — wrong data, not a wrong label, since Linear tracks and reports `completedAt` and `canceledAt` separately. Inbound, all three terminal state types collapsed back to `closed`, so a human setting Canceled in Linear had it erased permanently on the next pull.

**Pausing destroyed the one fact that made a pause interesting.** The only move was `--status deferred`, which overwrote `in_progress`, so "begun then set down" and "never started" became the same bead.

**One actor field for two questions.** `assignee` was in use on **zero** beads here: the human is the same on every bead so it carried no information, and the agent was not expressible at all. Meanwhile `tbd start` claimed beads by writing the agent *into* `assignee` — the slot reserved for whoever is accountable.

**Board position decided where work landed.** State resolution picked the lowest-position state of each type. Dragging a row in Linear's workflow editor is invisible to a sync and silently changed where work went.

## 2. The triage rule (where this started)

The table added in #241 said `any | 0 open children | 0 unchecked` → *closable*. Vacuously true for an epic that decomposed into nothing, so a spec nobody ever broke into beads read as complete. Measured: **33 candidate specs, 23 with no beads at all** — following it would have filed two dozen never-started plans as delivered.

The fix takes *total* children as a fourth signal, separating three cases the old table fused:

| Children (total) | Open | Unchecked | Meaning |
| --- | --- | --- | --- |
| **>0** | 0 | 0 | finished |
| **0** | 0 | 0 | **never decomposed — belongs in `draft/`** |
| **>0** | 0 | many | **stalled — began and stopped** |

Four further gaps from the same pass: status lines don't survive keyword matching (`Active (4 of 10 items landed…)` contains "landed" while saying the opposite); 22 of 104 epics had no `spec_path` and were invisible to the process meant to catch them; spec folders are a state vocabulary, now documented as one including `paused/`; and moving a spec is a link rewrite, with a sweep finding 1323 dangling references, 982 mechanically repairable.

## 3. The model

**Two axes beside `status`, each valid at one end.** `resolution` (completed/canceled/duplicate) and `hold` (blocked/paused), plus `started_at` and `duplicate_of`. Absent `resolution` reads as completed, so no bead needs backfilling and a plain `tbd close` is byte-identical to before. Position and modifier are independent, so `in_progress` + `paused` says "begun, then set down" without losing either half.

**Slots.** A board has more columns than tbd has statuses, and the extra ones are not extra *positions*: In Progress, Paused, Blocked and In Review are all started work; Todo vs Backlog is readiness; Done vs Canceled is resolution. A slot is that flattening done once, in the direction that keeps information. Nothing stores one — it is derived, so there is no second source of truth.

**Identity resolves instead of being configured.** `user_map` made every person a config entry keyed by an email someone had to paste. Humans now resolve against the tracker's own directory and bind **by provider user id**, so a rename keeps working and is merely reported as drift. Matching is exact at every step; several matches return ambiguous rather than the first hit, because a wrong actor publishes work under someone else's name and nothing downstream would flag it. Agents carry the identity tbd already mints (`agid-{ulid}`) and never publish except through an explicit `agent_map`.

**Corrections to the specs as written.** Linear assigns to humans and delegates to agents *only*, so the original human-delegate → `delegateId` mapping row was wrong. Bridge state syncs across clones rather than staying clone-local. Both specs were reworked so state owns the lifecycle and actor owns identity, with no cross-patching.

## 4. Implementation, in order

| Commits | What |
| --- | --- |
| `24d740b1`, `11352359` | The triage rule and the dangling-`spec_path` sweep |
| `c6c1e33f` … `89c97de9` | Both specs reworked; review findings closed |
| `aa49219e` | `resolution`, `duplicate_of`, `delegate`; `tbd close --as` |
| `66fcbc58` | `tbd start` claims via `delegate`; `--delegate` on create/update |
| `2c0e66f4` | Terminal mapping lossless both ways; states resolve by **name** |
| `1a39d717` | A push names the fields it drops |
| `ff182f1a` | The hold axis: `tbd pause` / `tbd resume`, `started_at` |
| `f8719390` | Directory-resolved identity, bound by provider user id |
| `53b84ba3` | **OS-351**: a null assignee no longer clears the tracker's |
| `9acec7a8` | Delegates publish via `agent_map` |
| `8fd37a11`, `c799d9af` | Board provisioning; offline resolution table in `doctor` |
| `c17077be`, `52e5aa4e`, `45f41d04` | Slots: vocabulary, matrix, both sides wired |
| `db92c788` | Refinement record settled **by experiment** |
| `2e4f1b82` | The `color` Linear requires when creating a state |
| `cc0dae7c` | Windows cleanup fix inherited from the upstream merge |

### The three that deserve detail

**OS-351, reported against this branch mid-flight, and partly mine.** A linked bead with `assignee: null` pushed that null outbound as an explicit unassign, clearing whoever a human had assigned. Populating `user_map` — the *documented prerequisite* for assignee sync — was by itself enough to start erasing people. An absent assignee is *no opinion*, not an instruction, so `canPushAssignee(null)` is now false. The second half: `field_sync.fields.assignee` was never consulted outbound at all, so the conservative default was not conservative in the direction that destroys data. **My own Phase 2 work had widened this** — directory resolution made the null-clear reachable with no `user_map` at all — and that regression is reverted with the fix.

**The slot upgrade, which nearly shipped wrong twice.** Every base written before slots holds a status, and a status cannot express what slots exist for: `open` could have meant Todo or Backlog and never recorded which. On this repository that is **231 linked pairs**. Read naively, the first sync after upgrading pushes a state write to every mirrored issue; the bulk guard would catch it, but the correct number is zero. A migrated base is therefore marked **coarse**, and while coarse the comparison is band-level. I caught two errors here only by checking that each test *fails without its fix* — it didn't, twice: the first migration turned a distinction nobody recorded into a pull, and the first test had local and remote agreeing so the base was never consulted. Removing the tolerance now fails exactly the two zero-writes tests while cross-band detection still passes.

**The refinement record, settled by measurement.** The spec left this open and said to settle it by testing concurrent behaviour. Both candidates ran through a **real git merge**: `extensions.<provider>` merges per namespace last-writer-wins, and a clone recording a refinement while another refreshed `external_key` **lost the key outright**. The link record resolves by newest observation as a whole record, so refinement and link cannot disagree — which is also what the schema already said: the bead records identity, the link record records dynamics. The experiments stay in the tree as `refinement-record.test.ts`.

## 5. Testing

- **2384 vitest tests, 0 failures**, run through `scripts/scrub-git-env.mjs` as the pre-push hook does.
- **1101 tryscript golden CLI tests**, 0 failures.
- **Five new test files**: `slots.test.ts`, `actor-axis.test.ts`, `hold-axis.test.ts`, `actor-binding.test.ts`, `refinement-record.test.ts`. `tbd start` previously had **no coverage at all** — not a unit test, not a tryscript.
- **CI green on every pushed commit**, all seven checks, four platforms.
- **Every commit through the full hook gate.** One early push used `--no-verify` before the worktree had `node_modules`; that was wrong, the worktree was set up, and every commit since has gone through.
- **Guards verified by removing them.** The OS-351 regression test, the zero-writes tolerance, and the naive-migration case were each confirmed to fail without their fix. Two of them initially did not, and were rewritten.

### Dogfooded live (2026-08-20, team `OS`)

- Provisioning bound all six existing columns and reported nothing to do — idempotent, binding rather than duplicating.
- Creating a **Paused** column placed it at **position 1003**, after In Review (1002) and the defaults: the explicit-position rule on a real board, not a fixture.
- `tbd pause` on a real epic set `hold: paused` with `status: in_progress` intact; OS-337 landed in **Paused/started** with Linear's `startedAt` preserved. `tbd resume` returned it to In Progress.
- `tbd close --as canceled` on `tbd-f8y4` — a genuinely superseded epic, the case that motivated #244 — landed as [OS-242](https://linear.app/finterm-ai/issue/OS-242) in **Canceled** with `completedAt=null` and `canceledAt` set.
- The `tbd start` migration ran on real data: agent-shaped assignees 1 → 0.
- Field-level skip reporting named the OS-351 guard by reason.

**Dogfooding found a bug, which is the point of it.** `WorkflowStateCreateInput.color` is `String!` in Linear's schema and the mock did not model it, so provisioning passed every test and failed on the first live create. This is the **third** defect here traceable to the mock being kinder than the API, so the mock now enforces the requirement rather than only patching the caller. The failure itself behaved well: `0 created, 1 blocked`, the API's own message, board untouched.

## 6. Loose ends

Everything in the earlier draft of this section has been either fixed or filed. Two were real defects and are fixed here:

**The exhaustiveness guards had been dead since f08** (`f62e7776`). `FIELD_STRATEGIES` and `ISSUE_CHANGE_FIELD_ORDER` both declare `Record<keyof Issue, …>` and document themselves as making a new field a compile error. `.passthrough()` widens `keyof Issue` to `string | number`, so neither had bitten — proven by adding six fields this session with no type error. Both now derive from the pre-passthrough shape, and a canary field fails both files. Turning the guard back on immediately found what it was hiding: `docs` and `refs` were added in f08 and never registered, so **`tbd changes` has never reported a doc or ref being attached to a bead**.

**A skip-only full sync still said "nothing to do"** (`b5ddbef9`). The last of the OS-351 family: a run whose only outcome was a field it could not publish reported nothing to do, and the line naming that field sits behind the early return that gates it. Both `nothingToDo` computations now count `skippedPushes`, and `skipPush` stops being hardcoded to `assignee` — the base-recording rule is a property of skipping, not of one field.

**The f08 no-bump claim now has a test** (`c9d75850`). Both specs declined a format bump on the strength of new fields round-tripping through an older build. Three cases, the important one being a three-way merge, since `tbd sync` rewrites beads during ordinary merges and a strip there loses the field for everyone.

**Two verifications completed rather than deferred:**

- **Directory-bound assignment, live.** With `user_map: {}` and `assignee: merge`, the bare handle `josh` resolved through the workspace directory and published to OS-337, and a binding was written keyed by **provider user id** carrying handle and display name and **no email** — the design proven on the real API. Config restored to `local` afterwards, so it stays opt-in.
- **Draft and Blocked columns, live.** Created at positions 1000 and 1004, placed by band. A real bead walked through both: `open` + blocked correctly lands in **Backlog** (un-started held work has no started-type column), and `in_progress` + blocked lands in **Blocked/started**. The board now carries every column the projection describes.

**Three unbuilt features, now built** (`79c0a871`):

- **Ambiguous state resolution asks, once.** The resolver reported candidates and refused to guess, which was the safe half. Setup now lists them, takes the answer, and writes it into `identity.state_map` so the question is asked once rather than per run. Silent with no ambiguity or no TTY.
- **Board order contradictions are reported and repaired.** A terminal column mid-board is the accident explicit positioning prevents for states tbd creates, but it predates tbd on any hand-built board. Detected always, fixed under `apply`. This one had a genuine conflict: an existing test caught the first version moving `Done`, which no map mentioned. The rules collide and **the map wins** — a tidier board does not justify writing to a column nobody asked tbd to manage. Reordering is limited to mapped states.
- **Invariant breaches read as sentences.** `tbd update x --hold blocked` on closed work printed a JSON dump around a message already written for a person. It now prints `hold: hold is only valid on work that is not closed`. This path only became reachable because this branch added the invariants, so it was this branch's to clean.

**Two pre-existing beads this work resolved** (`765aed9f`):

- `--due` and `--defer` now accept a plain date. Half that bug was already fixed by the error formatting above; this is the usability half, matching `tbd pause --until`.
- "Populate `identity.user_map` so assignee sync stops being skipped" is closed as **canceled**: superseded by directory resolution, verified live with `user_map: {}`.

**Behaviour changes worth a second opinion, unchanged:**

1. **`tbd ready` no longer treats `assignee` as "taken."** It excludes work that is *held* or has a *delegate*. One line to revert.
2. **Assignees do not publish unless `field_sync.fields.assignee` is `merge`.** Deliberate after a data-loss bug. Now live-verified rather than assumed.

**Separately filed upstream** (`tbd-639t`, in the main repo): the `update-specs-status` shortcut has accreted guidance that reads as workarounds for things tbd cannot do itself — triage that needs a query rather than a procedure, spec lifecycle as an unmodelled state machine, spec moves as a link rewrite the tool should perform, and status lines that are prose the process parses. Each is a candidate feature; the shortcut should shrink as they land.

## 7. Beads

Both epics closed. `tbd-og20` (state): Phases 1–5. `tbd-ncux` (actor): Phases 1–4. `tbd-k0rd` (OS-351) closed. `tbd-ha0a` tracks the review cycle. Every bead carries notes naming what landed and what did not, so the next session starts from fact rather than from this description.

---

> [!NOTE]
> **High Risk**
> Touches core sync/reconcile, Linear state and assignee/delegate publishing, and fixes a data-loss class bug on assignee push; large behavioral surface for mirrored repos.
> 
> **Overview**
> Introduces the **state axis** (`resolution`, `duplicate_of`, `hold`, `hold_until`, `started_at`) and **actor axis** (`delegate` separate from `assignee`), with CLI support via `tbd close --as`, `tbd pause`/`resume`, `tbd start` claiming on `delegate`, and create/update flags for `--delegate` and `--hold`.
> 
> **Linear sync** maps terminal outcomes and holds losslessly, resolves workflow states **by name** (with optional `state_map` / provisioning), derives **slots** for reconcile when both sides can name a column, and publishes delegates only through `agent_map`. Human assignees can resolve via directory **`listMembers`** + per-provider bindings under `bridge/<provider>/users/`, with `user_map` still overriding.
> 
> **Safety and visibility:** outbound assignee writes require `field_sync.fields.assignee: merge` and no longer treat a null assignee as “clear the tracker” (**OS-351**). Mirror and sync summaries report **field-level skips** (`skippedFields` / `skippedPushes`), and `nothing to do` no longer hides skipped pushes. `tbd doctor` flags agent-shaped `assignee` values and prints offline state-resolution plans.
> 
> Docs add the two active plan specs and expand **`update-specs-status`** triage (never-decomposed vs finished, `paused/` folder vocabulary, `spec_path` repair).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 765aed9f7f8273589f0a640b121d67d6cb45fc62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy:
Field-tested against the repository that prompted it (116 active specs, ~105 live epics). Two things came out of running it end to end, and I've pushed both to this branch.

**The lifecycle folders work, and `paused/` earned its place immediately.** 21 specs moved to `draft/` (never decomposed), 7 to a new `paused/` (began and stopped), taking `active/` from 116 to 88. The 7 had been sitting in `active/` for exactly the reason the doc now gives: `draft/` denies they started, `done/` is false, `archive/` overstates abandonment.

**The missing step was repointing beads after a move — and it compounds.** A bead whose `spec_path` still names the old folder makes its spec look like it has *no beads*, which the step 2 triage reads as *never decomposed*. So a finished or stalled workstream gets refiled as a draft. The damage lands on the **next** reconciliation rather than the one doing the move, which is why nobody notices.

Because nobody notices, it accumulates. Auditing the whole repository found **1323 dangling `spec_path` references, 982 of which were just a spec that had moved folder** — every past move had stopped at `git mv`. Repaired in one scripted pass, zero ambiguous, leaving 341 that need judgment (292 deleted specs, 49 deleted handoffs/research docs).

So the sweep is now a repository-wide step rather than a per-move reminder, bucketed by cause since only the moved-folder bucket is safe to script, with a matching invariant in the validation list.

Also confirmed in passing: the original bug this PR fixes was real in practice, not just in theory. The old rule matched 33 candidate specs on that repository, 23 of which had no beads at all.

Downstream application: finterm-ai/trading#339.

jlevy:
# Review: PR #245 — spec-triage fix and the tracker state/actor plan specs

## Scope

Full diff `main...634de98e` (4 files: `packages/tbd/docs/shortcuts/standard/update-specs-status.md`, the two plan specs under `docs/project/specs/active/`, one rename), all 11 commits, reviewed per `tbd shortcut review-code` with the GitHub PR scope. The diff is Markdown-only, so language-specific rules do not apply; the review checked internal consistency, factual accuracy against the code the specs cite (`mapping.ts`, `adapter.ts`, `reconcile.ts`, `start.ts`, `agent-identity.ts`, `bridge-state.ts`, `schemas.ts`), doc-guideline compliance, and design soundness of the specced plan. Reviewer note: the last two commits were authored in this session, so this pass deliberately re-read them adversarially rather than assuming them correct.

## Summary and verdict

**Verdict: approve with findings — two Medium spec gaps to close before implementation starts, two Low wording fixes.** The triage-rule fix is correct and well-evidenced; the restructured specs are coherent and implementable. The two Medium findings are both places where the new slot model is silently partial over states the current system can actually produce: an explicit `hold: blocked` on an un-started bead, and the legacy `blocked`/`deferred` status values that remain valid in the enum. Both are one-paragraph fixes in the spec, but left unfixed they would surface as undefined behavior mid-implementation in Phase 3.

## Findings

**R1 (Medium) — The slot ladder drops `hold: blocked` on open beads.**
`plan-2026-08-18-tracker-state-model-and-linear-mapping.md`, "The slot vocabulary" precedence ladder (steps 5–7) and the "Open end, by slot" table.
Step 5 maps `open` + `hold: paused` → `backlog`, but `hold: blocked` is valid "when open or in_progress" per the Approach block, and an open bead with `hold: blocked` falls through step 5. If its dependencies are met it reaches step 6 and computes `todo` — an explicitly blocked bead rendered as ready-to-begin. The underlying gap: the spec never states how `hold` interacts with readiness, and `tbd ready` is the input to the todo/backlog split.
**Fix:** make step 5 cover any `hold` on an open bead (→ `backlog`), state the rule that a bead carrying any `hold` is never `ready`, add that to the Phase 2 checklist (`tbd ready` excludes held beads), and change the `backlog` row's condition from "not ready or `paused`" to "not ready or held".

**R2 (Medium) — The slot function is undefined for legacy `blocked`/`deferred` statuses.**
Same file, Approach block (`status: open | in_progress | closed # position; unchanged`) and "The sync algorithm".
The current enum is five-valued: `mapping.ts:50–55` maps status `blocked` → `started` + `tbd:blocked` and status `deferred` → `backlog` + `tbd:deferred`, and the Non-Goals explicitly keep both working. The Phase 3 "legacy path keeps `statusToLinear`/`statusFromLinear` byte-for-byte" bullet covers repositories with no `state_map` — but a repository that adopts `state_map` still holds beads whose status is `blocked` or `deferred`, and the ladder assigns them no slot. The Approach block's "unchanged" comment also reads as if the enum were three-valued.
**Fix:** define the legacy rule in the slot section — status `blocked` → slot `blocked`, status `deferred` → slot `backlog`, matching today's outbound rendering, in force until the fold-in the Non-Goals defer — and annotate the Approach block that `blocked`/`deferred` remain valid enum values handled by that rule.

**R3 (Low) — Resolver-section evidence conflates pre- and post-provisioning state counts.**
Same file, "State resolution: names, not board position."
The section says one real team "already has four `started` states (In Progress, In Review, Paused, Blocked) and two `backlog` states". Blocked (and Draft) were created *by* tbd's provisioning; the pre-provisioning observation — the one that argues type-based resolution was already broken before tbd touched anything — was three `started` states. The four-state count belongs to the "Evidence from a provisioned board" section, where it is correctly framed.
**Fix:** restore the pre-provisioning claim here ("already had three `started` states … before tbd provisioned anything") and let the evidence section carry the provisioned four.

**R4 (Low) — Sibling headers overstate the cross-spec dependency.**
Both specs' **Sibling** front-matter lines say the actor spec "depends on this spec's resolver" / "depends on that resolver". Actor Phase 1 (delegate field, claim verb, reporting) is deliberately independent; only Phase 2 reuses the resolver-and-ask machinery. Both Implementation Plans state this correctly; the headers contradict them.
**Fix:** scope the header wording to the identity phases (e.g. "its identity-binding phase reuses this spec's resolver-and-ask machinery").

## Suggestions (non-blocking)

**S1** — `agent_map` example comment says "the only identity config that remains" while `user_map` persists as a parsed override two sections earlier. "The only identity mapping this design needs in config" would be exact.
**S2** — Actor spec Approach block: `assignee: <handle> # accountable. Unchanged field, unchanged meaning` sits oddly with the Background's admission that `tbd start` currently writes agent names there. "Restored meaning" is closer to what the spec actually does.

- **`update-specs-status.md` step renumbering**: all internal references checked ("See step 6" at line 63, "step 2" at lines 165/225/235) — every one points at the right renumbered step.
- **Margin-level tables and bullets inside numbered steps**: GitHub preserves explicit ordinal numbering across list interruptions, and the file already used this pattern before this PR; renders correctly.
- **`adapter.ts:794` citation**: the state-id-only-when-status-in-patch guard sits at the cited block (787–799); close enough to be a working reference.
- **Human delegate → GitHub "second assignee"**: GitHub issues accept up to 10 assignees; the future-mapping row is plausible as specced.
- **`--as <human-name>` claims escaping the migration heuristic**: a human who claimed via `tbd start --as josh` leaves `josh` in `assignee`, which the shape heuristic will not move — and should not, since the accountable-human reading is correct there. Doctor-plus-confirmation covers the ambiguous residue.

## Design assessment

The two-namespace identity model is the right call and is better than both alternatives actually on the table: the original `kind`-bearing `user_map` (a hand-maintained roster that the shipped `agid-` machinery already obsoletes, and whose omission-as-policy is exactly the silent-skip defect the spec documents) and a central `people/` registry (single-file concurrent-write hazard, correctly set aside with the revisit trigger named). Binding by provider user id mirrors the state resolver's resolve-by-name/bind-by-id pattern, so the config surface stays uniform. The per-provider split matches how bridge state is already laid out, and the stub-second-adapter test is the right forcing function to keep `listMembers()` honest before a GitHub adapter exists. The main structural risk is not in this PR: `tbd ready`'s semantics quietly become load-bearing for the board projection (todo/backlog split) and now for hold (R1), so Phase 2's definition of readiness needs to be treated as a contract, not an implementation detail.

The slot vocabulary as the widened canonical value for the existing three-way matrix is the strongest part of the state spec — it gets paused/blocked/in-review onto the board without touching the engine, and the owned-refinement rule (band for the matrix, exact state id outbound) cleanly absorbs team-custom states. The base migration note (zero writes on an unchanged repository, pinned by test) closes the classic mass-push upgrade hazard.

## Documentation

- Spec cross-links and the research-doc link all resolve post-rename (verified against the filesystem).
- `docs/docs-overview.md` already documents the `paused/` spec folder, consistent with the shortcut's new lifecycle table.
- The PR description's bead tables match the live bead graph (verified via `tbd show` on all nine).

## CI status

All checks green at review time: Test (ubuntu Node 22.12.0, ubuntu Node 24, macos Node 24, windows Node 24), Coverage & Lint, Benchmark, DeepSource Secrets. Cursor Bugbot skipped (usage limit).

jlevy:
Addressed the review findings in `89c97de9`:

- **R1: fixed** — ladder step 5 now catches any `hold` on an open bead (→ `backlog`); the rule "a bead carrying any `hold` is never `ready`" is stated as a contract on `tbd ready` and pinned by a Phase 2 checklist test; the `backlog` row reads "not ready or held".
- **R2: fixed** — legacy rule added after the ladder: status `blocked` → slot `blocked`, status `deferred` → slot `backlog`, matching today's `statusToLinear` rendering, in force until the deferred fold-in; the Approach block annotates that both legacy values remain valid; a Phase 3 test pins the behavior in a `state_map` repository.
- **R3: fixed** — the resolver section states the pre-provisioning count (three `started` states) as its evidence; the provisioned four-state board stays in the evidence section where it is correctly framed.
- **R4: fixed** — both Sibling headers now scope the dependency to the identity-binding phase, matching the implementation plans.
- **S1: applied** — `agent_map` comment reads "the one identity mapping this design needs in config".
- **S2: applied** — the `assignee` comment reads "Existing field, restored to that meaning".

Tracked as `tbd-ha0a` with one child bead per finding, all closed. Pre-push gate ran in full (format + flowmark check, typecheck, zero-warning lint, build, 149 files / 2246 tests) and CI is green on the fix commit across all seven checks.

Issue #246, opened 2026-08-19
Actor model: tbd cannot express who is accountable separately from who is executing, and Linear models that as a first-class axis


jlevy:
Actor model: tbd cannot express who is accountable separately from who is executing, and Linear models that as a first-class axis

tbd has one actor field, `assignee: z.string()`. Linear has two — `assignee` (accountable) and `delegate` (acting) — and its entire agent platform hangs off the second. GitHub has a comparable split via agent assignment. The collapse is lossy, and today it is silent.

This is the same shape as #244: a single tbd enum standing in for two axes that both providers model separately. #244 is the closed end of the status axis; this is the actor axis.

Found while assigning two epics in a ~900-open-bead repository where nearly all execution is done by coding agents. `assignee` is used on **zero** beads there — not from neglect, but because there is nothing useful to put in it: writing the human means the field is constant and carries no information, and writing the agent is not expressible at all.

## The three models, as they actually are

Verified against the installed `get-tbd@0.7.1` dist and the Linear SDK schema rather than recollection.

**tbd** — one opaque string:

```ts
assignee: z.string()
```

The Linear adapter maps it to exactly one field:

```js
assigneeId = patch.assignee === null ? null : await this.res…
```

and the default field-flow rule carries the design assumption in a source comment:

```js
assignee: FieldFlowRule.default("local")   // "`assignee` stays local because tracker assignees are people"
```

That comment is correct about 2024 and load-bearing today. It is the assumption this issue is about.

**Linear** — two actor fields, publicly mutable:

| Field | Meaning | Input |
| --- | --- | --- |
| `assignee` | accountable human | `IssueUpdateInput.assigneeId` |
| `delegate` | acting agent | `IssueUpdateInput.delegateId` |

Linear states the taxonomy explicitly as *delegation, not assignment*: the issue keeps a human assignee while the agent is the delegate. Setting `delegateId` automatically creates an `AgentSession` and fires the target agent's `AgentSessionEvent(created)` webhook. Agents are OAuth apps installed with `actor=app`, gated by `app:assignable` / `app:mentionable`, free of a seat, and surfaced natively in delegate-filtered views and Insights segmented by delegate.

**GitHub** — assignees are a list, and agent assignment is a distinct mechanism layered on it rather than a second value in the same slot.

### What the comparison shows

[...]

PR #247, opened 2026-08-19
Plan the actor axis and the default board projection

jlevy:
Stacks on #245. Base is its branch, so the diff here is one new spec; retarget to `main` once #245 merges.

The sibling spec on that branch (`plan-2026-08-18-tracker-state-model-and-linear-mapping.md`, epic `tbd-og20`, #244) separates *where work sits* from *why it is there*. This is the second axis of the same model — *who owns a bead* versus *who is doing it* — filed as #246.

## Why they are one design

tbd has one actor field, `assignee: z.string()`, and the Linear adapter maps it to `assigneeId` alone. Linear has two: `assignee` (accountable) and `delegate` (acting), and its entire agent platform hangs off the second. Same defect as #244, same shape, same file format, same adapter.

Measured on the repository that produced #244 and #245 — ~900 open beads, ~105 mirrored epics, nearly all execution by coding agents — `assignee` is used on **zero** beads. Not neglect: the human is constant across every bead so the field carries no information, and the agent is inexpressible. `delegateId`, `agentSession`, `app:assignable`, and `actor=app` appear in **zero** files in the dist today, so the agent side is new surface rather than a modification.

tbd states the assumption in its own source:

```js
assignee: FieldFlowRule.default("local")   // "`assignee` stays local because tracker assignees are people"
```

Correct when written. Linear's 2025 agent platform is what changed it.

## What the spec proposes

`delegate` beside `assignee`, plus an actor `kind` on `user_map` so "never publish this actor" is a property of the actor rather than its absence from a table:

```yaml
user_map:
  josh: { email: josh@example.com, kind: human }
  claude: { kind: agent }                              # no address: never pushed
  cyrus: { kind: agent, linear_app_user_id: <uuid> }   # pushable as a delegate
```

The bare `alias: email` form keeps parsing as `kind: human`. A Linear agent is an app user, not a person with an email, so the current value domain cannot name one even where you want to.

The governing rule this enforces: **issues in Linear are managed per human, not per agent.**

## The board projection

The two axes together make the open end expressible, which is where a planning board earns its keep:

| Column | Linear type | tbd condition |
| --- | --- | --- |
| Draft | `backlog` | `open`, not ready |
| Todo | `unstarted` | `open`, ready |
| In Progress | `started` | `in_progress`, no hold |
| Paused | `started` | `in_progress` + `hold: paused` |
| Blocked | `started` | `in_progress` + `hold: blocked` |
| Done | `completed` | `closed` + `completed` |
| Canceled | `canceled` | `closed` + `canceled` |

Draft-vs-Todo is **readiness**, which `tbd ready` already computes — no new field, no bookkeeping. Paused-vs-Draft is **`started_at`**, which is the strongest practical argument for that field landing early: without it the two are indistinguishable, and that is precisely what `deferred` overwriting `in_progress` loses today.

This amends the sibling's open-end table, which maps `open` → Todo unconditionally and `open + paused` → Backlog. Precedence between the two routes to Backlog/Draft is called out as an open question rather than silently resolved.

## Three surfaces, one vocabulary

#245 noted that its `paused/` folder is the spec-level instance of #244's gap. The spec finishes that observation:

| Spec folder (#245) | Bead state (#244) | Column |
| --- | --- | --- |
| `draft/` | `open`, not ready | Draft |
| `active/` | `open` ready, or `in_progress` | Todo, In Progress |
| `paused/` | `in_progress` + `hold: paused` | Paused |
| `done/` | `closed` + `completed` | Done |
| `archive/` | `closed` + `canceled` | Canceled |

One lifecycle vocabulary, three surfaces, expressible on one of them today.

## Not in scope

Agent **presence** — which session holds a bead right now — is explicitly excluded. Heartbeat cadence is seconds, `extensions` merges whole-object LWW, and sync-branch commits are the wrong granularity; that is a TTL lease in a database. `delegate` is the durable, low-frequency fact.

Refs #246, #244, #245.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change; no runtime, API, or schema code is modified in this PR.
> 
> **Overview**
> Adds **`plan-2026-08-18-actor-axis-and-board-projection.md`**, a draft design spec (sibling to the tracker state / Linear mapping plan) that describes future tbd behavior rather than implementing it.
> 
> **Actor axis:** optional bead field **`delegate`** next to **`assignee`**, plus **`user_map`** entries with **`kind`** (`human` vs `agent`) and optional **`linear_app_user_id`**, so humans stay accountable on Linear while agents can be named locally and only pushed as **`delegateId`** when publishable. The spec also calls for **reported** skips when sync cannot push an actor (fixing silent `skippedPushes` / misleading success summaries) and clarifies that **`field_sync` `local`** is not a second assignee gate.
> 
> **Board projection:** introduces a **slot** vocabulary (`backlog`, `draft`, `todo`, `in_progress`, `paused`, `blocked`, etc.) mapped through optional **`state_map`**, with reconcile comparing **slots** instead of the five-value status enum, **owned refinements** (e.g. Draft vs Backlog) preserved across sync, opt-in provisioning via **`tbd integration setup`**, and a phased implementation plan (actor axis → engine slots → setup/doctor).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17254c3078936f147bdbd4eb6e443d07b4e0e702. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/project/specs/active/plan-2026-08-18-actor-axis-and-board-projection.md
+++ b/docs/project/specs/active/plan-2026-08-18-actor-axis-and-board-projection.md
@@ -1,0 +1,322 @@
+---
+title: Plan Spec
+description: Separate who is accountable from who is executing, so an agent can hold a bead without displacing the human who owns it, and project the resulting lifecycle onto a default Linear board that distinguishes planned from ready and paused from never-started
+author: Joshua Levy (github.com/jlevy) with LLM assistance
+---
+# Feature: Actor Axis and Board Projection
+
+**Date:** 2026-08-18
+
+**Author:** Joshua Levy (github.com/jlevy) with LLM assistance
+
+**Status:** Draft
+
+**Design discussion:** [#246](https://github.com/jlevy/tbd/issues/246) — the provider
+comparison and the measured evidence are argued there and are not re-argued here.
+
+**Sibling:**
+[plan-2026-08-18-tracker-state-model-and-linear-mapping.md](./plan-2026-08-18-tracker-state-model-and-linear-mapping.md)
+(epic `tbd-og20`, [#244](https://github.com/jlevy/tbd/issues/244)) adds `resolution`,
+`hold`, and `started_at`, and replaces position-based Linear state resolution with
+name-based. This spec is the second axis of the same model and depends on that resolver;
+it does not restate the state design.
+
+## Overview
+
+tbd has one actor field, `assignee: z.string()`. Linear has two — `assignee`
+(accountable) and `delegate` (acting) — and its entire agent platform hangs off the
+second. The collapse is silent, and in an agent-driven repository it makes the field
+unusable in either direction.
+
+This adds `delegate` beside `assignee`, gives the identity table an actor **kind** so
+publishability is a property of the actor rather than of its presence in a map, and
+defines the default board projection that the two axes together make expressible.
+
+The state sibling separates *where work sits* from *why it is there*. This one separates
+*who owns it* from *who is doing it*. Same defect, same shape, same file format, same
+adapter — which is why they are planned as a pair.
+
+## Goals
+
+- Record that a human is accountable while an agent executes, which is the normal case
+  in an agent-driven repository and is currently inexpressible.
+- Keep agent identities out of a shared tracker **by construction**, not by remembering
+  to omit them from a config table.
+- Push a delegate to Linear when — and only when — the alias names a real Linear app
+  user.
+- Ship a default board that distinguishes *planned* from *ready* and *paused* from
+  *never started*, customizable per repository.
+- Report a field that a flow rule excluded, instead of reporting success for a write
+  that went nowhere.
+
+## Non-Goals
+
+- **Agent presence.** Which session is touching a bead right now is a TTL lease in a
+  database, not a bead field: heartbeat cadence is seconds, `extensions` merges as
+  whole-object last-writer-wins, and sync-branch commits are the wrong granularity.
+  `delegate` is the durable, low-frequency fact of who a unit of work is assigned to.
+- **Registering tbd as a Linear agent** (agent sessions, ACKs, typed activity streams).
+  Setting `delegateId` is a plain field write that happens to create an AgentSession on
+  Linear’s side; nothing here requires tbd to become an agent.
+- **GitHub adapter work.** The model is chosen to fit it; nothing here implements it.
+- Re-specifying `resolution`, `hold`, or state resolution.
+  That is the sibling spec.
+
+## Background
+
+Found during the same first-time Linear rollout that produced #244 and #245: ~900 open
+beads, ~105 mirrored epics, nearly all execution done by coding agents.
+
+`assignee` was in use on **zero** beads there.
+Not neglect — there was nothing useful to write.
+The human is the same on every bead, so the field carries no information; the agent is
+not expressible at all.
+Two independent config gates (`field_sync.fields.assignee`, defaulting to `local`, and a
+closed `user_map`) also mean the field cannot reach Linear without deliberately opening
+both, and neither gate says so when it excludes a field: a push reports
+`updated N, skipped 0, failed 0` while the assignee was never a candidate.
+
+tbd’s own source states the assumption that no longer holds:
+
+```js
+assignee: FieldFlowRule.default("local")   // "`assignee` stays local because tracker assignees are people"
+```
+
+That was right when it was written.
+Linear’s 2025 agent platform is what changed: agents are OAuth apps installed with
+`actor=app`, they appear in assignment menus under `app:assignable`, and Linear’s stated
+taxonomy is **delegation, not assignment** — the issue keeps a human assignee while the
+agent is the delegate.
+`delegateId`, `agentSession`, `app:assignable`, and `actor=app` appear in **zero** files
+in the tbd dist today, so this is new surface rather than a modification.
+
+The governing rule, from the repository this came from: **issues in Linear are managed
+per human, not per agent.** Linear is where a person decides what is on their plate.
+An agent in the assignee slot answers a question nobody asks there while destroying the
+answer to the one they do.
+
+## Design
+
+### Approach
+
+One field beside the existing one, and a value-shape change to the identity table:
+
+```
+assignee:  <alias>     # accountable. Unchanged field, unchanged meaning.
+delegate:  <alias>     # acting. Absent reads as "same as assignee".
+```
+
+```yaml
+integrations:
+  linear:
+    user_map:
+      josh: { email: josh@example.com, kind: human }
+      claude: { kind: agent }                              # no address: never pushed
+      cyrus: { kind: agent, linear_app_user_id: <uuid> }   # pushable as a delegate
+```
+
+The bare `alias: email` form keeps parsing as `kind: human`, so every existing config is
+valid with no rewrite.
+
+`kind` is what makes the closure a design rather than an omission.
+Today the only way to keep an agent out of a shared workspace is to leave it out of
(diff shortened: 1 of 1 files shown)


jlevy:
## Full review: will this work, and what does "default" actually get you

Reviewed the design against the real code (`packages/tbd/src/integrations/` at this branch, 7,265 lines across the engine and Linear adapter) and against a live provisioned board (team `FIN`, 96 issues, all ten columns). Verdict up front: **the design fits the existing engine with no new machinery — the change is the canonical vocabulary, not the sync algorithm** — and the riskiest piece is not sync at all but the two storage decisions (refinement record home, base migration). Both are now named in the spec with tests specified.

### Why the engine fits

Three facts from the code carry the whole feasibility argument:

1. **The reconcile engine is already value-agnostic.** `core/reconcile.ts` is a pure per-field three-way matrix over canonical values with a stored base; `local`/`remote` are ownership short-circuits with reported overwrites; `tie_break` handles both-changed. Nothing in it knows what a status *is*. Widening the compared value from the five-status enum to the ten-slot vocabulary touches the views and the decompose/compose steps, not the matrix.

2. **The adapter already reduces states losslessly enough to not fight humans.** `statusFromLinear` (`linear/mapping.ts`) reduces by type only, and `toInput` sends `stateId` only when `status` is in the patch (`adapter.ts:794`). Chain those and you get the no-fight property: a human dragging In Progress → Paused produces no canonical change, the matrix sees unchanged/unchanged, and the placement survives. **This is why the FIN board works by hand today, with zero upstream changes.** The design keeps the property and makes it deliberate (unmapped names become owned refinements).

3. **State selection is the one genuinely broken piece, and the sibling already replaces it.** `adapter.ts:634` picks one state per type by lowest position, with the comment "Map by type, never by name: names are user-editable, types are not." On the FIN board that comment's premise fails outright: four `started` states, two `backlog` states. Position is the *least* stable handle — the FIN board had Done/Canceled/Duplicate ordered *before* two `started` states (positions 3/4/5 vs 1002/2018.44) until repaired. Name-based resolution with id binding is the right call and this spec's projection depends on it, which fixes the stacking order: sibling Phase 1 → this spec's Phase 2.

### What works today vs. after (measured on the live board)

| Behavior | Today (v0.7.1) | After |
| --- | --- | --- |
| Human drags issue to Paused/Blocked/Draft/In Review | Survives (no-fight chain above) | Survives, and round-trips as `hold`/refinement |
| tbd closes a bead in Paused | Moves to Done — correct | Same, via `resolution` |
| tbd sets `blocked` on a bead in Paused | Moves to In Progress + `tbd:blocked` label — **column lost** | Moves to Blocked column |
| Unready open bead | Lands in **Todo** (`open`→`unstarted`, readiness never consulted) | Lands in Backlog; Todo means ready |
| Human cancels in Linear | Collapses to plain `closed`; distinction erased | `resolution: canceled` round-trips |
| Human moves issue to Draft | Reads as `open`; placement survives but means nothing locally | Owned refinement, preserved by rule |
| Team's own custom state (In QA) | Reads as `in_progress`; survives only until bead status changes | Owned refinement, preserved by rule |

The left column is a defensible interim: after repositioning, the type resolver picks the right defaults (In Progress at 2 beats Paused 2.2; Backlog 0 beats Draft 0.5), so **provision-and-reposition works with current tbd** as a manual board with correct automatic placement into the defaults. Worth stating in the rollout docs, since it's the state every adopter sits in until Phase 2 lands.

### The default-integration story, concretely

The spec now pins the flow the thread converged on, and I think it's right:

- **Setup proposes; the written config is the consent.** Fresh `tbd integration setup` shows the plan — bind-by-name (stock teams already have 7 of 10: Backlog, Todo, In Progress, In Review, Done, Canceled, Duplicate), create the missing 3 (Draft, Paused, Blocked) with explicit positions in slot order — and writes `state_map` on confirm. Absent map = today's behavior byte-for-byte, no prompt ever. This resolves the tension with the sibling's "Paused is the sole offered candidate" restriction without weakening its reasoning (a workflow state is team-wide): the confirmation footprint is identical, one setup decision, and no format bump (`f08` passthrough).
- **Re-run is the re-config path**: missing states (offer create), order contradictions (offer reposition — the FIN repair is the live precedent), renames (id bindings hold; doctor reports drift). Sync never provisions.
- **Custom maps are the same mechanism**: any subset, any names, omitted slots fall down the ladder (mapped state → carrier label + band default → band default, reported once), two-slots-one-state allowed with label disambiguation, and `tbd doctor` prints the resolved table offline.

### Risks, ranked

1. **Refinement-record home.** `extensions.<provider>` travels with the bead but merges whole-object LWW; the bridge base is per-pair but local to the syncing clone. Two clones syncing the same repo would disagree about a human's Draft placement under the second option. This is the one decision I'd settle with a test spike before implementation, and it's now the top open question.
2. **Base migration.** Every stored base holds five-value statuses. Without a mechanical rewrite on first slot run, every linked pair reads locally-changed and the first sync mass-pushes state writes — the bulk guard would catch the volume, but the correct number of writes on an unchanged repo is zero, and the spec pins that with a test.
3. **Readiness flips are silent.** A bead becomes ready when its *blocker* closes — nothing touches the bead itself. This works only because the engine reconciles every linked pair per run rather than a changed-set; worth a pinned test so a future changed-set optimization doesn't quietly break the Draft→Todo transition.
4. **`WorkflowState.type` is `String!`**, an open set (`mapping.ts` KNOWN_STATE_TYPES comment). The band-default fallback for unknown types is already fail-soft; the projection keeps that.
5. **Single `team_key`.** `state_map` is per-team by construction; multi-team repos need per-team maps eventually. Additive later; not blocking.
6. **GitHub degrades to labels for the whole open end** (no states, no snooze, no started concept). Real asymmetry, stated rather than papered over; the slot vocabulary is provider-neutral so nothing forecloses it.

### Two corrections from this pass, owned

- **The "two independent gates" claim in #246 was half wrong.** Reading the engine: `field_sync.assignee: local` is *not* a gate — `local` pushes and overwrites remote edits, reported (`reconcile.ts:324`). The one real gate is `user_map` → `capabilities.assignee` → `canPush` (`adapter.ts:198`). The observability gap is real but different than claimed: skips print as detail lines in `integration sync` (`integration.ts:483`, "push unsupported; left divergent") but never reach the summary counters (`integration.ts:370` counts pairs) or the umbrella `tbd sync` output — and the detail line names no reason. Spec Background rewritten; correction comment posted on #246.
- **The "sibling leans toward snooze" claim was wrong.** The sibling already maps paused to the named state with a label fallback. Live evidence *confirms* it rather than contradicting it, and per the design decision on this thread, snooze is now specified as a Linear-side layer tbd never writes: the column is canonical, `hold_until` stays tbd-native, and an issue a person snoozes is still in its column when it returns.

### What changed in the spec this pass

- New **“The sync algorithm”** section: slot precedence (resolution → hold → refinement → readiness → band default), name-based remote resolution, unmapped-names-as-owned-refinements (one rule covers In Review, Draft, and any team's custom state), pull decomposition, outbound ladder, base migration, field-level skip reporting with reasons.
- New **“Default integration and re-config”** section as above.
- Background corrected (gates); snooze paragraph corrected (confirms sibling; column canonical); phases rewritten (Phase 2 slots-in-engine, Phase 3 setup/provisioning); open questions updated — refinement home added, snooze and tbd-sets-Draft recorded as settled.

### Suggested order

Sibling Phase 1 (resolver + `resolution`) → this Phase 1 (`delegate` + `user_map` kinds) → base-migration + slots (this Phase 2) → setup/provisioning (this Phase 3). The refinement-home spike can run parallel to anything. Nothing here blocks #245 merging first — this PR retargets to `main` after.

PR #248, opened 2026-08-19
fix: doctor warns when npm's global bin is off PATH; add the agent-session-bootstrap guideline

jlevy:
## Summary

Two changes from downstream experience building a pinned-toolchain session bootstrap in
[metaproc](https://github.com/jlevy/metaproc/pull/19), kept upstream only where tbd is the
right owner:

1. **A real bug in tbd's own instruction.** `npm install -g get-tbd@latest` can exit 0 and
   leave `tbd` unresolvable. `tbd doctor` now says so.
2. **A new guideline**, `agent-session-bootstrap`, for the general question that work raised:
   when should a repository install its own toolchain at session start, and when should it not?

Deliberately **not** upstreamed: the bootstrap script itself. See "What stays downstream" below.

## 1. `fix`: doctor warns when npm's global bin is not on PATH

`npm install -g get-tbd@latest` is tbd's canonical install and upgrade instruction — it appears
in `skills/tbd/SKILL.md`, in `output.ts`, in four `doctor.ts` suggestions, and in the
incompatible-format error in `tbd-format.ts`.

When npm's global bin directory is not on PATH, that command **exits 0, prints a normal install
summary, and the binary still does not resolve.** To an agent this reads as "tbd is broken"
rather than "PATH is wrong", and the natural next move — run it again — does not help.

This is not hypothetical. It happens whenever Node was unpacked into a directory whose own
`bin/` is not on PATH, which is exactly what a session bootstrap that installs a pinned Node
itself produces. I reproduced it end to end while upgrading tbd 0.4.2 → 0.6.5 in such a
container: install reported success, `tbd` was not found.

**The check is silent when healthy.** It describes the surrounding environment, not repository
health, so a healthy machine adds no line to `tbd doctor` at all:

```
⚠ npm global bin - not on PATH (/tmp/off-path-prefix/bin)
    `npm install -g get-tbd@latest` will succeed but leave tbd unresolvable. Add
    /tmp/off-path-prefix/bin to PATH, or point npm at a directory already on it:
    npm config set prefix <dir>
```

That matches the quiet-when-no-action-needed output contract in
`plan-2026-08-14-github-cli-session-readiness.md`, and it is also what keeps the output stable:
the reported path differs per environment, and the finding flips to a warning *exactly* on the
machines the check exists for — so a golden capture that listed it when healthy would have broken
the suite for the contributors most likely to hit the real problem. The `Doctor full output`
golden in `cli-orientation-golden.tryscript.md` is therefore **unchanged**, and its passing is the
proof that the healthy path stays silent.

Design notes:

- **Warn, not error.** Nothing about the repository is unhealthy, and a tbd installed by other
  means keeps working.
- **A missing npm is "nothing to check", not a failure** — there is then no global install
  location to misconfigure. The session launcher already handles a missing package runner.
- **Diagnose, never mutate.** No prefix is written, no PATH is edited, no wrapper is installed.
  This follows the dispositions already recorded in
  `plan-2026-08-14-github-cli-session-readiness.md`, which drop global-install mutation and the
  `~/.local/bin` wrapper.
- **Scoped to npm** on purpose: `npm install -g` is the instruction tbd actually gives. pnpm and
  Bun have their own global bin resolution.
- Pure logic is an exported `classifyNpmGlobalBin`, mirroring `classifyRemoteSyncHealth`, so the
  finding is unit-testable without spawning npm. Path primitives live in
  `src/lib/npm-global-bin.ts`.
- **Path semantics come from the platform argument, not the host.** `node:path`'s bare
  `delimiter` and `resolve` follow the running platform, so a first pass split a POSIX PATH on
  `;` when run on Windows. `posix` / `win32` are now selected explicitly, and the behavior is
  pinned in tests, so the cross-platform matrix gets the same answers from every runner.

## 2. `docs`: the `agent-session-bootstrap` guideline

The general principle behind the downstream work, written so other repositories can apply it
without copying anyone's script.

Covers the three-part fit test; a table of the cases where a provisioned image, a platform
lifecycle hook, or an existing version manager is the better answer; nine rules for the script
itself; and the traps that make one fail silently.

Two points worth calling out:

- **"Fully up to date" means "matches the repo's pins", not "newest".** A newer language major
  ships a newer bundled package manager, which an `engines` range plus `engine-strict` turns into
  a hard install failure — and runtime "fetch latest" bypasses the release cool-off entirely.
- **It documents tbd's actual merge behavior as a guarantee downstream repos can rely on.**
  `setup.ts` merges session hooks as `[...filtered, ...hookEntries]`, so a repository's own
  `SessionStart` entry both survives `tbd setup --auto` and keeps its position *ahead* of tbd's
  launcher. That ordering matters, because the launcher needs `npx`. I verified this holds
  across a real regeneration downstream.

It sits **next to** `supply-chain-hardening` rather than inside it: that guideline owns *what you
are allowed to install*, this one owns *when a session may install it at all*. It cites
`ensure-gh-cli.sh` as the reference implementation, since that script is already this pattern
applied to one tool.

Also added to the README "Built-in Engineering Knowledge" table, per `tbd shortcut new-guideline`.

## What stays downstream, and why

metaproc's `devtools/ensure-toolchain.sh` installs Node from `.node-version` and uv from the
`uv.toml` `required-version` floor, with per-platform checksums pinned in the script.

**tbd should not own that.** tbd installs `gh` because tbd itself needs `gh` and can pin one
version for everyone. It has no basis to pin a Node or uv version for a downstream repository,
and a generic "install any pinned tool" engine would need per-repo config, per-platform checksum
tables, and a bump workflow — a large new surface for something each repo does in ~200 lines
against pins it already owns. The CI drift guard that keeps those pins honest is likewise
metaproc's, since it asserts against metaproc's files.

So the split is: **tbd teaches the pattern and fixes the one trap that breaks tbd's own
instruction; each repository owns its own pins.**

## Test Plan

- [x] `pnpm test` — 2256 tests across 150 files pass, including 13 new ones in
      `tests/doctor-npm-global-bin.test.ts`.
- [x] `pnpm lint:check` — typecheck, ESLint at `--max-warnings 0`, and the ESLint config contract.
- [x] `pnpm format:check` — Prettier and the pinned flowmark-rs Markdown check.
- [x] `pnpm build` — the guideline is bundled to `dist/docs/guidelines/agent-session-bootstrap.md`.
- [x] The pre-push hook ran `check`, `build`, and `test` again on the pushed commits; all green.
- [x] `pnpm --filter get-tbd test:coverage` — the gate CI actually runs: 2259 vitest tests plus
      **1101 tryscript assertions**, all passing. (Tryscripts run under `test:coverage`, not
      `test` — see "CI findings" below.)
- [x] **Live, healthy path**: `node packages/tbd/dist/bin.mjs doctor` emits no `npm global bin`
      line at all.
- [x] **Live, failing path**: `npm_config_prefix=/tmp/off-path-prefix ... doctor` →
      `⚠ npm global bin - not on PATH`, with the full suggestion. Both verified in this container.
- [x] **Guideline loads and lists**: `guidelines --list` shows it with its description;
      `guidelines agent-session-bootstrap` renders it.
- [x] **Docs cache manifest is stable**: `.tbd/config.yml` gains exactly one `docs_cache` entry,
      sorted correctly among the guidelines, and re-invoking the CLI does not add more.

Unit tests cover: POSIX vs Windows global-bin resolution; exact PATH match; trailing-separator
equivalence; rejecting a directory that is only a *prefix* of an entry (`/usr/local/bin` vs
`/usr/local/bin-other`); absent, empty, and empty-segment PATH values; Windows `;` splitting and
case-insensitivity against POSIX `:` splitting and case-sensitivity, from any host; and all three
classifier outcomes.

## CI findings, and what they changed

Two failures surfaced on the cross-platform matrix that a single-host run could not, and both are
fixed in this branch:

- **Windows (`31aa06e8`, 5 test failures).** `node:path`'s bare `delimiter` and `resolve` follow
  the *running* platform, so `isDirOnPath` split a POSIX PATH on `;` and `resolve` prefixed a
  drive letter. The functions took a `platform` argument but did not honor it. Fixed in
  `72297d6f` by selecting `posix`/`win32` explicitly; the Windows job passed on that commit.
- **`Coverage & Lint` (tryscript goldens).** `pnpm test` runs vitest only — **tryscripts run
  under `test:coverage`** — so a local `pnpm test` cannot catch a golden-output change. Three
  captures broke: two in `cli-doc-output` that pin the alphabetically first guideline, and
  `Doctor full output`. Fixed in `4f0113e3`: the two guideline captures are updated, and the
  doctor finding became silent-when-healthy, which is the better design independent of the test.

## Notes

- **No beads filed.** `tbd sync` would push to the shared `tbd-sync` branch, and I did not want an
  unmerged contribution mutating project tracking state. Happy to file them if you'd prefer.
- The doctor fix (`87910584`, `72297d6f`, `4f0113e3`) and the guideline (`31aa06e8`) are
  separable, though `4f0113e3` also carries the `cli-doc-output` golden update the guideline
  requires.

https://claude.ai/code/session_01XcBhCFuRcFbFZxyW8mWQcr

---

> [!NOTE]
> **Low Risk**
> Advisory doctor check plus documentation; it does not mutate PATH or npm config. Cross-platform PATH matching is the main implementation risk and is covered by unit tests.
> 
> **Overview**
> **`tbd doctor` now names a silent install failure:** `npm install -g get-tbd@latest` can exit 0 while the binary stays unresolvable if npm’s global bin is not on `PATH`. The new check warns with the missing directory and how to fix it (`PATH` or `npm config set prefix`). It is silent when healthy or when npm is missing, so healthy machines do not change doctor output.
> 
> Path logic lives in `npm-global-bin.ts` with platform-explicit POSIX/Windows semantics (including Windows `cmd.exe` for `npm.cmd`). Classification is unit-tested; doctor goldens tolerate the optional advisory.
> 
> Separately adds the **`agent-session-bootstrap`** guideline: when a repo should install its own pinned toolchain at session start (uncontrolled agent containers), when an image or version manager is better, and the checksum/`PATH`/pin-drift rules. Wired into docs cache, README, and skill listings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0e3dbf22385348f4f2f24e76d99adc08e5a034c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/skills/tbd/SKILL.md
+++ b/.agents/skills/tbd/SKILL.md
@@ -401,6 +401,7 @@ Load the **General engineering** group first, then the language or framework gro
 
 | Name | Description |
 | --- | --- |
+| agent-session-bootstrap | When and how to make a repository install its own pinned toolchain at agent session start, for repos whose agents run in containers they do not control. Covers the fit test, the alternatives that are usually better, the install rules a bootstrap must follow, and the PATH and pin-drift traps that make one fail silently. Use when an agent session starts without the tools the repo requires, when writing or reviewing a SessionStart hook, or when deciding between a session hook and a provisioned image. |
 | cli-agent-skill-patterns | A concise decision guide for portable skills, CLI-backed skills, safe bundle installation, and agent integration |
 | common-doc-guidelines | Common cross-project standards for writing and organizing docs, code comments, and text files—how to organize, structure, write, and format documents, plus the guideline footer convention. Downstream of github.com/jlevy/practical-prose. Use whenever writing or editing any documentation, README, guideline, or design doc. |
 | electrobun-app-development-patterns | Building desktop apps with Electrobun—runtime and process model, typed RPC, project layout, packaging and the delta updater, plus an evidence-based maturity and security assessment |
--- a/.claude/skills/tbd/SKILL.md
+++ b/.claude/skills/tbd/SKILL.md
@@ -401,6 +401,7 @@ Load the **General engineering** group first, then the language or framework gro
 
 | Name | Description |
 | --- | --- |
+| agent-session-bootstrap | When and how to make a repository install its own pinned toolchain at agent session start, for repos whose agents run in containers they do not control. Covers the fit test, the alternatives that are usually better, the install rules a bootstrap must follow, and the PATH and pin-drift traps that make one fail silently. Use when an agent session starts without the tools the repo requires, when writing or reviewing a SessionStart hook, or when deciding between a session hook and a provisioned image. |
 | cli-agent-skill-patterns | A concise decision guide for portable skills, CLI-backed skills, safe bundle installation, and agent integration |
 | common-doc-guidelines | Common cross-project standards for writing and organizing docs, code comments, and text files—how to organize, structure, write, and format documents, plus the guideline footer convention. Downstream of github.com/jlevy/practical-prose. Use whenever writing or editing any documentation, README, guideline, or design doc. |
 | electrobun-app-development-patterns | Building desktop apps with Electrobun—runtime and process model, typed RPC, project layout, packaging and the delta updater, plus an evidence-based maturity and security assessment |
--- a/.tbd/config.yml
+++ b/.tbd/config.yml
@@ -101,6 +101,7 @@ docs_cache:
     shortcuts/standard/update-specs-status.md: internal:shortcuts/standard/update-specs-status.md
     shortcuts/standard/watch-beads.md: internal:shortcuts/standard/watch-beads.md
     shortcuts/standard/welcome-user.md: internal:shortcuts/standard/welcome-user.md
+    guidelines/agent-session-bootstrap.md: internal:guidelines/agent-session-bootstrap.md
     guidelines/backward-compatibility-rules.md: internal:guidelines/backward-compatibility-rules.md
     guidelines/bun-monorepo-patterns.md: internal:guidelines/bun-monorepo-patterns.md
     guidelines/cli-agent-skill-patterns.md: internal:guidelines/cli-agent-skill-patterns.md
--- a/README.md
+++ b/README.md
@@ -297,6 +297,7 @@ opinionated rules with concrete examples, built from months of heavy agentic cod
 | [python-rules](packages/tbd/docs/guidelines/python-rules.md) | Type hints, docstrings, exception handling, resource management |
 | [python-cli-patterns](packages/tbd/docs/guidelines/python-cli-patterns.md) | Modern Python CLI stack: uv, Typer, Rich, Ruff, BasedPyright |
 | [backward-compatibility-rules](packages/tbd/docs/guidelines/backward-compatibility-rules.md) | Compatibility across code, APIs, file formats, and database schemas |
+| [agent-session-bootstrap](packages/tbd/docs/guidelines/agent-session-bootstrap.md) | Installing a repo’s pinned toolchain at session start when agents run in containers you don’t control |
 | [convex-rules](packages/tbd/docs/guidelines/convex-rules.md) | Convex function syntax, schema design, queries, mutations |
 | [convex-limits-best-practices](packages/tbd/docs/guidelines/convex-limits-best-practices.md) | Convex platform limits, workarounds, performance tuning |
 
--- a/docs/project/specs/active/plan-2026-08-18-actor-axis-and-identity.md
+++ b/docs/project/specs/active/plan-2026-08-18-actor-axis-and-identity.md
@@ -1,0 +1,494 @@
+---
+title: Plan Spec
+description: Separate who is accountable from who is executing, so an agent can hold a bead without displacing the human who owns it, and resolve actors without hand-maintained identity tables — humans bind per provider to ids discovered from the tracker's own directory, agents carry the identity tbd already mints
+author: Joshua Levy (github.com/jlevy) with LLM assistance
+---
+# Feature: Actor Axis and Identity Mapping
+
+**Date:** 2026-08-18
+
+**Author:** Joshua Levy (github.com/jlevy) with LLM assistance
+
+**Status:** Draft
+
+**Design discussion:** [#246](https://github.com/jlevy/tbd/issues/246) — the provider
+comparison and the measured evidence are argued there and are not re-argued here.
+One part of its recommendation is superseded: the kind-bearing `user_map` table it
+proposed is replaced by the two-namespace model below, which needs no identity table at
+all.
+
+**Sibling:**
+[plan-2026-08-18-tracker-state-model-and-linear-mapping.md](./plan-2026-08-18-tracker-state-model-and-linear-mapping.md)
+(epic `tbd-og20`, [#244](https://github.com/jlevy/tbd/issues/244)) adds `resolution`,
+`hold`, and `started_at`, resolves Linear states by name, and owns the slot vocabulary
+and board projection.
+This spec is the second axis of the same model; its identity-binding phase reuses that
+resolver-and-ask machinery, while its first phase is independent.
+It does not restate the state design.
+
+## Overview
+
+tbd has one actor field, `assignee: z.string()`. Linear has two — `assignee`
+(accountable) and `delegate` (acting) — and its entire agent platform hangs off the
+second. The collapse is silent, and in an agent-driven repository it makes the field
+unusable in either direction.
+
+This adds `delegate` beside `assignee` and replaces identity *configuration* with
+identity *resolution*:
+
+- **Humans** resolve against the tracker’s own member directory and bind by provider
+  user id, one binding per provider since a Linear UUID and a GitHub login are different
+  identifiers. Adding a person to the team requires no tbd config change.
+- **Agents** carry the identity tbd already mints (`agid-{ulid}` plus a friendly name,
+  from `tbd start` / `tbd whoami`). Agents come and go per session; nothing about them
+  is registered anywhere, and they reach a shared tracker only through one explicit,
+  narrow binding for installed Linear agents.
+
+Nobody maintains an alias table, and identities that must stay local stay local by
+construction.
+
+## Goals
+
+- Record that a human is accountable while an agent executes, which is the normal case
+  in an agent-driven repository and is currently inexpressible.
+- Keep agent identities out of a shared tracker **by construction**, not by remembering
+  to omit them from a config table.
+- Add a person without touching config: resolve against the workspace directory, bind by
+  provider user id, survive renames.
+- Keep identity per provider, since a person’s Linear UUID and GitHub login are
+  different identifiers, without making the bead field provider-specific.
+- Push a delegate to Linear when — and only when — it names an installed Linear agent
+  (an app user).
+- Land `tbd start` claims on the acting axis instead of overwriting the accountable one.
+- Report a field that a flow rule excluded, instead of reporting success for a write
+  that went nowhere.
+
+## Non-Goals
+
+- **Agent presence.** Which session is touching a bead right now is a TTL lease in a
+  database, not a bead field: heartbeat cadence is seconds, `extensions` merges as
+  whole-object last-writer-wins, and sync-branch commits are the wrong granularity.
+  `delegate` is the durable, low-frequency fact of who a unit of work is assigned to.
+- **Registering tbd as a Linear agent** (agent sessions, ACKs, typed activity streams).
+  Setting `delegateId` is a plain field write that happens to create an AgentSession on
+  Linear’s side; nothing here requires tbd to become an agent.
+- **GitHub adapter work.** The model is chosen to fit it; nothing here implements it.
+- Re-specifying `resolution`, `hold`, slots, or the board projection.
+  That is the sibling spec.
+
+## Background
+
+Found during the same first-time Linear rollout that produced #244 and #245: ~900 open
+beads, ~105 mirrored epics, nearly all execution done by coding agents, and `assignee`
+in use on **zero** beads.
+Not neglect — there was nothing useful to write.
+The human is the same on every bead, so the field carries no information; the agent was
+not expressible at all.
+
+### What tbd already has
+
+tbd already mints agent identity, and the design below reuses it rather than inventing a
+parallel one:
+
+- `tbd whoami --ensure-id` mints `agid-{ulid}` once per working directory at session
+  start and stores it in machine-local state (`.tbd/state.yml`, gitignored — a committed
+  value would be wrong for every other checkout).
+- A friendly name is joined onto the id, never concatenated into it; renaming an agent
+  does not change who it is.
+- Resolution is total and layered: `--as` beats `$TBD_AGENT` beats session state beats
+  the derived `<harness>@<host>` fallback (`lib/agent-identity.ts`).
+- The identity research
+  ([research-2026-08-14-agent-and-session-identity.md](../../research/current/research-2026-08-14-agent-and-session-identity.md))
+  recommends durable per-agent records at `agents/agid-{ulid}.yml` on the sync branch,
+  one file per identity — not a flat union-merged map, whose duplicate-key hazard §5.4
+  of that document records.
+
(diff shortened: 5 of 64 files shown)


jlevy (commented):
# Senior engineering review — PR #248

## Scope

Reviewed `fe55a36..4f0113e` (7 files, +370/−3) against `main` at `050b5a8`:
the `npm global bin` doctor check (`src/lib/npm-global-bin.ts`,
`src/cli/commands/doctor.ts`, `tests/doctor-npm-global-bin.test.ts`), the
`agent-session-bootstrap` guideline, and the README / `.tbd/config.yml` /
`cli-doc-output` golden updates.

Verification done, not just read: built the branch, ran `tbd doctor` on both the healthy
and the off-PATH path, ran `setup --auto` from the branch's own build to diff the
generated surfaces, ran the orientation golden under an off-PATH npm prefix, and checked
the `setup.ts` hook-merge claim the guideline makes.

## Summary and verdict

**Verdict: approve with changes — two of them worth fixing before merge.**

The diagnosis behind this PR is right and well argued: `npm install -g get-tbd@latest`
exiting 0 while leaving `tbd` unresolvable is a genuinely bad failure for an agent, and
naming it in `doctor` is the correct-sized fix. Warn-not-error, diagnose-never-mutate,
silent-when-healthy, and the pure/testable `classifyNpmGlobalBin` split are all the right
calls. The guideline is accurate — I checked its load-bearing claim about `setup --auto`
against the source and it holds. Commit hygiene and the write-up are excellent.

Two things undercut the change as shipped. The check **never runs on Windows** (R1), which
is the opposite of what commit `72297d6f` set out to guarantee, and the PR leaves the
repo's own generated skill files stale, so the new guideline is **invisible on the surface
agents actually read** (R3). Both are cheap to fix.

The PR also currently **conflicts with `main`** and cannot be merged as-is.

## Findings

### R1 — High — `readNpmGlobalPrefix` cannot run npm on Windows, so the whole check is a permanent no-op there

`packages/tbd/src/lib/npm-global-bin.ts:78`

```ts
const { stdout } = await execFileAsync('npm', ['prefix', '-g'], { … });
```

On Windows `npm` is `npm.cmd`. Node's documentation is explicit:

> On Windows, however, `.bat` and `.cmd` files are not executable on their own without a
> terminal, and therefore cannot be launched using `child_process.execFile()`.

So the call throws `ENOENT`, the bare `catch` maps it to `null`, `classifyNpmGlobalBin`
returns `status: 'ok'` with `'npm not available'`, and `doctor` then suppresses the line.
The result is a silent, permanent no-op on the one platform where the global-bin layout
actually differs — and `npmGlobalBinDir`'s `win32` branch is unreachable at runtime.

The green Windows CI job is not evidence against this: `doctor-npm-global-bin.test.ts`
only exercises the pure classifier and never spawns npm, so nothing in the suite touches
the spawn path on any platform.

This directly contradicts the intent of `72297d6f` ("the cross-platform matrix gets the
same answers from every runner") — the path semantics were made platform-correct, but the
process that feeds them never starts.

**Fix:** select the invocation from the platform, the same way the path helpers already
select their semantics:

```ts
export function npmPrefixCommand(platform: NodeJS.Platform): { file: string; args: string[] } {
  // Windows npm is npm.cmd, which CreateProcess cannot launch directly.
  return platform === 'win32'
    ? { file: 'cmd.exe', args: ['/d', '/s', '/c', 'npm', 'prefix', '-g'] }
    : { file: 'npm', args: ['prefix', '-g'] };
}
```

`src/cli/web/server.ts:363-364` already uses exactly this `cmd.exe /d /s /c` form, so it
is the in-repo precedent. Prefer it over `shell: true`, which Node deprecates for
argument-taking spawns (DEP0190). Exporting the helper makes the Windows branch unit
testable from a Linux runner, which is the same trick the rest of this module uses.

### R2 — Medium — the doctor golden is now environment-dependent, and breaks on exactly the machines this feature exists for

`packages/tbd/tests/cli-orientation-golden.tryscript.md:107-116`

The `HEALTH CHECKS` block is a literal capture with no wildcard, so when the finding fires
it inserts two unexpected lines and `Doctor full output` fails. Reproduced:

```
$ npm_config_prefix=/tmp/off-path-prefix pnpm exec tryscript run tests/cli-orientation-golden.tryscript.md
+⚠ npm global bin - not on PATH (/tmp/off-path-prefix/bin)
+    `npm install -g get-tbd@latest` will succeed but leave tbd unresolvable. …
2 passed, 1 failed
```

The PR body reasons that a golden listing the finding when healthy "would have broken the
suite for the contributors most likely to hit the real problem." That reasoning is right,
but silence-when-healthy only closes half of it: those same contributors still get a
failing suite, now via the warn branch instead of the ok branch. And because the reported
path is machine-specific, no capture can pin the warning line either — the only stable
answer is a capture that tolerates its absence *or* presence.

**Fix:** insert `...` (tryscript's zero-or-more-lines wildcard) after
`✓ Git version - [GIT_VERSION]`, with a comment saying why. Every other line stays pinned;
only that one optional environment advisory becomes tolerated.

I tried the tighter alternatives first and they do not work here, so they are not worth
your time: a `PATH` export in `before:` does not survive into the test commands (each runs
in a fresh shell), and `env:` does not expand `$TRYSCRIPT_PACKAGE_ROOT`, so a relative
`npm_config_prefix` resolves against the random sandbox cwd and cannot be matched by a
static `path:` entry.

### R3 — Medium — the generated skill files were not regenerated, so the new guideline is invisible where agents look for it

`.claude/skills/tbd/SKILL.md:400`, `.agents/skills/tbd/SKILL.md`

Both files are tracked in git, are generated by `tbd setup`, and carry the
`## Available Guidelines` directory that agents read. The PR updated the README table and
the `.tbd/config.yml` docs_cache manifest, but not these. Verified by building this branch
and running its own binary:

```
$ node packages/tbd/dist/bin.mjs setup --auto
$ git diff --stat
 .agents/skills/tbd/SKILL.md | 1 +
 .claude/skills/tbd/SKILL.md | 1 +
```

Both gain exactly one row under **Docs, process & tooling** (the catch-all group in
`src/file/doc-cache.ts:452`, which is where `agent-session-bootstrap` lands — it matches
none of the named groups). Until that lands, `tbd guidelines agent-session-bootstrap`
works but nothing advertises it on the primary discovery surface, which is step 9 of
`tbd shortcut new-guideline`.

Nothing in the suite catches this, which is why CI is green — worth a follow-up
independent of this PR.

**Fix:** run `tbd setup --auto` with the local build and commit both files. Revert the
`tbd_version` / `tbd_upgrades` churn a dev build writes into `.tbd/config.yml`; the
docs_cache entry the PR already added by hand is correct and correctly sorted.

### R4 — Low — `isDirOnPath` does not handle quoted Windows PATH entries

`packages/tbd/src/lib/npm-global-bin.ts:60-72`

Windows PATH entries containing spaces are commonly quoted
(`"C:\Program Files\nodejs"`). `win32.resolve` keeps the quotes, so the entry never
matches and doctor emits a spurious warning telling the user to add a directory that is
already there. A false positive here is worse than a miss, because the suggestion is
actively wrong.

**Fix:** strip surrounding double quotes from each entry before resolving, and pin it with
a test alongside the existing Windows cases.

### R5 — Low — the "pure with respect to platform" claim holds only for absolute inputs

`packages/tbd/src/lib/npm-global-bin.ts:26-31`

`posix.resolve` and `win32.resolve` both fall back to `process.cwd()`, so a relative PATH
entry (`.`, `node_modules/.bin`) still resolves against the *host* cwd, mixing separator
conventions. Harmless in practice — npm prefixes and PATH entries are absolute — but the
doc comment's "keeps them pure and testable from any host" overstates it.

**Fix:** name the absolute-input precondition in the comment.

## Suggestions (non-blocking)

- **S1 — the spawn path and the suppression predicate have no direct test.** "Silent when
  healthy" is currently proven only indirectly, by an unchanged golden on a runner that
  happens to be healthy — the same coupling as R2. Fixing R1 with an exported
  `npmPrefixCommand` gives you a direct assertion for free.
- **S2 — `doctor` now spawns npm on every run** (~180 ms measured in this container; more
  on a cold Windows or macOS start). Fine for a diagnostic command, noted for awareness.

Checked and confirmed benign — please leave these alone:

- **`warn` rather than `error`, and exit 0.** Correct, and exactly what the contract at
  `doctor.ts:620-627` documents.
- **A warn flipping `healthy: false` and the "Issues found" summary.** Consistent with
  existing environment-scoped checks — `checkGitVersion` returns `error` for an old git.
  Not a regression introduced here.
- **`classifyNpmGlobalBin` exported from the command module rather than `lib/`.**
  Deliberately mirrors `classifyRemoteSyncHealth`.
- **The guideline's claim that `setup --auto` merges session hooks and keeps a
  repository's own entry ahead of tbd's launcher.** Verified at `setup.ts:980`:
  `mergedHooks[hookType] = [...filtered, ...hookEntries]`.
- **Scoping the check to npm.** Right call; the warning text is accurate regardless of how
  tbd itself was installed.
- **The README row and the `.tbd/config.yml` docs_cache entry.** Both correct.

## Merge state

**`MERGEABLE: CONFLICTING` / `mergeStateStatus: DIRTY`.** One conflict, in
`packages/tbd/src/cli/commands/doctor.ts:45` — an import-block collision where both sides
added imports after `lib/ids.js` (this branch adds `npm-global-bin.js`; `main` added
`agent-identity.js` and `linear/mapping.js`). Purely mechanical: keep all three.

## CI status at review time

All green on head `4f0113e`: Benchmark, Coverage & Lint, Test on ubuntu (22.12.0 and 24),
macOS 24, Windows 24, DeepSource Secrets. Cursor Bugbot skipped (usage limit), so no AI
review ran on this PR. DeepSource static analysis graded A with no inline issues.

Note that green Windows CI does not cover R1 — no test in the suite spawns npm.

---

jlevy:
Addressed the review findings in `a0e3dbf2`:

- **R1: fixed** — `npmPrefixCommand(platform)` routes the prefix read through `cmd.exe /d /s /c` on win32, using the `openWebBrowser` form you pointed at rather than `shell: true`. Exported, so both branches are asserted from the Linux runner. `readNpmGlobalPrefix` now takes a platform argument defaulting to `process.platform`. Your reading of the failure chain was right end to end — ENOENT → bare catch → `null` → `'ok'` → suppressed — and the `win32` branch of `npmGlobalBinDir` was indeed dead at runtime.
- **R2: fixed** — `...` after `✓ Git version - [GIT_VERSION]`, with a note above the block explaining why that one line is tolerated. Verified in both directions: the golden passes on a healthy host and under `npm_config_prefix=/tmp/off-path-prefix`. I had only closed the `ok` half; you were right that the `warn` half left the same contributors with a red suite.
- **R3: fixed** — regenerated both SKILL.md files with the branch's own build; each gains the one row under **Docs, process & tooling**, as you predicted. Reverted the `tbd_version` / `tbd_upgrades` / `tbd_fallback_version` churn the dev build wrote into `.tbd/config.yml`.
- **R4: fixed** — surrounding double quotes are stripped from win32 PATH entries before resolving, with tests for both the Windows match and the POSIX case where quotes are literal filename characters.
- **R5: fixed** — the comment now names the absolute-input precondition and why it does not arise in practice.
- **S1: applied** — the suppression rule is now `shouldReportNpmGlobalBin`, called from the check site and tested directly, including the errored-check case. No longer proven only by a golden on a healthy runner.
- **S2: acknowledged, no change** — `doctor` is a diagnostic command and the probe is bounded at 5s. Worth revisiting only if `doctor` ever moves onto a hot path.

Also merged `origin/main`, resolving the `doctor.ts` import collision by keeping all three imports as you described. The branch is no longer conflicting.

Verified locally with `pnpm --filter get-tbd test:coverage` — the gate CI runs, and the one that catches tryscripts: 2408 vitest tests and 1101 tryscript assertions passing, plus `lint:check`.

One thing worth flagging beyond the review: R3 is invisible to the suite, so nothing stops the generated skill surfaces from going stale again on the next guideline. That is a gap in the drift coverage rather than a defect in this PR, and I have not widened this PR to close it — happy to open a follow-up for a test that asserts `setup --auto` leaves the tracked surfaces byte-identical.

---

PR #249, opened 2026-08-19
feat: resolve .env from the main worktree so credentials work in linked worktrees

jlevy:
## Summary

In a linked git worktree, tbd reported `LINEAR_API_KEY not found` even when the key sat in the repository's `.env`. `.env` is gitignored, and gitignored files do not propagate to worktrees, so a fully configured repository looked unconfigured from any worktree of it.

Credential resolution gains a third layer after the process environment and this working tree's `.env`: the main worktree's `.env`, resolved through git and verified before use. Status now names the exact file a credential came from.

Reproduced live in this repository: a worktree of `jlevy/tbd` reported the key missing while `/Users/levy/wrk/github/tbd/.env` held a working one.

## Changes

- **`lib/paths.ts`** — `resolveMainWorktree(baseDir)`, built on the existing `resolveGitCommonDir`. Returns undefined rather than throwing outside a repository or when `git` is unavailable.
- **`integrations/core/credentials.ts`** — layer 3 of the resolution order, and `ResolvedCredential.path` carrying the file a `dotenv` value came from.
- **`integrations/core/status.ts`** — the credential finding names its source path, and the `.env` safety finding reports on the file actually read rather than always the current directory.
- **`docs/shortcuts/standard/setup-linear.md`** — joining from a worktree needs no per-worktree key.
- **Spec** — `plan-2026-08-19-worktree-env-credential-resolution.md`.

### Why the obvious mechanism needed a guard

The epic proposed `dirname "$(git rev-parse --path-format=absolute --git-common-dir)"`, on the grounds that `--git-common-dir` cannot leave the repository. Measured, it can:

| Repository shape | `dirname(--git-common-dir)` | Correct? |
| --- | --- | --- |
| Normal repo, main checkout | the checkout | yes |
| Normal repo, linked worktree outside it | the main checkout | yes |
| `--separate-git-dir`, main checkout | the git dir's parent | **no** |
| `--separate-git-dir`, linked worktree | the git dir's parent | **no** |

Under `--separate-git-dir` the dirname lands on an unrelated sibling directory, which is exactly the cross-project `.env` read the approach was chosen to rule out. `git worktree list --porcelain` shares the flaw: under that shape its first entry reports the git dir rather than a checkout.

So the candidate is verified rather than trusted. Asked from inside it, git must agree both that it is a worktree root and that it belongs to the repository we started from. That costs a linked worktree of a `--separate-git-dir` repository its fallback, which is the right trade; its main checkout is unaffected because layer 2 already covers it.

Directory-walking was rejected for a separate reason recorded in the epic: five of six real worktrees measured on one machine lived outside their checkout, so a walk-up would not have reached the repository at all.

### Why the safety finding moved

`envFileFinding` asked only about the current working tree. Once a credential can come from elsewhere, vouching for a local file nobody read while the one in use goes unexamined is how a committed key stays invisible. It now follows the file that answered.

## Test Plan

- [x] **Quality gate** — `pnpm run ci:quality`: prettier, flowmark-rs, `tsc --noEmit` on both tsconfigs, `eslint --max-warnings 0`, config-contract probe.
- [x] **Build** — `pnpm build:check`.
- [x] **Full suite** — `pnpm test`: **2259 tests, 150 files, 0 failures**, run through the pre-push gate.
- [x] **Golden CLI suite** — `tryscript run 'tests/**/*.tryscript.md'`: **1101 passed**.
- [x] **New coverage** — 13 tests in `tests/worktree-credentials.test.ts`, over real `git worktree` checkouts created outside the main one rather than temp-directory stand-ins, since the whole question is what git reports about a directory:
  - main checkout resolves from a linked worktree outside it, and from the main checkout itself
  - key only in the main checkout resolves, and status names that path
  - worktree-local `.env` overrides the main one; an exported variable overrides both
  - nothing found when neither worktree holds a key
  - `--separate-git-dir` rejects the candidate and reads no `.env` outside the repository
  - outside a git repository, behavior unchanged and no new error
  - status reports `.env` safety for the file actually read, and says nothing extra when the local file answered
- [x] **Verified live against the real workspace**, from a worktree with no `.env` of its own and `LINEAR_API_KEY` explicitly unset, so layer 3 is what answers:

  ```
  ✓ .env: present and gitignored (/Users/levy/wrk/github/tbd/.env)

  linear:
    ✓ enabled: yes
    ✓ target: OS
    ✓ credential: ********hGDg from /Users/levy/wrk/github/tbd/.env
    ✓ reachable: Joshua Levy @ Finterm (team OS)
  ```

  The same command on `main` reports `✗ credential: LINEAR_API_KEY not found` and `.env: not present`, which is the bug.
- [x] **CI green** on macOS, Ubuntu (Node 22.12 and 24), and Windows, plus Coverage & Lint, Benchmark, and DeepSource.
- [x] **Hooks respected** — pre-commit and the full pre-push gate on every commit; no `--no-verify`.

### A note on flaky runs during development

Intermediate full-suite runs on a saturated machine (load average peaked above 60) produced varying failures: 5s/30s timeouts in `worktree-health`, `rescue-divergence`, `golden-output`, `git-remote`, a `performance` assertion at 604ms against a 500ms budget, and in tryscript a data-sync lock losing ownership plus `spawn /bin/sh ENOENT`. None of those files reference the changed code, each passed in isolation, and the final gated run on a settled machine was green across all 2259 tests. Recording it here so the pattern is recognizable rather than rediscovered.

The secrets analyzer also flagged the test fixtures twice. The placeholders originally carried the `lin_api_` vendor prefix and lived in constants named `*_SECRET`, either of which reads as a hardcoded credential. Since the code under test never inspects the format, they are now plain descriptive placeholders.

## Related Beads

| Bead | |
| --- | --- |
| `tbd-30t7` | epic: resolve `.env` from the main worktree ([OS-350](https://linear.app/finterm-ai/issue/OS-350)) |
| `tbd-mcw6` | main-worktree resolver |
| `tbd-fzm7` | credential fallback |
| `tbd-lxjr` | source reporting |
| `tbd-0tc3` | tests and setup guidance |

All five are closed by this change.

**Not in scope:** `tbd-k0rd` ([OS-351](https://linear.app/finterm-ai/issue/OS-351), assignee clearing) is deliberately excluded. It collides with PR #245 in all five files it would touch, and #245 already claims the `skipPush` generalization it needs. It is handed to that PR's author rather than raced here.

---

> [!NOTE]
> **Medium Risk**
> Touches secret resolution and cross-directory `.env` reads; mitigated by git-validated main-worktree lookup and explicit path reporting, with dedicated tests for separate-git-dir and precedence.
> 
> **Overview**
> Linked git worktrees no longer falsely report missing integration keys when `LINEAR_API_KEY` (and similar) live only in the main checkout’s gitignored `.env`.
> 
> **Credential resolution** adds a third layer after process env and the current tree’s `.env`: the **main worktree’s `.env`**, found via new `resolveMainWorktree()` in `paths.ts`. The candidate from `dirname(git-common-dir)` is validated (worktree root + same repo) so `--separate-git-dir` repos never read a sibling project’s secrets. `ResolvedCredential` now includes an optional **`path`** for dotenv sources.
> 
> **`integration status`** prints the full path for dotenv credentials and runs the `.env` gitignore safety check against the file actually used (and both local + main when credentials come from elsewhere).
> 
> Docs: active spec plus **setup-linear** guidance that worktrees inherit the main checkout key without duplicating it. **13 integration tests** use real `git worktree` fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4d5e107ce89a8f486e726e186990af67bb107e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/project/specs/active/plan-2026-08-19-worktree-env-credential-resolution.md
+++ b/docs/project/specs/active/plan-2026-08-19-worktree-env-credential-resolution.md
@@ -1,0 +1,224 @@
+---
+title: "Worktree .env Credential Resolution"
+description: Resolve integration credentials from the main worktree's .env so a configured repository stays configured when viewed from any linked worktree
+author: Joshua Levy (github.com/jlevy) with Claude assistance
+---
+# Feature: Worktree .env Credential Resolution
+
+**Date:** 2026-08-19
+
+**Author:** Joshua Levy (github.com/jlevy) with Claude assistance
+
+**Status:** Draft
+
+## Overview
+
+In a linked git worktree, tbd reports `LINEAR_API_KEY not found` even when the key sits
+in the repository’s `.env`. `.env` is gitignored, and gitignored files do not propagate
+to worktrees, so a fully configured repository looks unconfigured from any worktree of
+it.
+
+Add the main worktree’s `.env` as a final fallback in credential resolution, resolved
+through git rather than through the filesystem.
+
+## Goals
+
+- A worktree created anywhere on disk resolves the credential its repository already
+  holds.
+- The reported credential source names the file it came from, so a per-worktree override
+  is discoverable and a cross-directory read is never silent.
+- Resolution cannot reach outside the repository, under every repository shape.
+- Outside a git repository, behavior is unchanged and no new error appears.
+
+## Non-Goals
+
+- Writing to the main worktree’s `.env`. The fallback is read-only.
+- Sharing any other gitignored file across worktrees.
+  This covers integration secrets only.
+- Changing CI behavior.
+  GitHub Actions injects secrets as environment variables and does not use worktrees.
+
+## Background
+
+The failure mode is worse than a missing feature, because the diagnosis is misleading.
+`tbd integration status` prints what reads as “Linear is not set up here”, when the
+truth is “the key exists, in the main checkout, and this process did not look there”.
+That sent an operator into the guided first-time-setup path for a repository whose
+Linear integration was already configured and working.
+
+This is now common rather than exotic.
+Agent tooling creates worktrees automatically, and a single machine can carry many per
+repository, created by different tools in different locations.
+
+### Why Directory Walking Does Not Work
+
+Walking up the directory tree looking for `.env`, which is what `python-dotenv`’s
+`find_dotenv()` and `direnv` do, only works when the worktree happens to be nested
+inside the main checkout.
+Measured against one real machine’s worktrees for a single repository, five of six lived
+outside it:
+
+| Worktree location | Walk-up reaches the repo? |
+| --- | --- |
+| `~/wrk/{org}/{repo}/.claude/worktrees/{id}` | yes, nested |
+| `~/.cache/{tool}/{id}` | no |
+| `~/.codex/worktrees/{id}/{repo}` | no, three of these |
+| `~/.codex/worktrees/{id}/{other-repo}` | no |
+
+Directory walking also has a security footgun: past the repository root it can pick up
+an unrelated project’s secrets, and the deeper the walk the likelier that is.
+
+### Why Git’s Answer Needs Validation
+
+Asking git which repository this is beats inferring it from the filesystem:
+
+```bash
+dirname "$(git rev-parse --path-format=absolute --git-common-dir)"
+```
+
+The precedent is git’s own design.
+Config, hooks, and refs live in the common dir because they are repository-scoped rather
+than branch-scoped.
+A gitignored `.env` holding machine-local credentials is exactly that
+class of state: it does not vary by branch, so sharing it across worktrees is correct
+rather than merely convenient.
+
+That expression is right for the ordinary repository shape and wrong for one real
+variant. Under `git init --separate-git-dir`, the common dir is not inside the checkout,
+so its dirname lands on an unrelated sibling directory:
+
+| Repository shape | `dirname(--git-common-dir)` | Correct? |
+| --- | --- | --- |
+| Normal repo, main checkout | the checkout | yes |
+| Normal repo, linked worktree outside it | the main checkout | yes |
+| `--separate-git-dir`, main checkout | the git dir’s parent | no |
+| `--separate-git-dir`, linked worktree | the git dir’s parent | no |
+
+`git worktree list --porcelain` is no better: under the same shape its first entry
+reports the git dir rather than a checkout.
+
+So the candidate must be verified rather than trusted.
+A candidate is the main worktree only when git, asked from inside that candidate, agrees
+on both facts: the candidate is a worktree root, and it belongs to this same repository.
+Both checks together are what make the “cannot leave the repository” property true
+rather than assumed.
+
+## Design
+
+### Resolution Order
+
+Resolution order for `LINEAR_API_KEY` and any other integration secret:
+
+1. The process environment.
+   Always wins. Unchanged.
+2. `./.env` in the current working tree.
+   Present-but-different is a deliberate override, so a worktree can point at different
+   credentials.
+3. `<main-worktree>/.env`, resolved and validated through git.
+   This is the new fallback.
+4. Not found.
+
+Layers 1 and 2 keep their current precedence and behavior.
+
+### Resolving the Main Worktree
+
+`resolveGitCommonDir` in `lib/paths.ts` already asks git for the common directory, with
+the `--path-format=absolute` fallback and realpath normalization handled.
+The new helper builds on it rather than issuing its own `rev-parse`.
+
+Taking the dirname produces a candidate.
+Accept it only when both hold:
+
+- `git -C <candidate> rev-parse --path-format=absolute --show-toplevel` equals the
+  candidate, so the candidate is a worktree root rather than an arbitrary directory.
+- `git -C <candidate> rev-parse --path-format=absolute --git-common-dir` equals the
+  common dir we started from, so it belongs to this repository.
+
+Anything else returns nothing, and layer 3 is skipped.
+Outside a git repository, or when `git` is not on PATH, the helper returns nothing
+rather than raising.
+
+Under `--separate-git-dir` this costs a linked worktree its fallback, which is the right
+trade: the alternative is reading a neighboring project’s `.env`. The main checkout of
+such a repository is unaffected, since layer 2 already covers it.
+
+### Reporting the Source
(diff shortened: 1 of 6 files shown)


jlevy:
# Senior Review: PR #249 — worktree `.env` credential resolution

**Scope:** Full PR diff at `7fa01dc8` (6 files: `lib/paths.ts`, `integrations/core/credentials.ts`, `integrations/core/status.ts`, `setup-linear.md`, the plan spec, and `tests/worktree-credentials.test.ts`; 910 diff lines), reviewed against the repo's general, comment, error-handling, testing, and TypeScript guidelines, plus the spec's own acceptance criteria.

**Summary:** The mechanism is right and the hard part is done well: the main worktree is resolved through git and then *verified* (worktree root + same common dir) rather than trusted, which is what makes the `--separate-git-dir` shape safe, and that case is a test rather than a caveat. Precedence is preserved, the fallback is read-only, failure degrades to `undefined` instead of raising, and the tests run over real `git worktree` checkouts. One real gap in the relocated `.env` safety check should be fixed before merge; the rest is nits.

**Verdict: approve with one Medium fix required (R1); R2–R3 non-blocking.**

---

## Findings

### R1 (Medium, correctness regression): an unignored worktree-local `.env` is no longer flagged when the credential resolves from the main worktree

`integrationStatus` passes the *external* path to `envFileFinding` (`status.ts:229–231`), and `envFileFinding` then checks **only** `dirname(loadedFrom)` (`status.ts:90`). The local working tree's `.env` is never examined in that branch. Before this PR it always was.

Concrete failure: a worktree-local `.env` holds a `GITHUB_TOKEN` but is not gitignored on this branch — exactly the "highest-cost, lowest-visibility failure" this module's own docstring names. The Linear key resolves from the main checkout, so `external` is set, the safety check inspects only the main file, reports `✓ .env: present and gitignored (<main>/.env)`, and the exposed local file is vouched for by omission. The module's stated purpose is to make this state loud; the relocation quietly narrowed it to one file.

**Fix:** check both files when they differ. Run `checkEnvIgnored` on `repoRoot` unconditionally and on `dirname(external)` when external is set; report the more severe finding, always naming the file the finding refers to (the `(path)` suffix should appear on the local file too when the external branch is active, so the two are distinguishable). Add the regression test: worktree-local unignored `.env` without the key + main-resolved credential → `state: 'error'` naming the *local* path.

### R2 (Low, efficiency): `resolveMainWorktree` is unmemoized and runs per provider

`resolveCredential` has 5 call sites and `integrationStatus` loops providers (`status.ts:160`), so a status call with Linear and GitHub configured and no key anywhere spawns the full candidate-validation sequence (3–4 git subprocesses) twice for an identical answer. It also delays the `gh-cli` fallback, which now sits behind the main-worktree miss.

**Fix (pick one):** memoize per `repoRoot` at module level in `paths.ts` (the answer cannot change within a process run), or resolve once in `integrationStatus` and thread it through. Non-blocking: tens of milliseconds on the miss path only.

### R3 (Low, docs drift): the spec ships as `Status: Draft` with all implementation boxes unchecked

`plan-2026-08-19-worktree-env-credential-resolution.md` lands in this same PR with `**Status:** Draft` and four `- [ ]` items, while the beads it names (`tbd-mcw6`, `tbd-fzm7`, `tbd-lxjr`, `tbd-0tc3`) are closed by this PR. The repo's spec-status process (`update-specs-status`) treats exactly this mismatch as the drift to catch.

**Fix:** tick the four boxes and set `Status: Implemented` in this PR, so the spec is born consistent.

## Suggestions (non-blocking)

- `describeSource` falls back to `ENV_FILE_NAME` when `credential.path` is absent (`status.ts:72`), but every `dotenv` result now carries `path`. Fine as type-honest defense; a comment saying the fallback is vestigial would stop a future reader from hunting for the pathless case.
- The `remedy` strings in `envFileFinding` say "Add `.env` to `.gitignore`" without saying *which* checkout when the finding names the main worktree's file. Folding the directory into the remedy would make the R1 fix's output fully self-locating.

## False positives / checked, do not fix

- **Exit-code change from a worktree:** an unignored main-checkout `.env` now fails `tbd integration status` run from any worktree. Deliberate: the risk is repository-wide, and a test pins it.
- **Symlinked paths (macOS `/tmp`):** all comparisons go through `realpath`/`resolve` on both sides; the tests' `realish` helper mirrors this. Verified by the suite on macOS CI.
- **Windows path separators:** `--path-format=absolute` output is normalized via `resolve()` comparison; windows-latest CI is green including the new tests.
- **Write-safety:** `env-file.ts` exposes no write path at all; nothing in the diff writes to any `.env`. The "read-only fallback" claim holds by construction.
- **`maxBuffer` literal in `resolveWorktreeTopLevel`:** matches the inline convention of the adjacent `resolveGitCommonDir`; a shared constant would be a separate cleanup across pre-existing code.

## CI status at review time

All 8 checks passing: Test on macOS/Ubuntu (Node 22.12, 24)/Windows, Coverage & Lint, Benchmark, DeepSource Secrets. Cursor Bugbot skipped (usage limit).

jlevy:
Addressed the review findings in `d4d5e107`:

- **R1: fixed** — `envFileFinding` now evaluates both the local and the external `.env` and reports the more severe, naming the file and its checkout whenever two are in play (an exposed local file outranks the external one, since it is the one fixable from here). The single-file case keeps its historical wording, so existing output and tests are unchanged. Regression test added: an unignored worktree-local `.env` lacking the key, with the credential resolving from the main worktree, reports `error` naming the local path.
- **R2: fixed** — `resolveMainWorktree` is memoized per `baseDir` in `paths.ts`; both providers now share one answer per process. Env-file content is not cached and is reread on every resolution.
- **R3: fixed** — the spec ships as `Status: Implemented` with all four implementation boxes checked.

Both non-blocking suggestions were applied while in the area: `describeSource`'s bare-name fallback is commented as vestigial, and the `envFileFinding` remedies name the checkout (`in <dir>`) whenever two files are in play.

Verification: full suite 2260/2260 through the pre-push gate, golden CLI suite previously 1101/1101 (unaffected paths), CI green on `d4d5e107` across macOS/Ubuntu/Windows, Coverage & Lint, Benchmark, and DeepSource. Tracked as `tbd-monc` with one child per finding, all closed.

PR #257, opened 2026-08-23
Add Rust guideline group and cross-cutting topic routing

jlevy:
## Summary

This change establishes the infrastructure for Rust guidelines in tbd by adding a Rust group to the guideline classification system and introducing a new "Cross-cutting engineering topics" group for language-neutral guidelines. It also adds test coverage to prevent future misrouting of language-specific guidelines.

## Key Changes

- **Added Rust guideline group** to `GUIDELINE_GROUPS` in `doc-cache.ts`, positioned before the catch-all group so Rust-specific documents are correctly categorized
- **Introduced Cross-cutting engineering topics group** for language-neutral guidelines (`ci-and-gates-rules`, `code-review-rules`, `filesystem-rules`, `release-engineering-rules`) that apply across all languages
- **Refactored General engineering group matching** to use explicit name sets (`GENERAL_ENGINEERING_NAMES`) instead of substring matching, preventing language-specific testing rules (e.g., `rust-testing-rules`) from being incorrectly filed under "General engineering"
- **Exported `guidelineGroupFor()` function** to make guideline group assignment directly testable
- **Added comprehensive test suite** (`guideline-groups.test.ts`) that validates:
  - General engineering guidelines are correctly identified
  - Language-specific guidelines land in their respective language groups
  - Language-specific testing rules are not captured by substring matching
  - Cross-cutting topic guidelines are properly routed
  - Fallback behavior for uncategorized guidelines

## Implementation Details

The key insight is that substring matching on `'testing'` or `'tdd'` was silently capturing language-prefixed guidelines like `rust-testing-rules` and filing them under "General engineering," which tells agents to read them for any engineering work regardless of language. The fix uses exact name matching for both general and cross-cutting guidelines, with language-specific matching only on the prefix.

The exported `guidelineGroupFor()` function enables direct testing of routing logic, which was previously only visible in the generated directory output and easy to miss during review.

https://claude.ai/code/session_01PbC8y9fMpRmvf2BLwUQ5m9

PR #258, opened 2026-08-23
Extract cross-cutting engineering guidelines and add the Rust family

jlevy:
Continues the work planned in #257, which this branch builds on directly (its four commits are included here). #257 added the Rust and Cross-cutting groups to `GUIDELINE_GROUPS` and the routing test; this branch writes the documents those groups were wired for, and closes the four language-neutral coverage gaps the plan identified.

Net: **12 new guidelines and 1 new shortcut** (4 cross-cutting + 8 Rust), plus a substantially expanded `general-testing-rules`, a re-measured Rust lint floor with a checked-in reproducer, and two new gates for rules this repo stated but did not enforce.

Two review rounds have landed on this branch. The second one ([`review-2026-08-23-pr258-holistic-engineering-guidelines.md`](docs/project/reviews/review-2026-08-23-pr258-holistic-engineering-guidelines.md)) requested changes on 8 High + 5 Medium findings; all thirteen are addressed, and the sections below reflect the corrected state rather than the original claims.

## Why this shape

The plan's central observation was that most of the Rust suite was never Rust-specific. Three of the seven documents were split rather than copied, so a rule that applies in every language is now written once instead of once per language:

| Playbook document | Neutral core extracted to | Rust document keeps |
| --- | --- | --- |
| `rust-filesystem-rules` (9 of 10 sections neutral) | `filesystem-rules` | `Path`/`OsStr` types, the tempfile sequence, traversal crates, platform metadata |
| `rust-release-rules` (13 of 15 neutral) | `release-engineering-rules` | crates.io publishing, maturin wheels |
| `rust-code-review-rules` (6 of 7 neutral) | `code-review-rules` | unsafe and FFI review |
| `rust-project-setup` (4 of 12 neutral) | `ci-and-gates-rules` | Cargo shape, features, toolchain, rustfmt |

The four cross-cutting documents also absorb the practices this repo and [fdu](https://github.com/jlevy/fdu) enforce in config and document nowhere — the material that was executable but untaught. Among them: the pipeline-exit-status trap (`cmd | grep` reports on `grep`, so a failing `cmd` makes the check pass having read nothing), evidence a machine records about itself and therefore cannot falsify, single-platform lint blindness for `cfg(target_os)` code, generated-file ownership, suppression ratchets that carry a tracker ID, and gate noise as a slow repeal.

## The Rust lint floor: measured, then re-measured

The plan marked its lint table "a design proposal; Phase 3 validates it against a real Rust codebase before it ships," and left one open question: whether `clippy::indexing_slicing` is tolerable codebase-wide.

The first validation used a method that does not work. It split each file's diagnostics at its first `#[cfg(test)]` attribute — but that attribute applies to the *next item*, not to the rest of the file. Reconstructing that method over the raw diagnostics and comparing against a compile-unit classification: **145 of 415 diagnostics (34%) landed in the wrong bucket, in both directions.**

The measurement now asks the compiler instead. Two clippy passes over fdu at `d42d970` (35,081 lines across a library, a CLI, and a PyO3 extension module, already at `pedantic = deny` + `unwrap_used = deny`): `--lib --bins` for code that compiles without `cfg(test)`, `--all-targets` for everything, classified by cargo's per-diagnostic target.

| Candidate lint | Ships | Build scripts | Tests/examples | Verdict |
| --- | ---: | ---: | ---: | --- |
| `let_underscore_future` | 0 | 0 | 0 | Adopt |
| `panic` | 2 | 2 | 35 | Adopt in library code |
| `wildcard_enum_match_arm` | 12 | 0 | 9 | Adopt |
| `disallowed_methods` (fs writes) | 0 | 1 | 82 | Adopt |
| `expect_used` | 32 | 7 | 35 | Defer or ratchet |
| `indexing_slicing` | 79 | 0 | 119 | **Not a floor rule** |

Every nonzero figure moved. No verdict reverses, and two are now shown rather than asserted: half of `panic`'s non-test cost is `build.rs`, and `disallowed_methods` has *zero* shipping sites. The open question is answered — `indexing_slicing` is not a hard deny. The analogy to `noUncheckedIndexedAccess` is inexact: that flag changes an inferred type and is usually satisfied by a bounds check the code already has, whereas `indexing_slicing` demands `.get()` plus real error handling at all 79 sites.

Reproducer: `scripts/measure-rust-lint-cost.mjs`. Method and the full 415-row `lint`/`target`/`file`/`line` mapping: `docs/project/research/current/evidence-2026-08-23-rust-lint-cost.{md,tsv}`. Re-running the checked-in script emits a byte-identical TSV.

The floor is no longer justified by strictness. "Strictest wins" is replaced by six admission criteria — defect class, enforcement reliability, incidence, measured adoption cost, applicability, context cost — plus explicit departure conditions by project shape (MSRV below 1.81, crates whose macros expand to forbidden code, proc macros, projects with no pinned toolchain).

## Rules this repo stated but did not enforce

Two of the corpus's own rules were being broken by the corpus's own repository. Both now have gates.

**Action pinning.** The rule permitted an exact tag where no floating major tag exists, and `ci.yml` used that exception. A tag is a mutable reference whether it is `v4` or `v8.3.2`. The exception is gone, all 15 action references are pinned to commit SHAs with the release tag in a trailing comment, `.github/dependabot.yml` keeps them from freezing, and `scripts/check-action-pins.mjs` is wired into `lint:check` with negative tests for floating tags, exact tags, branches, and short SHAs.

**Context budget.** The always-load group was 2,233 lines / 11,806 words read before any repository code. It is now **909 lines / 4,663 words**: testing, TDD, goldens, compatibility, and commits are routed by changed surface instead. Membership is an explicit set rather than a `general-` filename prefix, so a new document cannot admit itself to what every session reads, and `guideline-budget.test.ts` asserts the ceiling.

## Things found by running the checks rather than reasoning about them

- Two prescribed gate recipes could pass while doing no required work, and both were reproduced failing before being fixed. The lint-policy loop exits zero whether or not it prints complaints, and globs `crates/*` so a member elsewhere is never examined — it now enumerates from `cargo metadata`, accumulates, asserts a non-empty member list, and exits nonzero. Cross-target linting skipped uninstalled targets, so a runner with none reports green after linting nothing — it is now a permissive local target and a strict CI target that asserts its target set first.
- The first corrected measurement run reported 9 diagnostics instead of 415: fdu denies warnings, so the run died in the first crate's build script. A measurement that stops at the first failure measures whichever crate compiled first.
- `import.meta.url === ` `` `file://${process.argv[1]}` `` never matches on Windows, where `argv[1]` is a drive path — the new gate's main block would have silently not run there.
- `clippy.toml`'s `allow-unwrap-in-tests` / `allow-expect-in-tests` cover inline `#[cfg(test)]` items but **not** integration tests under `tests/`, examples, or build scripts.
- `grep -L 'workspace = true' crates/*/Cargo.toml` is a false pass — the string also appears in inherited package fields (`edition.workspace = true`), so it reports clean for a crate with no `[lints]` table at all. Checking for inheritance is also the wrong question: a crate that genuinely cannot inherit (pyo3's macros expand to `unsafe impl` under `unsafe_code = "deny"`) declares its own `[lints]` table and must pass.
- `Path::with_extension`, `ErrorKind::CrossesDevices`, `NamedTempFile::persist` semantics, and the `BufWriter`/`into_inner` replacement sequence were verified by compiling them rather than asserted.

## A defect this work exposed, now guarded

The Cross-cutting group was wired into `doc-cache.ts` and asserted in `guideline-groups.test.ts` before any of its four documents existed. The routing test passed, the generated directory rendered an empty heading, and nothing failed. `guideline-groups.test.ts` now checks both explicit name sets against what is actually bundled — verified to fail when a named document is removed.

## Corrections from the second review

Beyond the lint evidence, the gates, and the budget: the FFI unwind matrix was backwards (a Rust panic escaping `extern "C"` aborts; a *foreign* exception entering Rust is the UB case). Atomic replacement was mandated for every write, which is wrong for appends, exclusive creation, streams, and scratch output — five write contracts are now named, each mapped to its `OpenOptions` spelling. Broken-pipe handling mapped any `BrokenPipe` to exit zero from a boundary helper that cannot tell stdout from a socket; the neutral contract now lives in `error-handling-rules`, and a closed stderr can never lower a nonzero status (the TypeScript family had the same drift). `abi3` was recommended on a `bindings = "bin"` path where it does not apply. Both quick-scan tables assigned default severities to grep patterns, which encourages exactly the pattern-matching review they warn against — each row is now the question that decides the finding. The authoring shortcut validated new documents against the previously published bundle.

## Also

- `rust` added to `DOC_CATEGORIES` and to the contract test's own copy of the vocabulary (which deliberately does not import the constant it validates).
- `review-code` loads `code-review-rules` and routes topic guidelines by changed surface; new `review-code-rust` shortcut mirrors the TypeScript one.
- `typescript-lint-format-rules` §Hooks and Gates and `typescript-rules` §File Operations point at the new documents instead of restating them.
- `general-testing-rules` expanded from six bullets — the addition that matters most is that an absent test looks exactly like a passing one.
- Committed agent surfaces regenerated (they had drifted from their sources); what remains of that bead is a drift check, noted with the trap that a naive `setup --auto && git diff --exit-code` cannot work.

## Not in scope here

Phase 5 (simplifying the playbook to consume tbd's copies) needs a PR against `jlevy/rust-porting-playbook`; this session has read-only access to it. Also deliberately not done, and recorded in the plan: the `topics` / `appliesTo` / load-policy metadata refactor (the routing change delivers the budget without the schema change), the adversarial decision-scenario suite, and per-rule admission records.

## Validation

`pnpm test` — 2427 passed, 160 files. `pnpm ci:quality` clean. CI green on all platforms including Windows. Rust examples compile; both corrected gate recipes were run against positive and negative fixtures and watched failing.

https://claude.ai/code/session_019d5xS7vDcKteJfYTs6fLDo

## Second-pass preservation audit

A preservation-first review at `0c34446b` compared all 54 changed files with `main`, treating pre-existing tbd documents as authoritative. It found R14-R22; all nine are addressed in `a55041a0`. The original `general-testing-rules` core is restored verbatim, with new evidence guidance kept as a separate addition rather than a stylistic rewrite.

Focused corrections make the lint-cost reproducer reject partial failed Cargo runs and use `--locked`; parse action references structurally as YAML with fail-closed tests; correct PyO3 ABI, crate-type, and panic boundaries; define release units; distinguish noisy timing thresholds from controlled regression gates; continue high-risk review after an unrelated gate failure; state Rust ownership contracts precisely; and align neutral and TypeScript filesystem enforcement with write intent.

Standalone review: [`review-2026-08-23-pr258-holistic-engineering-guidelines.md`](docs/project/reviews/review-2026-08-23-pr258-holistic-engineering-guidelines.md), section **Second-Pass Preservation and Precision Review**.

Validation: `pnpm run ci` and the pre-push hook both passed; 161 test files, 2,431 tests. The lower-priority composite-action and Docker-digest extension is tracked separately as `tbd-yk6w`.


jlevy:
_The review below is posted in full from the standalone review doc `docs/project/reviews/review-2026-08-23-pr258-rust-guidelines-extraction.md`, now committed on this PR's branch (merge `4173c22`). The review itself covers commit `078938d`._

# Senior Review: PR #258 — Cross-Cutting Guidelines and the Rust Family

**PR:** https://github.com/jlevy/tbd/pull/258 (branch
`claude/rust-guidelines-extraction-o9x2yy`)

**Reviewed commit:** `078938d`

**Date:** 2026-08-23

**Scope:** All 29 changed files (+3,802/−32): the four cross-cutting guidelines, the
eight Rust guidelines, the `review-code-rust` shortcut, the expanded
`general-testing-rules`, the `doc-cache.ts`/`doc-categories.ts` changes and their tests,
and the generated and registration surfaces (SKILL.md files, `.tbd/config.yml`, README).
Every new document was read in full and compared against the existing corpus
(`typescript-rules`, `typescript-lint-format-rules`, `python-rules`,
`common-doc-guidelines`, the `new-guideline` shortcut) for convention fit.
Factual claims were verified against primary sources where possible: the `tempfile`
crate’s documentation and Windows implementation, the repo’s own `eslint.config.js` and
`scripts/check-eslint-contract.mjs`, and rustc/clippy/cargo semantics.
The full test suite was run locally on the branch: 158 files, 2,415 tests, all passing,
matching the PR’s claim.

**Review calibration:** The audience for these documents is highly capable coding
agents. Content earns its context cost when it is opinionated-and-likely-correct, covers
what is commonly done inconsistently, or teaches the neglected and non-obvious (the
`tryscript`/golden-testing class of material).
Restating what a strong model already produces from its priors is negative value.
Findings below apply that bar.

## Summary and Verdict

**Verdict: approve with fixes.** The design is right, most of the content clears the
high bar, CI is green, and the code changes are clean and well-tested.
One factual error in a shipped guideline (R1) and one verification step that contradicts
its own principle (R2) should be fixed before merge—both are small, localized edits.
The duplication cluster (R3) is the main simplification opportunity and directly serves
the stated goal of not wasting context on capable agents; it can land in this PR or as
an immediate follow-up, but it should land.

What is strong, specifically:

- **The extraction shape is correct.** Neutral core in four shared documents, thin
  language-specific residue, each split document opening with “read the neutral one
  first; this owns only what is Rust-specific.”
  That structure scales to future language families (the plan already notes Python lacks
  filesystem/release documents) and writes each rule once.
  `rust-release-rules`, `rust-code-review-rules`, and `rust-filesystem-rules` execute
  the split best.
- **The measured-adoption-cost table in `rust-lint-format-rules` is the model to
  replicate.** Measuring candidate lints against a real 35k-line codebase, splitting
  production from test sites, and rejecting two commonly proposed rules with numbers
  ("62 production sites in a codebase already at this floor") converts taste into
  evidence. This is exactly the opinionated, non-obvious content this corpus exists
  for—as are the `#[expect]` expiry rationale, the workspace-lints opt-in trap, the
  `allow-*-in-tests` scope limits, and the false-pass `grep -L 'workspace = true'`
  correction.
- **The four cross-cutting documents are dense with material that is genuinely
  neglected**: the pipeline-exit-status trap with a worked example, self-recorded
  evidence, single-platform lint blindness, the lefthook `parallel: false`/`priority`
  interaction, `GIT_DIR` scrubbing at two layers, suppression ratchets with tracker IDs,
  the unpublished-sibling packaging trap, smoke-testing the packaged artifact rather
  than the build output.
  Verified spot-checks all held (see False Positives).
- **The `doc-cache.ts` fix is better than the plan proposed.** Replacing substring
  matches with exact-name sets removes the root cause (any future `<lang>-testing-rules`
  misroute) rather than working around it with group ordering, and `guidelineGroupFor`
  is exported and tested.

## Design Assessment

The alternative to this shape—complete per-language documents with no shared core—was
rightly rejected: it multiplies every neutral rule by the number of language families
and guarantees drift.
The cost of the chosen shape is pointer-chasing across documents, which the **Related**
blocks and “read it first” openers handle adequately.
The strictest-wins, no-menus floor policy fits the agent audience: an agent handed
ranked alternatives takes the cheapest, so a single stated default with named departure
conditions is the correct genre.
No materially better structure was identified.

Where execution lags the design is discipline about which document owns an artifact.
Several rule statements, config blocks, and examples appear verbatim in two or three
documents that are co-loaded by construction (R3). For this corpus, duplication is not
just a maintenance risk; it is a per-session context tax multiplied across every
downstream repo, and the documents’ own `common-doc-guidelines` names avoiding it as a
rule.

On verbosity calibration: the cross-cutting four, `rust-lint-format-rules`,
`rust-release-rules`, `rust-filesystem-rules`, and `rust-code-review-rules` pass the
“would a capable agent already know this” test well.
`rust-rules` and parts of `rust-project-setup` and `rust-cli-rules` carry the most
textbook material (S1 names specific cuts).
This is consistent with the existing family (`python-rules` and `typescript-rules` also
carry basics), so it is a trim-going-forward suggestion, not a blocker.

## Findings

Severity vocabulary per `code-review-rules`: Blocker/High/Medium/Low.

**R1 (High). `rust-filesystem-rules.md:101` states the opposite of `persist`’s actual
Windows behavior.** The document says “**`NamedTempFile::persist` fails rather than
clobbering on Windows**.” Verified against the `tempfile` crate: the documentation says
“If a file exists at the target path, persist will atomically replace it,” and the
Windows implementation calls `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` when
overwriting is requested.
`persist` replaces the destination on every platform—that is what makes it usable as the
atomic-replace primitive the same section recommends.
The claim as written also contradicts the section’s own purpose: if `persist` failed on
existing destinations, it could not implement step 5 of `filesystem-rules`’ replacement
sequence on Windows at all.
An agent believing it would treat `persist` as collision-safe and silently clobber,
inverting `filesystem-rules`’ “never make overwrite the fallback” rule.
**Fix:** Replace the sentence with the accurate distinctions: `persist` atomically
replaces an existing destination on all platforms; on Windows it can fail with a
permission error when the destination is open (no POSIX-style replace-while-open);
`persist_noclobber` is the variant that fails when the destination exists.

**R2 (Medium). `rust-lint-format-rules.md:281-284` — the “confirm the effective lint
level” command changes the effective lint level.** Step 2 of Verifying the Floor runs
`cargo clippy --workspace --all-targets -- -W clippy::pedantic 2>&1 | head`. Trailing
command-line lint flags are passed to rustc after the `[lints]`-derived flags, and the
last flag wins, so `-W clippy::pedantic` demotes the manifest’s `deny` to `warn` for
that run. The command therefore cannot confirm the configured level—it overrides it—in a
section whose stated principle is “confirm the effective lint level, not the config
text.” The deliberate-violation probe in the next sentence is the real check.
**Fix:** Delete the `-W` command and let the step be the probe: introduce a violation
(an `unwrap()` in library code, an undocumented public item), run the normal gate with
no extra flags, and confirm it fails.

**R3 (Medium). Whole sections, config blocks, and examples are duplicated across
co-loaded documents.** The inventory, each with the suggested single owner:

- (a) “Timeouts Record a Measurement”—full section including the identical
  `bridge-merge`/`isWindows` example—appears in both `ci-and-gates-rules.md:281-297` and
  `general-testing-rules.md`. The plan assigned this rule to `general-testing-rules`;
  keep it there and reduce the CI document to one sentence and a pointer.
- (b) `rust-testing-rules` restates three sections of the co-loaded
  `general-testing-rules` with only wording changes and no Rust content: “Keep Tests
  Deterministic” (`rust-testing-rules.md:47-58`), most of “Treat Fixtures as Inputs With
  Provenance” (60-69; the Rust part is `include_str!`/`include_bytes!`), and “Keep
  Ignored and Flaky Tests Actionable” (180-189). The “do not duplicate the same
  assertion at every layer” bullet also now appears in both.
  Cut the neutral restatements; keep the Rust-specific residue.
  This is the one document that does not honor its own “supplements the
  language-agnostic guidance” framing.
- (c) The `filter_map(Result::ok)` bad/good example is verbatim in
  `filesystem-rules.md:166-186` and `rust-filesystem-rules.md:124-144`. It is Rust code
  either way; since `filesystem-rules` needs a concrete illustration for all languages,
  keep it there and have the Rust document add only the
  `entry.io_error().map(|e| e.kind())` refinement and the `filter_entry` pruning note.
- (d) The `disallowed-methods` TOML block is verbatim in
  `rust-lint-format-rules.md:164-172` and `rust-filesystem-rules.md:61-69`. One owner
  (the lint document, which also carries the measured cost), one pointer.
- (e) The `rust-toolchain.toml` block plus the pin-versus-MSRV paragraph appear in both
  `rust-lint-format-rules.md:139-151` and `rust-project-setup.md:122-137`, nearly
  verbatim including the “compiles *and* tests” sentence.
- (f) The verify-gate command block (fmt/clippy/test/doc/deny) appears in
  `rust-lint-format-rules.md:243-250` and `rust-project-setup.md:177-183`.
- (g) The workspace-lints opt-in rule is stated twice within `rust-project-setup` itself
  (`:82-87` and `:162-164`) on top of its full treatment in
  `rust-lint-format-rules.md:107-116`.
- (h) The paired review questions ("did this change make a check unable to fail?"
  / “does this test still assert what its name claims?”) appear in three documents:
  `code-review-rules.md:169-172`, `general-testing-rules.md`, and
  `rust-code-review-rules.md:107-110`. State once in `code-review-rules`; point
  elsewhere.

**Fix:** One owner per artifact plus a pointer, per the split convention the PR itself
establishes. Rough saving is 150-200 lines (~1.5-2k tokens) concentrated in exactly the
document sets that load together, and it removes the drift risk of parallel copies.

**R4 (Medium). `review-code.md` step 5 routes TypeScript and Python but not Rust.** The
general review shortcut (which `review-github-pr` invokes) instructs loading language
rules for TS and Python files and now for topic guidelines, but a Rust diff gets no
language instruction, despite this PR shipping the Rust family and a routing table in
the plan that covers this scenario.
**Fix:** Add a “For Rust files” line to step 5, mirroring the TypeScript one: load
`rust-rules` and `rust-lint-format-rules` (plus `rust-code-review-rules`).

**R5 (Medium). `review-code-typescript` and `review-code-python` do not load
`code-review-rules`, though the plan checks off “Have the `review-code*` shortcuts load
`code-review-rules`.”** Only `review-code` and `review-code-rust` do.
The Rust variant also demonstrates a better structure for a language-focused review
(process + rules + surface routing + gate confirmation) than the older two variants.
**Fix:** Add the `code-review-rules` load line to both older variants (minimum), or
bring them up to the `review-code-rust` shape; alternatively correct the plan checkbox.
The first option is cheap and matches the plan’s intent.

**R6 (Medium). The skill templates that gate discovery never mention Rust, and their
guideline count is stale.** `packages/tbd/docs/install/claude-header.md` lists
TypeScript, Python, and Convex in both the “Use for:” and “Invoke when user mentions:”
trigger lists—an agent hearing “Rust best practices” gets no trigger word.
`skill-baseline.md:12` and `skill-minimal.md:45` still say “25+ engineering guidelines
(TypeScript, Python, …)” while the README on this branch says 40+ (43 are bundled).
The baseline’s User Request → Agent Action table has “Use TypeScript best practices” and
“Use Python best practices” rows but no Rust row, so the primary routing surface never
loads the Rust pair at all.
(Regeneration of the *committed* skill surfaces is separately and deliberately tracked
as `tbd-eidy`; this finding is about the source templates, which that bead does not
cover.) **Fix:** Add Rust to both trigger lists, add a “Use Rust best practices” routing
row that loads `rust-rules rust-lint-format-rules`, and update the counts (or make them
future-proof, e.g. “40+”).

**R7 (Medium). `guideline-groups.test.ts:76-85` cannot catch the next empty-heading
regression—the class it was written for.** The bundled-existence check tests a
hand-copied `named` array.
A future name added to `GENERAL_ENGINEERING_NAMES` or `CROSS_CUTTING_NAMES` in
`doc-cache.ts` without a bundled document reproduces the exact “group names documents
that do not exist” defect the PR description reports, and every test still passes (the
copied list doesn’t contain the new name, and the non-empty-heading check stays
satisfied by the other members).
The deliberate-copy pattern is right for contract vocabularies (`doc-categories.test.ts`
pins an error-message contract); here the set is routing configuration where drift means
silent failure. **Fix:** Export the two sets (or one combined
`EXPLICITLY_GROUPED_GUIDELINES` constant) from `doc-cache.ts` and assert every member is
bundled.

**R8 (Medium).
The seven migrated Rust documents carry two Related lists each; the family
convention is one.** All of `rust-rules`, `rust-project-setup`, `rust-cli-rules`,
`rust-testing-rules`, `rust-filesystem-rules`, `rust-release-rules`, and
`rust-code-review-rules` have both the top `**Related**:` block and a trailing “##
Related Guidelines” section with substantially the same links.
`typescript-rules`, `python-rules`, the four new cross-cutting documents, and the
freshly authored `rust-lint-format-rules` all have the top block only—so the PR is
internally inconsistent as well as duplicative.
The plan’s conversion checklist instructed keeping both ("Keep that section and add the
bolded Related header block"); the instruction is the bug.
**Fix:** Delete the trailing sections (folding any link that appears only there into the
top block) and correct the plan checklist item.

**R9 (Low).
`rust-filesystem-rules.md:89` — the comment describes a writer that is not in
the example.** `staged.flush()?; // flushes the BufWriter, not the file` annotates code
with no `BufWriter`; `NamedTempFile`’s `flush` forwards to `File::flush`, which is a
no-op. The flush-versus-`sync_all` trap is real, but the comment as written teaches a
wrong model of this exact code.
**Fix:** Either wrap the writes in a `BufWriter` so the comment is true, or reword: “a
no-op here, but required the moment a buffered writer wraps the file; `sync_all` is what
reaches the device either way.”

**R10 (Low). Typo in `ci-and-gates-rules.md:81-82`:** “works exactly until the once
somebody forgets” → “until the one time somebody forgets.”

**R11 (Low). Session and plan hygiene.** Beads for work completed in this PR remain open
(`tbd-japx`, `tbd-gr6y`, `tbd-jfs3`, `tbd-3w73`, `tbd-g3b5` at minimum), and the last
bead sync (05:39Z) predates the branch’s final commits.
The plan header says “Phases 1-4 complete in tbd” while Phase 1 leaves two items
unchecked (confirm `rpp-u657`/file the receiving bead; record per-document conversion
deltas) and Phase 3 leaves the Rust config-contract probe unchecked.
The plan also does not record that the shipped group fix (exact-name matching) differs
from the planned one (ordering the Rust group first)—a better fix worth one outcome
note, in the document that otherwise records outcomes diligently.
**Fix:** Close the completed beads and sync; make the status line agree with the
checkboxes; add the one-line outcome note.

## Suggestions (Non-Blocking)

**S1. Trim `rust-rules` (and spots in `rust-project-setup`/`rust-cli-rules`) against the
capable-agent bar.** Candidates where a strong model needs no instruction: the first two
Ownership bullets (borrow-when-reading / own-when-storing), “Organize modules around
coherent responsibilities,” “Explain why, not syntax” (this is `general-comment-rules`,
which is always loaded), the first two Performance bullets, and the explanatory halves
of the `thiserror`/`anyhow` bullets (keep the default-plus-departure-condition framing,
one line each; cut the rationale).
In `rust-project-setup`, “Document the Supported Surface” and “Keep Repository
Configuration Minimal” are mostly inventory a capable agent produces unprompted—the two
bullets worth keeping are `.gitattributes` newline policy and license files matching the
manifest expression.
Keep, emphatically: the clone-as-design-signal destructuring idiom, `Cow` criteria,
no-lock-guards-in-APIs, `Default`-requires-a-decision, `LazyLock` sufficiency,
no-compatibility-re-exports, cancellation safety, bounded queues, the Edition 2024
review list, and the whole SIGPIPE section in `rust-cli-rules`—that is the non-obvious
material.

**S2. Calibrate description frontmatter length.** Self-routing descriptions are the
right goal (they make `tbd guidelines --list` and the directory tables useful), but
several run 60-80 words and render in full in the four generated SKILL surfaces every
session loads (`code-review-rules`, `ci-and-gates-rules`, `filesystem-rules`,
`release-engineering-rules`, `rust-filesystem-rules`, `rust-code-review-rules`). Most
compress ~40% with no routing loss by dropping the trailing “Load …” clause (the group
note already says when to load) and one level of enumerated detail.

**S3. Update the `new-guideline` shortcut to match the process this PR actually
exercised.** It currently says “one-line description,” and omits: the `category`
frontmatter field, `docs_cache.files` registration in `.tbd/config.yml`, the
`GUIDELINE_GROUPS` explicit-name sets for non-prefixed names (plus the
`guideline-groups.test.ts` update), the top `**Related**:` block convention, and
`globs`/`alwaysApply` for always-load language documents.
The next guideline author will rediscover each of these the hard way; this PR is the
evidence of the real checklist.

**S4. Consider ordering “Cross-cutting engineering topics” directly after “General
engineering”** in `GUIDELINE_GROUPS`. Its note says “in any language,” and it currently
renders after Convex, below all language groups it applies to.

**S5. Dedupe the quick-scan tables.** About five rows in
`rust-code-review-rules.md:84-105` are not Rust-specific and repeat the neutral table in
`code-review-rules` (success-before-verify, discarded required result,
one-implementation trait, suppression without tracker ID, ignored test without tracker).
Keep the genuinely Rust rows (`unsafe` without safety argument, lock guard across
`await`, `filter_map(Result::ok)`, `#[allow]` where `#[expect]` expires, `mod.rs`
convention).

**S6. The `tbd-eidy` bug’s proposed CI check is this corpus’s own rule applied to this
repo.** “Regenerate in CI and fail on any difference” (`ci-and-gates-rules`, Generated
Files Have Exactly One Owner) is precisely a drift test for `tbd setup --auto` over the
committed skill surfaces—worth doing while the rule is fresh.

## False Positives (Checked, Confirmed Benign)

- **`@typescript-eslint/no-restricted-imports` in `filesystem-rules`** matches the
  repo’s real `eslint.config.js:132` (the extension rule, not core
  `no-restricted-imports`)—the example faithfully quotes the enforced config.
- **The `check-eslint-contract.mjs` excerpt in `ci-and-gates-rules`** matches
  `scripts/check-eslint-contract.mjs` (`curly`, `no-floating-promises`,
  `use-unknown-in-catch-callback-variable`), including the survived-prettier comment.
- **The pipeline-exit-status example** is correct in both directions: a failing
  `cargo tree` yields empty grep input, `grep -q` returns 1, `!` inverts to 0; the
  “good” form fails the command’s own error and preserves the inverted-match check.
- **README “40+ guideline documents”**: 43 are bundled on this branch.
- Verified correct as stated: `#[expect(lint, reason)]` warning when unfulfilled;
  `unsafe_code` `deny`-not-`forbid` rationale; `forbid` being non-overridable;
  `allow-unwrap-in-tests` not covering `tests/`/examples/build scripts (matches the PR’s
  empirical 996-versus-25 measurement); `disallowed_methods` having no test-scoping
  option; lefthook honoring `priority` only when not parallel; `npx`/`pnpm dlx`/`bunx`
  fetch-on-missing versus `pnpm exec` fail-on-missing; `ErrorKind::CrossesDevices`
  (stable 1.83) under the examples’ 1.85 MSRV; `with_extension` turning `archive.tar.gz`
  into `archive.tar.old`; Rust ignoring SIGPIPE by default; edition 2024 requiring
  `rust-version` ≥ 1.85; clap `default_value_t` with `ValueEnum`; virtual workspaces
  needing an explicit `resolver`; the Clippy `restriction` group containing
  contradictory lints.
- **`generateShortcutDirectory`’s heading-based double lookup** after
  `guidelineGroupFor` is slightly indirect but keeps the rendered directory on the exact
  function the tests pin; fine as is.
- **The fdu lint-cost numbers** (external repo at `d42d970`) are not independently
  reproducible from this session; the method is stated, internally consistent, and the
  two traps it surfaced are documented—accepted as evidence.

## Documentation

- Plan spec: status line, two Phase 1 checkboxes, Phase 3 probe checkbox, and the
  group-fix outcome note (R11); the “Bundle or retire” Phase 2 item reads as done while
  the outcome note honestly records two documents left undecided—reword the checkbox or
  the note so they agree.
- `docs/development.md` needs no changes (the new material is bundled package content,
  not repo workflow).
- README table and count updated correctly; `typescript-testing-guidelines`
  superseded-note is a good pattern for future absorptions.

## CI Status

All green at `078938d`: ubuntu (Node 22.12.0, Node 24), macOS, Windows test jobs;
Coverage & Lint; Benchmark; DeepSource (grade A, secrets clean).
Local run on the branch: 158 files, 2,415 tests passed.
Coverage bot flags `doc-cache.ts` at ~64% lines—the uncovered ranges are pre-existing
rendering paths, and `guidelineGroupFor` itself is covered by the new test.


---

jlevy:
All eleven findings addressed in `be56558`, plus S3, S4, and S5. Merged your review doc onto the branch first (`9ae5dff`).

**R1 — you were right, and I verified it independently before changing it.** From `tempfile-3.27.0`: the doc comment at `src/file/mod.rs:170` and `:734` reads "If a file exists at the target path, persist will atomically replace it"; `TempPath::persist` calls `imp::persist(.., true)` at `:205` while `persist_noclobber` passes `false` at `:260`; and `src/file/imp/windows.rs:105` sets `MOVEFILE_REPLACE_EXISTING` when overwrite is requested. My claim was backwards. Replaced with the real distinction (`persist` replaces on every platform, `persist_noclobber` is the fail-if-exists variant) plus the genuine Windows caveat — `persist` can fail with a permission error when the destination is open, since there is no POSIX-style replace-while-open.

Your reading of the consequence was the sharp part: an agent believing it would treat `persist` as collision-safe and silently clobber, inverting `filesystem-rules`' own "never make overwrite the fallback" rule.

**R2 — the command overrode what it verified.** Deleted. The step is now the deliberate-violation probe run with no extra flags, with the ordering rule stated (`[lints]` flags come first, args after `--` come later, last of equal specificity wins) so nobody re-adds it. I'd hit this exact behaviour earlier in the session without drawing the conclusion: an `-A warnings` run against fdu silently suppressed everything the manifest denied.

**R7 — the sharpest finding.** You're right that the hand-copied array couldn't catch the class it was written for. `doc-cache.ts` now exports `EXPLICITLY_GROUPED_GUIDELINES` and the test reads it. Confirmed by adding `observability-rules` to `CROSS_CUTTING_NAMES` with no document: the new test fails with `named in a guideline group but not bundled: observability-rules`, where the old one passed.

**R3** — all eight deduplicated. (b) was the worst: `rust-testing-rules` restated three neutral sections of a document it declares itself a supplement to. What replaced them is Rust-only — `include_str!` versus runtime reads and why, and `#[ignore = "reason"]`, which I checked actually prints the reason back (`test tests::skipped ... ignored, flaky under CI: tbd-1234`), making it strictly better than a comment.

**R4, R5, R6, R8, R9, R10** — all applied. On R8, you identified the real defect precisely: the plan's conversion checklist *instructed* keeping both lists. I've struck that line in the plan rather than silently fixing the output.

**R11** — status line now agrees with the checkboxes, and the two genuinely-open items say why (one needs the playbook repo, one needs a Rust codebase). Added the outcome note you asked for, and it's worth stating plainly: the shipped group fix is better than the planned one because ordering only *moves* the collision — the next `<lang>-testing-rules` hits it again, and so does any group added above.

**S3, S4, S5** taken. `new-guideline` now documents the real registration checklist, including the colon-space YAML trap — an unquoted `foo: bar` in a `description` parses as a nested mapping, which broke the build mid-branch and which the shortcut gave no warning about.

**S1 (trimming `rust-rules`) and S2 (description length) not done in this PR.** Both are judgment calls about the corpus-wide bar rather than defects in this change, and S1 in particular would touch material inherited from the playbook that `python-rules` and `typescript-rules` also carry. Better as a deliberate pass across the whole family than as a late edit here — happy to take it as a follow-up if you'd like it done now.

On your note that the fdu lint numbers aren't reproducible from your session: fair, and worth recording. Method was `cargo clippy --locked -p fdu-core -p fdu --all-targets --all-features --message-format=json` with each candidate at `-W`, deduplicating primary spans by `(lint, file, line, column)` — `--all-targets` compiles the lib twice, so raw counts roughly double — then splitting each file at its first `#[cfg(test)]` line.

Validation: `pnpm test` 2415 passed / 158 files, `pnpm ci:quality` clean, skill surfaces regenerated.

---

jlevy:
A second, holistic review is published at [Holistic engineering-guideline review](https://github.com/jlevy/tbd/blob/13dbb8c861131af62c93328d04bf2801237e94be/docs/project/reviews/review-2026-08-23-pr258-holistic-engineering-guidelines.md).

Verdict: **request changes** — 8 High and 5 Medium findings. This pass reviews the corpus as a cross-language decision system for highly capable coding agents and does not repeat the 11 findings already addressed from the earlier review.

The main blockers are the lint-cost evidence method, two false-green gate recipes, mutable action pins, overbroad atomic-write policy, mandatory context cost, universal “strictest wins” selection, FFI unwind semantics, and broken-pipe exit handling.

jlevy:
All thirteen findings from the holistic review are addressed in `cc13b07`. Three of them were places where this branch broke rules the corpus itself states, so those got fixed by demonstration rather than by editing prose.

**R1 — the evidence was wrong, and by how much.** You were right that a `#[cfg(test)]` split is not a boundary. I reconstructed that method over the raw diagnostics and compared it against a compile-unit classification: **145 of 415 diagnostics (34%) were misfiled, in both directions** — 35 `expect_used` hits in inline test modules counted as production, and 27 `indexing_slicing` hits in production code counted as tests.

The measurement now runs two clippy passes (`--lib --bins` vs `--all-targets`) and classifies by cargo's per-diagnostic target, into three buckets — ships / build scripts / tests. Reproducer at `scripts/measure-rust-lint-cost.mjs`, the 415-row raw mapping and method at `docs/project/research/current/evidence-2026-08-23-rust-lint-cost.{md,tsv}`.

| Lint | Ships | Build | Tests | Was published as |
| --- | ---: | ---: | ---: | --- |
| `let_underscore_future` | 0 | 0 | 0 | 0 / 0 |
| `panic` | 2 | 2 | 35 | 2 / 4 |
| `wildcard_enum_match_arm` | 12 | 0 | 9 | 7 / 11 |
| `disallowed_methods` (fs) | 0 | 1 | 82 | 1 / 82 |
| `expect_used` | 32 | 7 | 35 | 25 / 45 |
| `indexing_slicing` | 79 | 0 | 119 | 62 / 136 |

Every nonzero figure moved. No verdict reverses, and two are now shown instead of asserted: `panic`'s build-script share is half its non-test cost, and `disallowed_methods` has *zero* shipping sites.

One thing fell out of doing this that is worth its own line in the corpus: the first corrected run reported 9 diagnostics instead of 415, because fdu denies warnings and the run died in the first crate's build script. A measurement that stops at the first failure measures whichever crate compiled first — the gate failure mode arriving through the back door. The script now caps lints at `warn` and the writeup says why.

**R2 — both recipes reproduced failing before they were fixed.** The lint-policy loop exits 0 both when it prints complaints *and* when a member sits outside `crates/*` (in that case it prints nothing at all). The corrected version enumerates from `cargo metadata`, accumulates, asserts a non-empty member list, and exits nonzero — verified against fdu (passes, 3 manifests), a workspace with an unlinted member at `tools/helper` (fails), and a zero-member workspace (fails). The strict cross-lint target fails on this runner, which has no cross targets installed; the permissive one skips and passes. Both are now two named targets rather than one.

**R3 — no tag exception, and the repo now follows the rule.** All 15 action references pinned to commit SHAs with trailing version comments, `dependabot.yml` added so pinning doesn't mean freezing, and `scripts/check-action-pins.mjs` added as the gate (wired into `lint:check`, with negative tests for floating tags, exact tags, branches, and short SHAs). Also fixed the unpinned `actions/checkout@v6` example inside `supply-chain-hardening`.

**R5 — the budget is now a gate, not a comment.** Always-load core: 2,233 lines / 11,806 words → **909 lines / 4,663 words**, by routing testing, TDD, goldens, compatibility, and commits by changed surface. Membership is an explicit set rather than a `general-` prefix, so a new `general-*` document can no longer admit itself to what every session reads; `guideline-budget.test.ts` asserts the ceiling and a negative case proves the prefix no longer admits. `skill-baseline` and the generated directory now describe the same three-layer model.

I took the prose-note route for routing rather than the `topics` / `appliesTo` metadata refactor — the note in the generated directory is what the agent actually reads, so it delivers the budget without the schema change. Called out as not-done in the plan alongside S2's decision-scenario suite and per-rule admission records from S1.

**R6, R7, R9, R10** — strictest-wins replaced with six admission criteria plus per-project-shape departure conditions (MSRV < 1.81, macro-expanded `unsafe impl`, proc macros, unpinned toolchain). FFI matrix corrected in both directions. `abi3` removed from the binary-wheel path, with extension modules given their own subsection — including the feature-polarity trap, stated as the invariant rather than a polarity, since fdu deliberately defaults it on and tests with `--no-default-features`. The replacement example is rebuilt around a real `BufWriter` with `into_inner` and conditional sync, and compile-verified.

**R4, R8, R11, R12, R13** — five named write contracts with atomic replacement scoped to authoritative-path replacement; the broken-pipe contract moved to `error-handling-rules` with both language recipes fixed so only the primary stdout renderer converts and a closed stderr can never lower a nonzero status; count-based test proxies replaced with independent-evidence selection; both quick-scan tables converted to "question that decides it" plus consequence; the authoring shortcut builds before validating and validates through `packages/tbd/dist/bin.mjs`.

**S4** is now a rule in `ci-and-gates-rules`: a check nobody has watched fail is not a gate, and checks that matter belong in a file with a positive and a negative test beside them.

Validation: `pnpm ci:quality` green, 2426 tests across 160 files pass, Rust examples compile clean, both gate recipes exercised against positive and negative fixtures.

---

jlevy:
Follow-up implementation is now stacked in #260: https://github.com/jlevy/tbd/pull/260. It is rebased on the parent’s R1–R13 response, preserves that work, and adds explicit decision routing, the budgeted admission-controlled core, executable routing and Rust-gate checks, the cross-language CLI I/O contract, and the corresponding tbd runtime fix.

jlevy:
The preservation-first second pass is now published in the standalone review: [R14-R22: Second-Pass Preservation and Precision Review](https://github.com/jlevy/tbd/blob/a55041a0/docs/project/reviews/review-2026-08-23-pr258-holistic-engineering-guidelines.md#second-pass-preservation-and-precision-review).

Verdict at reviewed commit `0c34446b`: **request changes**. All nine findings are addressed in `a55041a0`. The important preservation correction is R14: the original `general-testing-rules` list is restored verbatim, and the new independent-evidence material is additive. The executable fixes have red-before/green-after coverage for partial Clippy evidence and YAML action-reference parsing.

Validation: `pnpm run ci` passed, and the pre-push hook independently passed all 2,431 tests across 161 files. GitHub CI is running now. Deferred omission: `tbd-yk6w` tracks composite-action recursion and immutable Docker digest handling.

jlevy:
Windows CI correctly rejected the first regression-test shim: Node does not execute a `.cmd` file through `execFileSync`, so PATH fell through to the runner Cargo. `1fca4385` replaces it with a real `cargo.exe`/`cargo` executable name backed by the current Node binary and a preload hook; no Rust installation is required by the Node test. The focused test, quality gate, and pre-push run of all 2,431 tests pass. CI has restarted on the corrected head.

PR #264, opened 2026-08-26
fix: honor deferred_until and route guidelines by declared category

jlevy:
Three defects from the open bug beads, each verified live against the code before
fixing and pinned by a test that fails against the old behavior. No release in this PR —
`packages/tbd/package.json` stays at `0.8.1`.

## tbd-5av0 — `deferred_until` was recorded and then ignored

`--defer` wrote the field and `tbd ready` went on offering the bead. Verified on the
0.8.1 build before the fix: a bead deferred to `2027-01-01` appeared in `tbd ready`
today, so the field read as scheduling while scheduling nothing. The bead reports this
found during a real triage — 21 beads deferred to 2027 all stayed in `ready` until each
was separately set to `--status deferred`.

`readyIssueIds` now treats a future deferral as the hold it already documents for
`blocked` and `paused`. A deferral exactly at `now` counts as arrived.

The cutoff is a **parameter**, not a `Date.now()` read inside the filter.
`issue-changes.ts` computes this set twice to diff two snapshots; two independent clock
reads could report a deferral that merely elapsed between them as a ready transition no
edit caused. That path now passes one instant to both calls. `planMirror` is documented
as `Pure.`, so it takes an explicit `readyAt` rather than quietly acquiring a clock
dependency.

This changes readiness everywhere it is computed — `ready`, `web`, and the Linear
mirror — which is the point: `issue-query.ts` exists so two surfaces answering the same
question cannot disagree.

## tbd-v8lv — `--defer-before` was a silent no-op

`ListOptions` declared `deferBefore` and the flag appeared in `--help`, but no filter
ever read it. Verified before the fix: `tbd list --defer-before 2020-01-01` returned a
bead whose `deferred_until` was 2027.

It now filters. A bead with no `deferred_until` is not "deferred before" any date, so it
is excluded whenever the flag is set, rather than treated as deferred since the
beginning of time.

An unusable date is now **rejected at the CLI boundary** instead of degrading into a
filter that matches everything — that would be the same silent-no-op failure in a new
place:

```
$ tbd list --defer-before not-a-date
Error: Invalid --defer-before value: not-a-date. Expected a date or timestamp.
$ echo $?
2
```

`parseDateOption` moves from `update.ts` into `cli/lib/issue-input-validation.ts` so
`--due`, `--defer`, and `--defer-before` all fail the same way.

## tbd-dado — guidelines routed by filename prefix, not declared category

Every bundled guideline declares a `category`, and `doc-categories.ts` states plainly
that "the old name-based inference is retired in favor of the declared field." Grouping
was still matching name prefixes, and the two disagreed:

| Guideline | Declared | Was filed under |
| --- | --- | --- |
| `electron-app-development-patterns` | `desktop` | TypeScript & JS ecosystem |
| `electrobun-app-development-patterns` | `desktop` | **Docs, process & tooling** |
| `tauri-app-development-patterns` | `desktop` | **Docs, process & tooling** |

`n.startsWith('electron-')` is false for `electrobun-`, and `tauri-` had no matcher at
all — so an agent asked to work on a Tauri app would not find the guidance under any
framework heading. Tauri's core is Rust, so widening the TypeScript matcher would have
been wrong in the other direction.

Grouping now reads the declared field. The three frameworks share a **Desktop app
frameworks** heading, and the catch-all is left holding only genuinely miscellaneous
docs.

The regression guard is the general one, not a spot check: **every declared category
with bundled documents must route out of the catch-all**, so the next category added
cannot land there unnoticed — which is exactly how this shipped.

Also moved `release-notes-guidelines` into the cross-cutting group beside
`release-engineering-rules`, whose own description tells the reader to load the two
together. Flagging it separately since the bead only called it "arguably" misfiled —
easy to drop if you disagree.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm test` | 2464 passed, 164 files (was 2449 / 163) |
| `pnpm lint:check` | pass |
| `pnpm format:check` | pass |

Red-green on all three: the 9 new assertions were confirmed failing against the old code
before any fix, reproducing each reported symptom. Then re-verified through the built
CLI end to end — `ready` excludes the deferred bead, `--defer-before` filters and
rejects a bad date with exit 2.

## Beads

Closes tbd-5av0, tbd-v8lv, tbd-dado. Separately verified **tbd-az97** as already fixed
(a single-branch clone now discovers the remote sync branch, shares its history, and
converges) — not closed here since it is unrelated to this diff.

Left open deliberately: **tbd-pht1 / tbd-iiys**, which are live. An orphaned
`data-sync.lock` blocks every bead write and `tbd doctor` reports the repository
healthy; I reproduced a `tbd create` hanging until killed. The 30-minute `staleMs` is a
deliberate safety contract (`lockfile.ts` says so in as many words), so the fix is
observability — doctor flagging the lock, and the blocked command saying what it waits
on — not a timing change. Worth its own PR.

---

https://claude.ai/code/session_013eaEeT9CYT5TDyNGAw95t9

### DIFF
--- a/.agents/skills/tbd/SKILL.md
+++ b/.agents/skills/tbd/SKILL.md
@@ -407,6 +407,7 @@ Load the **General engineering** core, then only guidelines matching the task.
 | general-testing-rules | Rules for keeping test volume low while preserving broad evidence—rejecting vacuous tests, choosing portable black-box tests when they preserve coverage, keeping the inner loop fast, controlling nondeterminism, and never letting an empty or skipped selection look like a pass. |
 | golden-testing-guidelines | Guidelines for implementing golden/snapshot testing for complex systems |
 | release-engineering-rules | Language-neutral rules for turning a reviewed commit into artifacts users execute—one release identity, a pre-release gate that runs where publishing happens, least-privilege publishing authority, build-once-and-promote, packaging and checksums, smoke-testing the packaged artifact rather than the build output, multi-channel coordination, testable release logic, and incident preparation. Load for any release, alongside release-notes-guidelines and the language-specific release document. |
+| release-notes-guidelines | Rules for release notes that describe the published delta and exclude defects introduced and corrected before release from separate Fixes entries |
 | supply-chain-hardening | Strongly recommended for EVERY repo—apply it if a repo has not been hardened yet. Cross-ecosystem policy for installing dependencies safely (the 14-day cool-off, disabled install scripts, lockfile discipline, untrusted-repo handling). Use whenever a user mentions hardening, security, supply chain, or setting up a new repo; before adding/upgrading dependencies; when auditing for compromised packages; or when reviewing install/build/run commands across npm/pnpm, PyPI, Cargo, or Go. |
 
 ### TypeScript & JS ecosystem
@@ -416,7 +417,6 @@ Load the **General engineering** core, then only guidelines matching the task.
 | Name | Description |
 | --- | --- |
 | bun-monorepo-patterns | Modern patterns for Bun-based TypeScript monorepo architecture |
-| electron-app-development-patterns | Building a clean, minimal, standalone Electron app—process model, modern Vite-based build system, attaching a Node/Bun/Python backend, security baseline, packaging, code signing, and auto-update |
 | pnpm-monorepo-patterns | Modern patterns for pnpm-based TypeScript monorepo architecture |
 | typescript-cli-tool-rules | Rules for building CLI tools with Commander.js, picocolors, and TypeScript |
 | typescript-code-coverage | Best practices for code coverage in TypeScript with Vitest and v8 provider |
@@ -459,6 +459,16 @@ Load the **General engineering** core, then only guidelines matching the task.
 | convex-limits-best-practices | Comprehensive reference for Convex platform limits, workarounds, and performance best practices |
 | convex-rules | Guidelines and best practices for building Convex projects, including database schema design, queries, mutations, and real-world examples |
 
+### Desktop app frameworks
+
+*Select the document for the framework in use; do not load this whole group by default.*
+
+| Name | Description |
+| --- | --- |
+| electrobun-app-development-patterns | Building desktop apps with Electrobun—runtime and process model, typed RPC, project layout, packaging and the delta updater, plus an evidence-based maturity and security assessment |
+| electron-app-development-patterns | Building a clean, minimal, standalone Electron app—process model, modern Vite-based build system, attaching a Node/Bun/Python backend, security baseline, packaging, code signing, and auto-update |
+| tauri-app-development-patterns | Building desktop apps with Tauri 2—the Rust core and system webview model, capabilities and permissions, typed commands and IPC, attaching Rust or non-Rust backends, packaging, signing, and the signed updater |
+
 ### Docs, process & tooling
 
 | Name | Description |
@@ -466,9 +476,6 @@ Load the **General engineering** core, then only guidelines matching the task.
 | agent-session-bootstrap | When and how to make a repository install its own pinned toolchain at agent session start, for repos whose agents run in containers they do not control. Covers the fit test, the alternatives that are usually better, the install rules a bootstrap must follow, and the PATH and pin-drift traps that make one fail silently. Use when an agent session starts without the tools the repo requires, when writing or reviewing a SessionStart hook, or when deciding between a session hook and a provisioned image. |
 | cli-agent-skill-patterns | A concise decision guide for portable skills, CLI-backed skills, safe bundle installation, and agent integration |
 | common-doc-guidelines | Common cross-project standards for writing and organizing docs, code comments, and text files—how to organize, structure, write, and format documents, plus the guideline footer convention. Downstream of github.com/jlevy/practical-prose. Use whenever writing or editing any documentation, README, guideline, or design doc. |
-| electrobun-app-development-patterns | Building desktop apps with Electrobun—runtime and process model, typed RPC, project layout, packaging and the delta updater, plus an evidence-based maturity and security assessment |
-| release-notes-guidelines | Rules for release notes that describe the published delta and exclude defects introduced and corrected before release from separate Fixes entries |
-| tauri-app-development-patterns | Building desktop apps with Tauri 2—the Rust core and system webview model, capabilities and permissions, typed commands and IPC, attaching Rust or non-Rust backends, packaging, signing, and the signed updater |
 | tbd-sync-troubleshooting | Common issues and solutions for tbd sync and workspace operations |
 
 <!-- END SHORTCUT DIRECTORY -->
--- a/.claude/skills/tbd/SKILL.md
+++ b/.claude/skills/tbd/SKILL.md
@@ -407,6 +407,7 @@ Load the **General engineering** core, then only guidelines matching the task.
 | general-testing-rules | Rules for keeping test volume low while preserving broad evidence—rejecting vacuous tests, choosing portable black-box tests when they preserve coverage, keeping the inner loop fast, controlling nondeterminism, and never letting an empty or skipped selection look like a pass. |
 | golden-testing-guidelines | Guidelines for implementing golden/snapshot testing for complex systems |
 | release-engineering-rules | Language-neutral rules for turning a reviewed commit into artifacts users execute—one release identity, a pre-release gate that runs where publishing happens, least-privilege publishing authority, build-once-and-promote, packaging and checksums, smoke-testing the packaged artifact rather than the build output, multi-channel coordination, testable release logic, and incident preparation. Load for any release, alongside release-notes-guidelines and the language-specific release document. |
+| release-notes-guidelines | Rules for release notes that describe the published delta and exclude defects introduced and corrected before release from separate Fixes entries |
 | supply-chain-hardening | Strongly recommended for EVERY repo—apply it if a repo has not been hardened yet. Cross-ecosystem policy for installing dependencies safely (the 14-day cool-off, disabled install scripts, lockfile discipline, untrusted-repo handling). Use whenever a user mentions hardening, security, supply chain, or setting up a new repo; before adding/upgrading dependencies; when auditing for compromised packages; or when reviewing install/build/run commands across npm/pnpm, PyPI, Cargo, or Go. |
 
 ### TypeScript & JS ecosystem
@@ -416,7 +417,6 @@ Load the **General engineering** core, then only guidelines matching the task.
 | Name | Description |
 | --- | --- |
 | bun-monorepo-patterns | Modern patterns for Bun-based TypeScript monorepo architecture |
-| electron-app-development-patterns | Building a clean, minimal, standalone Electron app—process model, modern Vite-based build system, attaching a Node/Bun/Python backend, security baseline, packaging, code signing, and auto-update |
 | pnpm-monorepo-patterns | Modern patterns for pnpm-based TypeScript monorepo architecture |
 | typescript-cli-tool-rules | Rules for building CLI tools with Commander.js, picocolors, and TypeScript |
 | typescript-code-coverage | Best practices for code coverage in TypeScript with Vitest and v8 provider |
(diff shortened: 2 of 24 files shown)


jlevy:
## Review: PR #264

**Scope.** `cb771d2` vs `main` (`2c2beff`), 11 files, +294/−21, via `gh pr diff`. Verified by building this branch (`0.8.2-dev.1.cb771d2`), running the two new test files, mutation-testing each fix (revert the predicate, rerun only the relevant file, restore), and exercising `ready`, `list --defer-before`, and `create --defer` end to end in a scratch repo. Loaded: `general-eng-agent-principles`, `code-review-rules`, `typescript-rules`, `general-testing-rules`, `backward-compatibility-rules`.

**Verdict: sound on the paths it touches; not ready to un-draft yet.** The three fixes do what the title says, the boundary choices are deliberate and tested, and every fix is red/green (details under R4). Two things should land first: the committed agent surfaces are stale (R3, part of tbd-dado's stated fix) and the wiring that was actually broken is still unguarded (R4). R1 and R2 are the same two-clock defect shape the PR's own comments describe, in consumers the PR did not touch; small enough to belong here, but I would not block on them. Un-draft after R3 and R4.

**CI:** all green at `cb771d2` (Benchmark, Coverage & Lint, DeepSource, Test x4). Branch is 4 commits behind `origin/main`; those commits touch no file in this PR, `mergeable: MERGEABLE`.

### Findings

**R1 (Medium)** `packages/tbd/src/lib/issue-changes.ts:448`. The module header says "Pure, deterministic issue-snapshot diffing" and this line adds a hidden `Date.now()`. The PR threaded `readyAt` through `planMirror` for exactly this reason (its docstring says `Pure.`), then did the opposite here. Consequence: `tbd changes` / `tbd watch --ready` output for a fixed pair of snapshots now depends on wall clock, tests cannot pin it, and `board.ts:993`, which already has an injectable `dependencies.now()`, cannot pass it through.
**Fix:** add a required `readyAt: number` to `CreateIssueChangesReportOptions`; pass `Date.now()` from `sync-branch-changes.ts:357` and `this.dependencies.now().getTime()` from `board.ts:993`. Keep the "one instant for both snapshots" comment; it is correct.

**R2 (Medium)** `packages/tbd/src/cli/web/board.ts:942` and `:1109` vs `issue-query.ts:111`. One board response reads the clock twice. `snapshot.readyIds` is computed at load time and drives the row `ready` marker (1109); the Ready checkbox goes through `filterIssues`, which calls `readyIssueIds(issues)` fresh per request. Before this PR readiness was time-independent so the two agreed. Now a bead whose deferral elapses while `tbd web` is up shows in the Ready-filtered view without its ready marker until some unrelated data change triggers a reload. Low likelihood, but it is a correctness inconsistency this PR introduces in an untouched consumer.
**Fix (pick one):** (a) compute the ready set once per `buildBoardResponse` at `this.dependencies.now()` and use it for both the row flag and the filter, which needs `filterIssues` to accept a `now` (or a precomputed `readyIds`); (b) drop the cached `snapshot.readyIds` and compute it per response with the same instant the filter uses.

**R3 (Medium)** `.claude/skills/tbd/SKILL.md:462-471`, `.agents/skills/tbd/SKILL.md:462-471`. The committed generated directory still files `electrobun-app-development-patterns`, `tauri-app-development-patterns`, and `release-notes-guidelines` under `### Docs, process & tooling`. `tbd doctor` on this build reports both files as `stale managed file`. tbd-dado listed regenerating the surfaces as part of the fix, and this repo's agents read these files, so the routing fix is invisible here until they are regenerated.
**Fix:** `node packages/tbd/dist/bin.mjs setup --auto` and commit the two SKILL.md files. I ran it: the SKILL.md diff moves exactly the expected rows. It also touches `.tbd/config.yml`; check that hunk before committing it.

**R4 (Medium)** `packages/tbd/src/cli/commands/list.ts:219-222`, `packages/tbd/src/file/doc-cache.ts:628`. Mutation results: removing the `deferralPending` check fails 2 tests, removing the `deferBefore` filter fails 4, and making the desktop matcher return false fails 2, so the predicates are guarded. But dropping the category argument at `doc-cache.ts:628` (`guidelineGroupFor(doc.name)`) leaves `guideline-groups`, `guideline-budget`, and `doc-cache` green at 43/43, and nothing exercises `--defer-before` through commander (`list.ts` is 6% covered per the coverage report; no tryscript mentions the flag). The defect being fixed was a flag that parsed and was never read: that wiring is still the untested part, and a typo in `options.deferBefore` would restore the silent no-op with CI green. tbd-v8lv asked for a transcript.
**Fix:** extend the "Deferred Issues" block in `tests/cli-edge-cases.tryscript.md:386` (it already creates a task deferred to 2099): assert `tbd ready --json` omits it, `tbd list --defer-before 2100-01-01 --json` includes it, `--defer-before 2000-01-01` is empty, and `--defer-before nope` exits 2. Add one test that renders `generateShortcutDirectory` (or the setup output) and asserts `tauri-app-development-patterns` sits under `### Desktop app frameworks`.

**R5 (Low)** `packages/tbd/src/cli/commands/update.ts:64-70`. The `parseDateOption` docstring was left behind and now sits on `class UpdateHandler` ("Read a date option into a full timestamp..."). It also carries the better half of the rationale (why bare dates are accepted), which the moved function's new docstring lacks.
**Fix:** fold that paragraph into the docstring at `issue-input-validation.ts:39-45` and delete it from `update.ts`.

**R6 (Low)** `packages/tbd/src/cli/commands/create.ts:191-192`. The description says `--due`, `--defer`, and `--defer-before` "all fail the same way"; that holds for `update` only. On this build `tbd create "x" --defer 2027-01-01` still fails at the schema with `deferred_until: Invalid datetime`, exit 1. tbd-5av0, which this PR closes, names that bare-date failure, and also asks for a bulk path to `--status deferred`.
**Fix (pick one):** route `create --due/--defer` through `parseDateOption` (two lines; the helper was just made shared for this) and add the bare-date case to the tryscript; or correct the description and file the two leftover tbd-5av0 items as their own beads before closing it.

**R7 (Low)** `packages/tbd/src/lib/issue-query.ts:186-246`. `describeQuery` ignores `deferBefore`: a query with it set is described as a plain `tbd list`, and as an `exact` `tbd ready`. Unreachable today (`board.ts:538` never sets it), but the field was added to the query the describer exists to describe.
**Fix:** add `--defer-before <iso>` to the list branch and `query.deferBefore !== null` to the ready branch's `unsupported` set.

**R8 (Low, judgment)** `packages/tbd/src/lib/issue-selection.ts:81`, `packages/tbd/src/integrations/core/mirror.ts:50`. `now = Date.now()` and `readyAt?` are optional with clock defaults, so the three untouched callers (`issue-query.ts:111`, `sync-engine.ts:555`, `board.ts:942`) silently took their own clock; that is how R1 and R2 arose. The repo's TypeScript rules prefer explicit parameters where omission causes bugs.
**Fix:** make `now` required on `readyIssueIds` and `readyAt` required on `MirrorContext`; the `planMirror` test call sites can share one constant. Defensible to decline for a one-instant CLI process.

**R9 (Low)** `packages/tbd/tests/deferred-until.test.ts` (6 sites). The `as Partial<Issue>` casts are unnecessary: `Issue` has `deferred_until?: string | null`, and `tsc --noEmit` passes with all six removed. Casts on test data are where a genuinely missing field would hide.
**Fix:** delete them.

**R10 (Low)** Docs. `packages/tbd/docs/tbd-docs.md:281` still says "Deferred before date" (state: exclusive; beads with no `deferred_until` are excluded); `:460`, `:671`, `:831` define ready as "open, unblocked, unassigned"; `tbd-design.md:2764`; `IssueQuery.ready` docstring at `issue-query.ts:48`. Also worth one sentence at `tbd-docs.md:829-831`: `watch --ready` never wakes for a deferral that merely elapsed, by design of this PR.
**Fix:** add "and not deferred into the future" at those spots, and the `watch` note.

### Design assessment

Fixing readiness in `readyIssueIds` is the right place; `ready`, `web`, and the mirror all inherit it and I found no better alternative. Not taken from tbd-5av0: having `--defer` also flip status to `deferred`. That leaves the two fields able to disagree (`status: deferred` with no date, `status: open` with a future date), and `--deferred` (status) and `--defer-before` (date) filter different things. Pre-existing, out of scope, but worth a bead if it is not intended.

**Blast radius.** `tbd ready`, the web Ready view, and Linear ready-child counts now exclude open beads with a future `deferred_until`. `tbd list` without flags is unchanged (verified). DO NOT MAINTAIN is correct under `backward-compatibility-rules`: no consumer can legitimately depend on a deferred bead being offered as ready. The Linear change is a visible remote write on the next push; `CHANGELOG.md` has no Unreleased section, so record both in the 0.8.2 notes.

### False positives / do not fix

- Boundaries: `deferralPending` uses `until > now` (exactly now is ready); `deferredBefore` uses `until < cutoff` (exclusive). Verified on the CLI: `--defer-before 2027-01-01` excludes a bead deferred to `2027-01-01T00:00Z`. Consistent with each other and with what `--defer` stores. Do not flip.
- Date-only input parses as UTC midnight (`Date.parse('2027-01-01')`), so `--defer-before 2027-01-01` means before `2027-01-01T00:00Z`, not local midnight. Pre-existing `parseDateOption` behavior shared with `--defer`/`--due`, so both sides agree.
- `readyAt` is computed in `issue-changes.ts` even when `needsReadySets` is false: harmless.
- `Date.parse(cutoff)` runs per issue inside `deferredBefore`: negligible; hoisting is optional.
- `sync-engine.ts:555`: one `readyIssueIds` call per reconcile, no straddle within a run. Its ready-child counts changing is the intended blast radius, not a miss.
- Unparseable `deferred_until` fails open for ready and closed for `--defer-before`. The schema validates on write and both are documented; fine.
- `IssueQuery` gained a required field; every literal (`board.ts:522`, `ready.ts:57`) spreads `defaultIssueQuery()`, and tsc would catch a miss.
- `--defer-before` under the default closed-hiding still hides closed beads unless `--all`, like every other filter.
- The `general` category being skipped in the new routing test is correct: `general` is routed by explicit membership, not by category.

jlevy:
Addressed the review findings in `c1235d3c`. All ten fixed, none deferred or rebutted. CI green on all 7 checks.

**R1, R2, R8 turned out to be one root cause**, which your R8 note already identified: `readyIssueIds` took an optional `now` defaulting to `Date.now()`, and once readiness became time-dependent that default was a defect generator. Fixed structurally rather than site by site:

- **R8: fixed** — `now` is required on `readyIssueIds`, `readyAt` is required on `MirrorContext`, and `filterIssues`/`selectIssues` take the instant explicitly. Taking it rather than declining: with the parameter required, `tsc` enumerated every affected call site, which is how R1 and R2 stopped being possible to reintroduce silently.
- **R1: fixed** — `issue-changes.ts` no longer reads the clock; `readyAt` is a required option. Both test suites now pin the instant, which they could not before, so the purity claim in the module header is testable rather than aspirational.
- **R2: fixed** — took your option (b). Readiness is no longer cached on `BoardSnapshot` at all, since it is a function of the clock and a snapshot is not. It is computed once per `buildBoardResponse` and reused for both the Ready filter and the row marker. `buildBoardResponse` is synchronous, so the per-response field cannot interleave; noted in a comment.

**R3: fixed** — regenerated; the diff adds the `Desktop app frameworks` section, moves electron out of TypeScript, and moves release-notes-guidelines into cross-cutting. Thanks for the `.tbd/config.yml` warning: that hunk was only a dirty dev-build version stamp, so I reverted it and committed just the two SKILL.md files.

**R4: fixed**, and this was the most useful finding. A rendered-output test now asserts the heading a guideline lands under; mutation-checked, dropping the category argument at `doc-cache.ts:628` fails it, where it previously left 43/43 green. Added five tryscript cases driving `--defer-before` through commander per your suggestion: bare-date create, ready exclusion, both boundary directions, and exit 2 on an unparseable value. All 47 tryscript cases pass.

While writing that test I asserted `release-notes-guidelines` would land in the catch-all and it landed in cross-cutting. The test was wrong, not the code; corrected to the real grouping.

**R5: fixed** — folded into the `parseDateOption` docstring, deleted from `update.ts`.

**R6: fixed** — took the first option. `create --due/--defer` now route through `parseDateOption`. Verified end to end: `--defer 2027-01-01` stores `2027-01-01T00:00:00.000Z`, `--defer nope` exits 2 with the named error, and `ready` excludes the result. The bare-date case is in the tryscript.

**R7: fixed** — the list branch renders `--defer-before`, and the ready branch counts it as unsupported.

**R9: fixed** — casts removed, `tsc` clean.

**R10: fixed** — readiness is documented as excluding future-deferred beads at all four sites plus the `IssueQuery.ready` docstring, `--defer-before` is documented as exclusive and as excluding beads with no `deferred_until`, and `watch --ready` now says an elapsed deferral never wakes it.

Nothing in your false-positives section was touched. On the design note about `--defer` not flipping status: agreed it is pre-existing and out of scope; filing a bead rather than changing it here.

**Un-drafting**, since you gated that on R3 and R4 and both are done.

Beads: `tbd-d153` (parent) with `tbd-a3u1`, `tbd-7ab3`, `tbd-z2nv`, `tbd-dh9k`, `tbd-5cix`, `tbd-o3au`, `tbd-3zzt`, `tbd-o3kf`, `tbd-q3hl`, `tbd-pmf7`.

PR #266, opened 2026-09-02
fix: enforce a gh version floor, bump the pin to 2.97.0, and add stacked-PR support

jlevy:
Review of how we use `gh` locally and of all `gh` guidance in tbd, plus the fixes that review turned up.

## What the review found

**Local state:** `gh` 2.98.0, authenticated via keyring, with `gh stack` v0.1.0 already installed by hand. tbd knew about none of it, so the local machine had drifted well ahead of what tbd provisions.

**Three defects in the install path:**

1. **The pin (2.92.0) missed four security advisories** fixed in 2.97.0. Two land on paths agents use constantly:
   - `gh auth status` printed part of the auth token in plaintext for `ghs_*`, `github_pat_*`, and `ghu_*` formats ([GHSA-cg6r-mpgc-h9mm](https://github.com/cli/cli/security/advisories/GHSA-cg6r-mpgc-h9mm)). The ensure script and every PR shortcut run this command, and agents capture its output.
   - `gh api` and `gh pr diff` printed externally controlled content without neutralizing terminal escape sequences ([GHSA-3m3g-3wcr-px46](https://github.com/cli/cli/security/advisories/GHSA-3m3g-3wcr-px46)). Our review shortcuts pipe both straight into agent context.
2. **`ensure-gh-cli.sh` never checked the version.** It gated on `command -v gh` alone, so any `gh` was accepted forever and a distro package silently defeated the pin. `setup-github-cli.md` already claimed the script reinstalls on a wrong version; it did not. Code and doc now agree.
3. **No extension or skill provisioning,** so `gh stack` was absent on every fresh machine.

**One guidance bug, the most severe:** all four PR shortcuts hardcoded `--base main`. On a stacked branch that retargets the PR at the trunk, shows the reviewer every lower layer's diff, and breaks the stack on GitHub. An agent following our own shortcut silently undid the user's stack.

## Why 2.97.0 and not 2.98.0

2.98.0 was published 2026-08-20, making it 11 days old, which fails the 14-day cool-off in `SUPPLY-CHAIN-SECURITY.md`. 2.97.0 (2026-07-31, 31 days) carries all four fixes and the mature `gh skill` commands. The script never downgrades, so machines already on 2.98.0 keep it.

## Adopting GitHub's own skill rather than writing ours

`github/gh-stack` ships an official agent skill. Pinned at v0.1.0 it is 892 lines of operational detail we would otherwise have to reverse-engineer: which subcommands open a blocking TUI under a PTY, the exact flags that avoid prompts, exit codes, conflict recovery, `--json` parsing.

So the split is: **the official skill owns the mechanics, tbd owns the policy.** The new `stacked-prs` shortcut covers only when stacking is worth it, layer discipline, how stacks map to beads, and how the other shortcuts change. It does not document `gh stack` commands.

Both the extension and the skill are pinned to `v0.1.0` (published 2026-07-29, 33 days old, clears the cool-off).

## Stacking stays opt-in

Per review feedback, this does not push anyone toward stacked PRs:

- Provisioning is behind an explicit `--with-stack` flag, deliberately **not** in the SessionStart hook, so sessions that never stack pay no network cost. Failures warn and exit 0 rather than blocking a session.
- The guidance says to honor a stack when the user asks for one, offer it once when a change plainly decomposes, and drop it when declined.

## Verification done here

- Version comparison unit-tested across 12 cases: lexicographic traps (2.9.0 < 2.97.0), octal-looking segments (2.08.0), prerelease suffixes, and the no-downgrade case.
- Default script path run end to end on macOS: correctly keeps 2.98.0 and now reports auth honestly. It previously printed `GH_TOKEN not set - some operations may require authentication` on a fully working keyring-authenticated machine.
- Pinned skill fetch verified into a scratch directory (v0.1.0 resolved to `a1b4a3d4`).
- `gh repo view` and `gh stack view --json` verified against the live CLI. This caught a bug in my own first draft: `gh repo view` takes the repo positionally and has no `--repo` flag.
- Full repo gate green: 163 files, 2449 tests.

**Not verified locally:** the user-scope skill install branch, since it changes the machine globally and `gh skill` has no uninstall command. Phase 4 of the QA playbook covers it.

## How to validate this PR

`tests/qa/gh-cli-and-stacked-prs.qa.md` is written for a fresh Linux instance and is the acceptance test. Phases 1 and 6 matter most:

- **Phase 1** starts from a deliberately **old** `gh`, the case the current script silently fails.
- **Phase 6** runs `create-or-update-pr-simple` against a stacked branch and asserts the PR is not retargeted at the trunk.

The playbook asks the validating agent to report what actually happened, including anything contradicting the doc, rather than confirming it.

Beads: `tbd-ewsw` (epic) with children `tbd-hvcr`, `tbd-um28`, `tbd-70a2`, `tbd-njez`, `tbd-d2f6`, `tbd-gc0p`, `tbd-oudg`, `tbd-j4wy`.

### DIFF
--- a/.agents/skills/tbd/SKILL.md
+++ b/.agents/skills/tbd/SKILL.md
@@ -370,6 +370,7 @@ Run `tbd shortcut <name>` to use any of these shortcuts:
 | revise-architecture-doc | Update an architecture document to reflect current codebase state |
 | setup-github-cli | Ensure GitHub CLI (gh) is installed and working |
 | setup-linear | Set up the Linear integration end to end—first-time configuration for a repository, or adding your own API key to a repository your team already configured |
+| stacked-prs | When to split work into a stack of dependent PRs, how stacks line up with beads, and how the PR shortcuts change when a branch is part of a stack |
 | suggest-upstream-improvements | Review local doc-fork customizations and contribute the generally useful changes back upstream |
 | sync-failure-recovery | Handle tbd sync failures by saving to workspace and recovering later |
 | update-specs-status | Reconcile active specs, the top-level work index (e.g. TODO.md), and tbd beads into one current status map |
--- a/.claude/scripts/ensure-gh-cli.sh
+++ b/.claude/scripts/ensure-gh-cli.sh
@@ -8,6 +8,10 @@
 # that bypasses the cool-off window. To bump the pin, pick a release that is >=14
 # days old and copy its checksums from:
 #   https://github.com/cli/cli/releases/download/v<VERSION>/gh_<VERSION>_checksums.txt
+#
+# Presence is NOT sufficient: an already-installed gh is accepted only if it meets
+# GH_MIN_VERSION. A distro-packaged gh is routinely several minor versions behind,
+# which is exactly the case the pin exists to fix.
 
 set -euo pipefail
 
@@ -26,8 +30,31 @@ cleanup() {
 # Add common binary locations to PATH
 export PATH="$HOME/.local/bin:$HOME/bin:/usr/local/bin:$PATH"
 
+# Stacked-PR tooling is opt-in. SessionStart runs this script with no arguments, so the
+# default path stays exactly as fast and quiet as before for users who never stack.
+WITH_STACK=0
+for arg in "$@"; do
+    case "$arg" in
+        --with-stack) WITH_STACK=1 ;;
+        -h|--help)
+            echo "Usage: ensure-gh-cli.sh [--with-stack]"
+            echo "  --with-stack  Also install the pinned gh-stack extension and its agent skill"
+            exit 0
+            ;;
+        *) echo "[gh] WARNING: ignoring unknown argument: $arg" ;;
+    esac
+done
+
 # Pinned gh release (>=14 days old per supply-chain cool-off) and its checksums.
-GH_VERSION="2.92.0"
+GH_VERSION="2.97.0"
+
+# Minimum acceptable gh version; an older gh is replaced with the pinned build.
+# The floor tracks the pin because v2.97.0 fixed four advisories, two on paths agents
+# use constantly: `gh auth status` printed part of the token in plaintext for ghs_*,
+# github_pat_* and ghu_* formats (GHSA-cg6r-mpgc-h9mm), and `gh api` / `gh pr diff`
+# emitted externally controlled content without neutralizing terminal escape
+# sequences (GHSA-3m3g-3wcr-px46). It also carries the mature `gh skill` commands.
+GH_MIN_VERSION="2.97.0"
 
 # GitHub hosts to exempt from a session HTTPS proxy when that proxy intercepts
 # GitHub (proxied remote sessions, e.g. Claude Code cloud). Scoped and additive:
@@ -50,22 +77,65 @@ run_bounded() {
     fi
 }
 
-# SHA-256 checksums from gh_2.92.0_checksums.txt, keyed by asset suffix.
+# SHA-256 checksums from gh_2.97.0_checksums.txt, keyed by asset suffix.
 checksum_for() {
     case "$1" in
-        linux_amd64.tar.gz) echo "b57848131bdf0c229cd35e1f2a51aa718199858b2e728410b37e89a428943ec4" ;;
-        linux_arm64.tar.gz) echo "c2248526dd0160c08d3fccca2332c3c1a07c15a78b23978e77735f1b5a18cfee" ;;
-        macOS_amd64.zip)    echo "ae9bb327ab0d91071bdada79f8f14034a2a0f19b0e001835a782eafa519d2af0" ;;
-        macOS_arm64.zip)    echo "b11c54f6bd7d15ed6590475079e5b2fcf36f45d3991a80041b29c9d0cc1f1d07" ;;
+        linux_amd64.tar.gz) echo "a2c9b8497e1f85b1ad0dfcb78b5a622e098801b8e461e459e88e1ee12f018112" ;;
+        linux_arm64.tar.gz) echo "73ea440ecad9c9e284429997ee6f93577bc6f7bc6fba357ef62c53ad8fb641a5" ;;
+        macOS_amd64.zip)    echo "63298c998cc2a924c9e254c6af6a1caad6ece281122687a91f079bc0a462700e" ;;
+        macOS_arm64.zip)    echo "a58b8fd77b417a38f47a0b54d1370c59b0fcdb324ccc9ca002b0998f7c4c999e" ;;
         *) echo "" ;;
     esac
 }
 
-# Check if gh is already installed
+# Compare dotted versions without sort -V (absent on stock macOS) or python.
+# Returns 0 when $1 >= $2. Any prerelease suffix is stripped before comparing, so
+# 2.97.0-rc1 compares equal to 2.97.0; gh ships no prereleases through this path.
+version_ge() {
+    local have="$1" want="$2" i have_part want_part
+    local -a have_parts want_parts
+    IFS='.' read -r -a have_parts <<< "${have%%-*}"
+    IFS='.' read -r -a want_parts <<< "${want%%-*}"
+    for i in 0 1 2; do
+        have_part=$(printf '%s' "${have_parts[i]:-0}" | tr -cd '0-9')
+        want_part=$(printf '%s' "${want_parts[i]:-0}" | tr -cd '0-9')
+        have_part=${have_part:-0}
+        want_part=${want_part:-0}
+        if [ "$((10#$have_part))" -gt "$((10#$want_part))" ]; then return 0; fi
+        if [ "$((10#$have_part))" -lt "$((10#$want_part))" ]; then return 1; fi
+    done
+    return 0
+}
+
+# `gh --version` prints "gh version 2.97.0 (2026-07-31)" on its first line.
+installed_gh_version() {
+    "$1" --version 2>/dev/null | awk 'NR==1 {print $3}'
+}
+
+# Decide whether to install. Presence alone is not enough: a distro-packaged gh is
+# routinely several minor versions behind, and accepting it silently is how the pin
+# fails to apply on exactly the machines that need it.
+NEED_INSTALL=0
 if command -v gh &> /dev/null; then
-    echo "[gh] CLI found at $(which gh)"
+    GH_PATH="$(command -v gh)"
+    GH_CURRENT="$(installed_gh_version "$GH_PATH")"
+    if [ -z "$GH_CURRENT" ]; then
+        echo "[gh] CLI at ${GH_PATH} does not report a usable version; installing pinned v${GH_VERSION}"
+        NEED_INSTALL=1
+    elif version_ge "$GH_CURRENT" "$GH_MIN_VERSION"; then
+        # Newer than the pin is fine and is left alone; this must never downgrade.
+        echo "[gh] CLI found at ${GH_PATH} (v${GH_CURRENT})"
+    else
+        echo "[gh] CLI at ${GH_PATH} is v${GH_CURRENT}, below the required v${GH_MIN_VERSION}"
+        echo "[gh] Installing pinned v${GH_VERSION} to ~/.local/bin (existing gh left in place)"
+        NEED_INSTALL=1
+    fi
 else
     echo "[gh] CLI not found, installing pinned v${GH_VERSION}..."
+    NEED_INSTALL=1
+fi
+
+if [ "$NEED_INSTALL" = "1" ]; then
 
     INSTALL_TMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/tbd-gh.XXXXXX")
     trap cleanup EXIT
@@ -142,11 +212,25 @@ else
     echo "[gh] Installed to $HOME/.local/bin/gh"
 fi
 
-# Verify gh is now in PATH
+# Verify gh is now in PATH. Clear bash's command lookup cache first, otherwise a
+# freshly installed binary can be masked by the path resolved earlier in this shell.
+hash -r 2>/dev/null || true
 if ! command -v gh &> /dev/null; then
     echo "[gh] ERROR: gh CLI still not found in PATH after installation"
     echo "[gh] Ensure ~/.local/bin is in your PATH"
     exit 1
+fi
+
+# Confirm the gh that PATH actually resolves meets the floor. An older gh earlier in
+# PATH would otherwise shadow the pinned build we just installed.
+GH_RESOLVED="$(command -v gh)"
+GH_RESOLVED_VERSION="$(installed_gh_version "$GH_RESOLVED")"
+if [ -n "$GH_RESOLVED_VERSION" ] && ! version_ge "$GH_RESOLVED_VERSION" "$GH_MIN_VERSION"; then
+    echo "[gh] WARNING: PATH resolves gh to ${GH_RESOLVED} (v${GH_RESOLVED_VERSION}),"
+    echo "[gh] which is below the required v${GH_MIN_VERSION}. The pinned build is at"
+    echo "[gh] $HOME/.local/bin/gh — put ~/.local/bin ahead of ${GH_RESOLVED%/gh} in PATH."
+    echo "[gh] Older gh versions leak part of the auth token via 'gh auth status' and do"
+    echo "[gh] not neutralize terminal escape sequences in 'gh api'/'gh pr diff' output."
 fi
 
 # Check authentication status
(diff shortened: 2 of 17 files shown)


jlevy:
## Review: PR #266 (content review, prose and guidance quality)

**Scope**: all 16 files in `gh pr diff 266`. Claims were checked against the live CLI
(`gh` 2.98.0, `gh stack` v0.1.0) rather than read. CI: all 7 checks green at `7249b29`.

**Verdict: request changes.** The design split (GitHub's skill owns mechanics, tbd owns
policy) is the right call and the prose is mostly disciplined. But `--with-stack` does
not actually install the skill, and reports that it did; and the documented stacked
happy path produces draft PRs that `gh stack merge` then refuses. Both are reachable by
following the docs exactly as written.

Findings are ordered by severity. Items marked *(judgment)* are opinion, not defect, and
can be declined without argument.

* * *

### R1 — Blocker: `--with-stack` never installs the skill, and claims it did

`packages/tbd/docs/install/ensure-gh-cli.sh:323` (and the two byte-identical copies)

`gh skill install <repo>` with no skill selector does not install anything. With one
skill in the repo it still wants a selection, so in a non-interactive shell it prints
the available skills and **exits 0**. Reproduced against `github/gh-stack@v0.1.0`:

```
$ gh skill install github/gh-stack --pin v0.1.0 --dir ./t --force < /dev/null
Using ref v0.1.0 (a1b4a3d4)
gh-stack   Manage stacked branches and pull requests with...
$ echo $?; find ./t
0
./t            <- empty, nothing installed
```

Adding the selector fixes it:

```
$ gh skill install github/gh-stack gh-stack --pin v0.1.0 --dir ./t --force
✓ Installed gh-stack (from github/gh-stack@v0.1.0) in ./t
```

Because the pipeline exits 0, line 325 prints `[gh] Installed the gh-stack agent skill`.
The `gh skill list | grep -q gh-stack` guard at line 321 then fails on every subsequent
run, so the no-op repeats silently forever. Net effect: every user who opts in gets the
extension but not the skill, and the entire "the official skill owns the mechanics"
premise of this PR quietly does not hold. Under a PTY the same call is likely to open a
selection prompt and block, which is the exact failure mode the surrounding comment at
line 314 warns about.

**Fix**: pass the skill name, `gh skill install "$GH_STACK_REPO" gh-stack --pin ...`
(or `--all`), and verify the result instead of trusting the exit code, e.g.
`gh skill list --agent "$GH_SKILL_AGENT" --json skillName -q '.[].skillName' | grep -qx gh-stack`.
`gh skill install` returning 0 on a no-op means exit status is not a usable signal here.

Note QA playbook Phase 4 (`gh skill list` | a `gh-stack` row) is exactly the check that
would have caught this. The playbook is well aimed; it just has not been run yet.

### R2 — High: the documented stacked path creates draft PRs, and `gh stack merge` rejects drafts

`create-or-update-pr-simple.md:61`, `create-or-update-pr-with-validation-plan.md:79`,
`stacked-prs.md:38`, `stacked-prs.md:114`

`gh stack submit --help`: "With `--auto`, new PRs are created as **drafts** unless you
pass `--open`." `gh stack merge --help`: "Only basic pull request state is checked before
merging (open and **not a draft**)." GitHub's own skill states this plainly at two places
in its SKILL.md (`Push branches + create draft PRs | gh stack submit --auto` /
`Create PRs as ready for review | gh stack submit --auto --open`).

So, following the shortcuts verbatim on a stacked branch:

- Step 6 runs `gh stack submit --auto`, creating drafts.
- Step 9 tells the user "the PR is ready for review". It is not.
- `stacked-prs.md:114`'s `gh stack merge <pr-number> --yes` then refuses the whole stack.

The end-to-end path documented in this PR cannot complete. This is also the one place the
PR inlines a mechanic rather than delegating, and the inlined version drops the flag that
matters.

**Fix**: use `gh stack submit --auto --open` in both PR shortcuts, and in
`stacked-prs.md:38` replace the parenthetical with the load-bearing fact:
`` `gh stack submit --auto --open`; `--auto` alone creates drafts, which `gh stack merge`
will not merge. ``

### R3 — High: `main` is still hardcoded in step 4, immediately after step 1 resolves `$TRUNK`

`create-or-update-pr-simple.md:45-46`, `create-or-update-pr-with-validation-plan.md:45-46`

```
4. Review all commits on this branch since it diverged from main:
   - Run `git log main..HEAD --oneline` to see commits
   - Run `git diff main...HEAD` to see all changes
```

The PR's stated defect ("all four PR shortcuts hardcoded `--base main`") is fixed in step
6 but left in place in step 4, in both files. Two consequences: on a repo whose trunk is
not `main` these commands fail outright; and on a stacked branch they enumerate every
lower layer's commits, so the agent writes a PR body describing the whole stack for a PR
whose diff is one layer. That is the same flattening failure in the prose layer.

**Fix**: `git log $TRUNK..HEAD --oneline` / `git diff $TRUNK...HEAD`, and on the stacked
path scope to the layer below rather than the trunk.

Related, same step 1: `TRUNK=$(gh repo view ...)` needs network and auth, where
`--base main` did not. If it returns empty, step 6 runs `gh pr create --base ""`. Worth
one clause: "if this returns nothing, fall back to `git symbolic-ref
refs/remotes/origin/HEAD` or ask."

### R4 — Medium: "exit 2 means not stacked" is wrong on the machine most users have

`merge-upstream.md:16`, `address-pr-review.md:66`

Verified: `gh stack view --json` exits **2** when not in a stack (good, the docs are
right about that), but exits **1** with `unknown command "stack" for "gh"` when the
extension is absent. Since stacking is opt-in and deliberately not in the SessionStart
hook, *absent is the default state*. An agent told "exit 2 means it is not" and handed
exit 1 has no documented branch.

The two PR shortcuts get this right ("Exit 2 ... or a missing `gh stack` means the normal
path applies"). The two carve-outs above do not.

**Fix**: use the PR shortcuts' phrasing in both places: any non-zero exit, whether 2 or a
missing extension, means the normal path applies.

### R5 — Medium: the install command assumes the Claude Code harness

`stacked-prs.md:31`, `setup-github-cli.md:59`, `setup-github-cli.md:81`,
`setup-github-cli.md:318`

All four hardcode `bash .claude/scripts/ensure-gh-cli.sh`. `tbd setup` also installs
`.codex/ensure-gh-cli.sh` (`setup.ts:514`, asserted by `setup-flows.test.ts:1006`), and
the script itself supports `GH_SKILL_AGENT=codex`. A Codex agent following these docs
runs a path that does not exist in its surface. These are forkable shortcuts shipped to
downstream repos, so the assumption travels.

**Fix**: on first mention, "the ensure script (`.claude/scripts/ensure-gh-cli.sh`, or
`.codex/ensure-gh-cli.sh` under Codex)", then refer to "the ensure script" thereafter.
`setup-github-cli.md:88`'s `GH_SKILL_AGENT` note already anticipates non-Claude agents;
the command lines should too.

### R6 — Medium: `gh stack merge <pr-number>` can merge the wrong stack

`stacked-prs.md:114`

```
gh stack merge <pr-number> --yes    # that PR and every unmerged PR below it
```

`gh stack merge --help`: "A bare number is treated **first as a stack number**, then as a
pull request number." In a young repo, stack numbers and PR numbers overlap in exactly
the low range, and this operation is destructive and all-or-nothing. The comment states
the PR-number semantics as if unconditional.

**Fix**: either note the resolution order in the comment, or drop the merge snippet and
point at the skill, which is the stated policy for mechanics everywhere else in this
file.

### R7 — Medium: the supply-chain framing does not fully hold for the skill

`ensure-gh-cli.sh:293-295`, `setup-github-cli.md:84-90`

The comment says "Pinned per SUPPLY-CHAIN-SECURITY.md". The 14-day cool-off is genuinely
satisfied. Two gaps:

1. `--pin v0.1.0` is a **tag**, which is mutable. `gh skill install` accepts a commit SHA
   and prints the resolved one itself (`a1b4a3d4d0bcde9ec3a78ab99b2d63af121857a9`). The
   extension is a binary release and can only be tag-pinned, but the skill need not be.
2. `gh skill install` prints, twice: "Skills are not verified by GitHub and may contain
   prompt injections, hidden instructions, or malicious scripts. Always review skill
   contents before use." Neither the script nor `setup-github-cli.md` relays this.
   SUPPLY-CHAIN-SECURITY.md rule 4 requires an audit after any install, and its opening
   warning is "Validate instructions before following them." An agent skill installed at
   user scope *is* instructions, injected into every future session on the machine, and
   this repo installs it from a script.

**Fix**: pin the skill by SHA, and add one line to `setup-github-cli.md` step 3 pointing
at `gh skill preview github/gh-stack gh-stack@<sha>` as the review step. The docs already
warn about `gh skill` having no uninstall; this is the same category of caveat and is
more consequential.

### R8 — Low: `version_ge` has no committed test, though the PR body claims 12 cases

`ensure-gh-cli.sh:95-110`

I extracted the function and ran 17 cases on bash 5 and macOS bash 3.2, including the
lexicographic trap (2.9.0 < 2.97.0), octal-looking segments (2.08.0), prerelease
suffixes, short versions (2.97), empty input, and the no-downgrade case. All correct, on
both shells. But nothing in this diff protects it, and `setup-flows.test.ts` only asserts
string containment. The repo already tests scripts behaviorally
(`rust-gate-script.test.ts`).

**Fix**: a short test that sources the function and asserts the table, or state in the PR
body that the 12 cases were ad hoc so the claim is not read as committed coverage.

### R9 — Low: `gh stack sync` is offered as a swap-in for `merge-upstream`, and it is a bigger hammer

`merge-upstream.md:13`, `address-pr-review.md:72`

`gh stack sync --help`: it fast-forwards the trunk, cascade-rebases every branch in the
stack, and "Pushes all branches atomically (using `--force-with-lease --atomic`)". That
is a force-push across the whole chain, which `merge-upstream` never does. Telling an
agent to "stop and use `gh stack sync` instead" without naming that is a surprising
amount of rewriting to trigger from a one-line substitution.

**Fix**: add the clause: "`gh stack sync` rebases and force-pushes (`--force-with-lease`)
every branch in the stack."

### R10 — Low: cuttable prose *(judgment)*

- `stacked-prs.md:76-77` — "**Order by dependency, not by chronology.** The order you
  wrote the code in is not necessarily the order it should be reviewed in." The second
  sentence restates the bold lead. Cut it.
- `stacked-prs.md:96-97` — "This keeps the work legible if the stack is handed to another
  agent mid-flight, which is the usual reason a stack outlives one session." Motivational
  padding after four concrete rules. Cut.
- `stacked-prs.md:81-83` vs `:104` — "Edit the layer that owns the code" is stated in full
  in Layer Discipline and again in the `code-review-and-commit` row of the table, in the
  same document. Keep one; the table row can be a pointer.
- `merge-upstream.md:14-16` — two sentences make the same argument ("kept current by
  rebasing, not by merging" / "a merge commit rewrites the layer's relationship"). One
  suffices.
- `qa/gh-cli-and-stacked-prs.qa.md:209-214` — Phase 7 step 2 diffs the three script
  copies by hand, which `setup-flows.test.ts:968` and `:997` already assert
  automatically. Drop the step and cite the tests.

The cross-file repetition (the `--json` warning, the owning-layer rule) I would **keep**:
each lands in a shortcut an agent may read alone, which is the case for repeating it.

### R11 — Low: the "roughly 400 lines" offer trigger over-fires *(judgment)*

`stacked-prs.md:70`

"Work already crossing roughly 400 lines or several unrelated concerns" sits in the list
of cases where an agent should proactively offer a stack. Line count is a poor proxy for
decomposability: this PR is ~1,400 diff lines across 16 files and correctly should not be
stacked, because it is one concern. The other three bullets are all shaped as dependency
relationships, which is the real signal.

**Fix**: drop the line threshold, or subordinate it: "several unrelated concerns, which a
large diff often but not always indicates."

* * *

Checked and confirmed sound. Please do not churn on these.

- **The three copies of `ensure-gh-cli.sh` are byte-identical.** Same md5
  (`526fe9e5…`), same 15,931 bytes, `diff` clean both ways. Also enforced at setup time by
  `setup-flows.test.ts:968` and `:997` (the latter covers `.codex/` too).
- **`version_ge` is correct.** 17 cases, bash 5 and bash 3.2, detailed in R8. The
  `10#` normalization, the `tr -cd '0-9'` guard, and the `-` suffix strip all behave. No
  change needed to the logic.
- **`gh stack view --json` really does exit 2 when not in a stack**, with
  `✗ current branch "…" is not part of a stack` on stderr. Verified directly.
- **`gh repo view` really has no `--repo` flag** (`unknown flag: --repo`), and
  `gh repo view <repo> --json defaultBranchRef -q .defaultBranchRef.name` returns `main`.
  The parenthetical in step 1 of both PR shortcuts is correct and worth keeping.
- **Stacking is genuinely opt-in.** `.claude/settings.json` SessionStart runs
  `ensure-gh-cli.sh` with no arguments; `WITH_STACK` defaults to 0; nothing outside
  `--with-stack` touches the network for stack tooling. The "suggest once, accept the
  answer, do not raise it again" guidance is well calibrated and should not be softened
  further.
- **`set -euo pipefail` does make the `| sed` pipelines propagate `gh`'s exit status**, so
  the WARNING branches at lines 309 and 327 are reachable. R1 is not a pipefail bug; it is
  `gh skill install` returning 0 on a no-op.
- **Discoverability of stacking is adequate.** An unfamiliar agent meets it via the
  shortcut index, the README table, step 3 of both PR shortcuts (which runs
  unconditionally), and four carve-outs. No further cross-linking needed.
- **The `gh stack view` TUI warning is well placed**, appearing in `stacked-prs.md:37`,
  both PR shortcuts, and the `setup-github-cli.md` quick-reference table. Four is right
  for a hazard that hangs a session.
- **`stacked-prs.md:38`'s "bare `submit` prompts for every PR title"** matches GitHub's
  own skill wording (its rule 2) even though the CLI help describes a single-screen
  editor. Not worth diverging from the upstream authority; R2 is about `--open`, not this.
- **`gh extension list` / `gh skill list` expected output in the checklists** is right
  (real output is tab-separated, renders as the doc shows).

## CI status

All checks passing at `7249b29`: Benchmark, Coverage & Lint, DeepSource Secrets, and Test
on ubuntu (22.12.0 / 24), macos (24), windows (24). No prior human or formal review on
this PR.

jlevy:
Addressed the review findings in `0176d239`. All eleven fixed, none deferred or rebutted.

R1 (blocker): **fixed** — reproduced independently before fixing (0 files installed, exit 0). `--with-stack` now passes the skill name and verifies presence with `gh skill list --agent "$GH_SKILL_AGENT" --json skillName`, because as you noted the exit status is not a usable signal here. My own pre-PR check missed this only because I tested with `--all`, which the script did not use.

R2: **fixed** — `gh stack submit --auto --open` in both PR shortcuts, and `stacked-prs.md:38` now states the load-bearing fact rather than the parenthetical.

R3: **fixed** — step 4 uses `$TRUNK` in both files, with a stacked-branch clause to scope to the layer below. Added your suggested fallback: if the trunk query returns empty, fall back to `git symbolic-ref refs/remotes/origin/HEAD` instead of creating a PR against an empty base.

R4: **fixed** — both carve-outs now say any non-zero exit, whether 2 or a missing `gh stack`, means the normal path applies. You were right that absent is the default state.

R5: **fixed** — the ensure script is introduced as `.claude/scripts/ensure-gh-cli.sh`, or `.codex/ensure-gh-cli.sh` under Codex, then referred to as "the ensure script".

R6: **fixed** — the resolution order is now stated where the merge snippet lives. Kept the snippet: landing a stack is policy-adjacent enough to earn its place, and the caveat is what was missing.

R7: **fixed** — skill pinned by SHA (`a1b4a3d4…`), and gh's own "not verified by GitHub" warning plus the `gh skill preview` review step are relayed by both the script and the docs. Agreed that an agent skill at user scope is instructions, not data.

R8: **fixed** — added `packages/tbd/tests/ensure-gh-cli-script.test.ts`, which extracts the shipped `version_ge` and runs your 12-case table against real bash, plus the pin/floor/checksum invariants and an assertion against the R1 regression. Mutation-checked both ways: injecting a lexicographic compare fails `2.100.0 >= 2.97.0`, and removing the skill selector fails the R1 test. Took the test rather than the weaken-the-claim option.

R9: **fixed** — the force-push (`--force-with-lease`, whole chain) is named at both sites.

R10 (judgment): **fixed** — took all five cuts. Also took your steer to keep the cross-file repetition of the `--json` warning and the owning-layer rule, since each lands in a shortcut that may be read alone.

R11 (judgment): **fixed** — dropped the line threshold in favour of "several unrelated concerns, which a large diff often but not always indicates". The counterexample was well chosen.

Also acted on the observation attached to R1: QA playbook Phase 4 now verifies the skill with `gh skill list --json` and says explicitly that install output is not evidence, since that is the check that would have caught this.

No changes made to anything in your false-positives section.

Beads: `tbd-7d27` (parent) with `tbd-32kx`, `tbd-193r`, `tbd-g8ok`, `tbd-04vx`, `tbd-jaj0`, `tbd-9xqf`, `tbd-qqd4`, `tbd-o2bu`, `tbd-rad8`, `tbd-xyak`, `tbd-9ymm`.

PR #279, opened 2026-09-08
fix: Preserve embedded provider comments during workspace recovery

jlevy:
## Current status (2026-09-14)

- **Merge, release, and format upgrade map** for the whole native comment effort, including safe release on existing repositories and the later f09 activation: [#277](https://github.com/jlevy/tbd/pull/277) (stage A).
- **Not ready to merge yet.** Base #278 (`d2b1b728`); head `0b09e169`, restacked after #280 merged, with a tree identical to the validated `c9c65169`. All checks pass and GitHub reports it mergeable. The stack evidence below predates that restack; #282 and #283 have since been restacked onto this head.
- **Merge gates** (layer bead `tbd-8rnq`, under `tbd-m88s`):
  - `tbd-s3zx`: the comment-preservation postcondition runs on every `extensions` namespace that has a `comments` array, so a merge silently rewrites third-party data in that shape: non-object entries are dropped, same-identity entries collapse, and the order changes. Scope it to provider namespaces.
  - `tbd-apnu`: no test exercises the lineage rule through `tbd sync` or the merge-refs path, and no test checks where a sync conflict's losing namespace goes.
  - `tbd-cskr`: PR279-R1 (High) was fixed in `c9c65169` with an author disposition only; it needs an independent re-review.
- **Release gate** `tbd-ajq2`: on the ordinary `tbd sync` path, conflicts are counted and reported as "preserved in attic" but no attic entry is written (pre-existing on `main`). So an incompatible namespace lost during sync is recoverable from git history only, and the "archive the complete loser" wording below holds for workspace and outbox recovery but not for sync. Decided: sync will always save every conflict to the attic, documented as a recovery store.
- **Mixed-version teams:** the lineage protection holds only when the merging client runs this code. A v0.8.1 merger still unions comments onto a replacement link.

---

## Problem

Workspace and outbox recovery used an older current snapshot as an approximate merge
base. Independent embedded provider comments could be dropped, while the recovery
postcondition could also union comments from an old provider link into a replacement
link. Direct outbox import could then clear the only remaining copy or leave a pending
comment attached to the wrong external issue.

## Change

- union `extensions.<provider>.comments` only when both namespaces carry the same
  nonempty provider issue ID, or when both legacy namespaces omit `id`
- fail closed for different IDs and known/missing, empty, or malformed ID pairs: keep
  the selected namespace and archive the complete loser before any source clearing
- apply the same lineage rule in the ordinary namespace merge and the approximate-base
[...]

PR #287, opened 2026-09-14
fix: stop three silent data-loss paths and one dishonest dry run

jlevy:
Four release-gate defects. The first three are silent data loss live on `main` and absent from v0.8.1; each is reproduced by a failing test before its fix.

## `tbd-s3zx` (P1) — third-party `extensions` data wiped on merge

Shipped in #279. The provider comment union ran on **every** `extensions` namespace carrying a `comments` key, and `extensions` is the documented home for third-party data. `commentsShareLinkLineage` returns true when neither side has an `id` (`file/git.ts:698`), so an unknown namespace qualified, and `unionCommentArrays` drops non-object entries and re-sorts (`lib/comment-union.ts:42`). A third-party `comments: ["a","b"]` came back `[]` on any structured merge — `tbd sync`, rescue, workspace import.

Writing the test surfaced something the bead did not record: the archived conflict's `lost_value` had `comments` stripped as well, so **the discarded side was not recoverable either**.

Fix: identify a comment log by shape, not by key name. New `isCommentLog` requires every entry to carry comment identity (`local_id` or `id`), an `at`, and a `body` — the `CommentEntry` contract. Both call sites (`mergeNamespace`, `preserveNamespaceComments`) now require it on both sides; anything else takes the ordinary LWW path, which preserves the winner verbatim and archives the loser intact.

## `tbd-dmkd` (P0) — corrupted-worktree repair deletes after a failed backup

`repairWorktree` wrapped the backup `cp` in a bare `catch {}` commented "Continue with repair anyway", then ran `rm(..., {recursive: true, force: true})` unconditionally and returned `backedUp: backupPath` for a path that was never written. A corrupted data-sync worktree can hold bead writes that were never synced, so that deletes the only copy and points the caller at an empty directory.

Fix: the backup is a precondition for removal. A failed copy returns `success: false` naming the reason and the untouched worktree path; nothing is removed. Leaving a corrupted worktree in place is recoverable by hand; deleting it is not.

## `tbd-0oz8` (P0) — Beads import overwrites an unrelated issue

`import.ts` resolved an incoming Beads short ID against `existingByShortId` and adopted that issue's internal ID *before* the collision check ran. That map was built from every existing issue with no source-identity check, so an unrelated native issue owning short ID N was matched by an incoming `xx-N` and had its file overwritten. The collision branch that exists to prevent exactly this was unreachable for the destructive case.

Every import writes `extensions.beads.original_id`, so a bead imported before is already matched by `existingByBeadsId`; anything reachable only by short ID is by construction a different issue. Fix: drop the source-blind branch, always mint a new issue for the occupied case, take the replacement short ID from `deriveShortIdFromUlid` rather than the random `generateUniqueShortId` so two clones importing one file agree, and warn unconditionally rather than only under `--verbose` (the bead does not keep the id its source file names).

## `tbd-uygb` (P2) — `tbd sync --dry-run` silently skips the tracker

`tbd integration sync --dry-run` prints the credential error and exits 1; `tbd sync --integrations --dry-run` printed **nothing** and exited 0, because `cli/commands/sync.ts:274` gates the whole tracker surface on `!dryRun`. Silence reads as "the tracker is settled", and this is the command agents run to preview a sync.

Scoped deliberately: it still does not evaluate the tracker — that needs the locked data-sync context a dry run must not open, since opening one repairs the worktree and commits pending state. It now *says* so and points at the command that does. Uses `notice()`, not `info()`: info is verbose-only (documented at `sync.ts:161` for the adjacent notice), so `info()` would have left the default run as silent as before. Running the surface under `--dry-run` stays open as Phase 1b reporting work.

## Validation

- Full gate green: `pnpm run ci` — 2600 tests, 172 files, format + lint + build clean.
- Every fix red first: 5 new cases in `merge-namespaces.test.ts`, 1 in `worktree-health.test.ts`, new `import-short-id-collision.test.ts` driving the built CLI.
- The worktree case was additionally verified red by reverting the fix in place and re-running (`expected true to be false` — the old code reported success after destroying the data).
- Two guard cases pin the provider comment path unchanged, so the #279 behavior this narrows is not regressed.
- No format change: `CURRENT_FORMAT` stays f08, and nothing here alters what is stored.

## Release gate

`tbd-lz1q` loses four blockers. Still open: `tbd-ajq2` (sync claims attic preservation, writes nothing), `tbd-apnu` (#279 lineage tests), `tbd-cskr` (independent re-review of #279's fix), `tbd-tia7` (validate `agent_map` at config load), `tbd-80vz` (delegate resent every push — needs a live Linear workspace to settle one open question). Outside the gate but still blocking a release cut from `main`: `tbd-od0z` (Phase 1a slot round-trip) and `tbd-bdkj` (the Phase 0 branch, three commits with no PR ever opened).

https://claude.ai/code/session_014oXhUN2vYNgbpKYQF1r6r5


jlevy:
## Review: PR #287 — fix: stop three silent data-loss paths and one dishonest dry run

**Scope:** `0a31ef57` vs `main` (`bee1b830`). Full diff read; `comment-union.ts`, `git.ts` (merge + `repairWorktree`), `import.ts`, `sync.ts`, all four test files, plus callers, `id-mapping.ts`, `comment-store.ts`, `output.ts`, and the three specs that describe comment merging. CI green on all 7 checks. Local: `merge-namespaces`, `merge-refs`, `worktree-health` (66 tests) pass. Several claims below were verified by probe (scratch repos, a throwaway vitest file run against `mergeIssues`), noted inline. Local clone is shallow, so pre-#198 history was checked through the GitHub API.

**Verdict: mergeable with fixes.** The three data-loss fixes (third-party `extensions` wipe, corrupted-worktree delete-after-failed-backup, import overwrite) are correct and the tests for them are pinned to the fix. No Blocker. Findings 1–3 should land before merge: #1 leaves the fourth defect open in the mode agents use (`--json`), #2 adds a notice to every `tbd sync --dry-run` in repos with no tracker, and #3 introduces a narrow new instance of the exact bug class this PR removes. The rest are comment accuracy and test strength.

Severity vocabulary is the repo's (`code-review-rules`): Blocker / High / Medium / Low, mapped to Blocking / Should-fix / Nit.

### Findings

1. **Should-fix (High)** — `packages/tbd/src/cli/commands/sync.ts:283-286`. `notice(message)` with no `jsonData` emits nothing under `--json` (`cli/lib/output.ts:405-413`: JSON mode only writes when `jsonData !== undefined`). Verified in a scratch repo: `tbd sync --dry-run --json` prints only `{"dryRun":true,"action":"Would sync repository",...}` on stdout, nothing on stderr, exit 0; `tbd sync --integrations --dry-run --json` prints nothing at all, exit 0. That is the defect `tbd-uygb` describes, still present in the mode agents run. The adjacent notice at `sync.ts:173-181` passes `{ skippedSurfaces: ['integrations'] }` precisely so it survives `--json`, and `tests/cli-sync-surface-honesty.tryscript.md:108-123` pins that stderr-JSONL contract. The comment at `sync.ts:280-282` says "same reasoning as the surface-narrowing notice above" but omits the payload that makes the reasoning hold. The new notice also has no test (no test references its text; the PR's validation section lists tests for the other three fixes only). **Fix:** pass a structured second argument, e.g. `{ skippedSurfaces: ['integrations'], reason: 'dry-run' }`, and add text and `--json` cases to the tryscript next to the existing `--dry-run` ones (lines 76-100).

2. **Should-fix (Medium)** — `sync.ts:279`. The notice fires whenever `syncIntegrations` is true, with no config check. Verified: fresh `tbd init` repo with no integration configured, `tbd sync --dry-run` prints "Dry run does not evaluate external trackers. Use `tbd integration sync --dry-run`...". Every user without a tracker now gets that on every dry run, pointing at a command that does nothing for them. Both neighbours check first: the real run at `sync.ts:295-301` (`integrationsInert(preConfig)`, `posture.runs`) and the surface-narrowing notice at `sync.ts:167-172`. **Fix:** hoist the `readConfig` + `syncFoldPosture` + `integrationsInert` computation above both branches and gate the notice on `!inert`.

3. **Should-fix (Medium)** — `packages/tbd/src/file/git.ts:840-842`. The `?? []` idiom makes `comments: null` pass as an empty log, so this is not "the same shape gate as the namespace merge" (comment at 833-835): in `resolveNamespace` (793-798) `null` disqualifies (`localNs.comments === undefined` is false), here it qualifies. Verified consequence (probe through `mergeIssues`): a third-party namespace `{ myapp: { comments: null } }` identical on base, local, and remote — touched by nobody — with an unrelated top-level field edited, comes out as `{ myapp: { comments: [] } }`. `preserveExtensionComments` runs over every namespace (884-895), lineage passes because neither side has `id` (700-702), `unionCommentArrays(null, null)` is `[]`, `deepEqual(null, [])` is false, so the namespace is rewritten. The pre-PR guard `!Array.isArray(local.comments) && !Array.isArray(remote.comments)` left it alone, so this is new behaviour introduced by this PR, in the class it fixes (YAML `comments:` with no value parses to `null`). It also fails the substantive-equality check at 1223-1224, so the bead is re-versioned on every merge. Secondary inconsistency, same root cause: `{id:'u', comments:null}` vs `{id:'u', comments:[A]}` goes LWW in `resolveNamespace` (archiving remote's comments as `lost_value`), then this postcondition unions them back — the attic records a loss that did not happen. **Fix:** one predicate for both sites, e.g. `function commentLogsUnionable(a, b) { const aLog = isCommentLog(a.comments); const bLog = isCommentLog(b.comments); return (aLog || bLog) && (aLog || a.comments === undefined) && (bLog || b.comments === undefined); }`, used at 795-798 and in place of 840-842; add the null/null and null-vs-log cases to `merge-namespaces.test.ts`.

4. **Should-fix (Medium)** — `packages/tbd/src/cli/commands/import.ts:592-593` (and the PR description). The comment says the derived short ID means "two clones importing the same file independently compute the same replacement instead of diverging". That is not true: `generateInternalId()` (`lib/ids.ts:116-118`) is `makeInternalId(ulid())`, a fresh random ULID per run, so the derived short ID differs per clone. The property actually gained over `generateUniqueShortId` is that the replacement is a function of the bead's own ULID, which is what `resolveDuplicateShortIds` (`file/id-mapping.ts:245-254`) recomputes if two clones' `ids.yml` later collide — so a displaced bead lands where a later repair would put it (modulo the length argument: import passes the default 4, the resolver passes `calculateOptimalLength(...)`). The preserved-short-ID branch at 597-603 is equally non-deterministic across clones and always was. **Fix:** rewrite the comment to state the real property or drop the determinism claim; correct the PR description.

5. **Nit (Low)** — `packages/tbd/tests/merge-refs.test.ts:366-367`. The winner is deterministic: theirs is newer (`updated_at` 2025-01-04 vs 01-03), `mergeBeadAcrossRefs` passes ours as local (`git.ts:2609`), and namespace LWW is `nsLocalTime >= nsRemoteTime` (`git.ts:1204`), so the merged value must be exactly `{ comments: ['a','b','c'], note: 'remote' }`. `expect.arrayContaining(['a','b'])` also passes for `['a','b']` (local won), `['b','a','c']` (re-sorted), and `['a','b','c','x']`; the comment on 365 says "nothing is dropped or re-sorted" but the assertion cannot see a re-sort. **Fix:** `expect(myapp).toEqual({ comments: ['a', 'b', 'c'], note: 'remote' })`.

6. **Nit (Low)** — `packages/tbd/tests/merge-namespaces.test.ts:248-266`. Both "guard" cases are one-side edits: local equals base, so `resolveNamespace` returns the remote edit at 757-759 before `bothSidesAreLogs` is evaluated; the union they observe is `preserveNamespaceComments` re-deriving the same array. They pass against the pre-PR code and do not pin the gate's positive path. That path is covered by `tests/integrations-comments.test.ts:166` (both sides append), so this is mislabelled redundancy rather than a gap. **Fix:** make both sides change (local `[A, C]` vs remote `[A, B]` → `[A, B, C]`; empty case: local `{ id, comments: [], x: 1 }` vs remote `{ id, comments: [A] }`), or drop the two cases.

7. **Nit (Low)** — `packages/tbd/tests/import-short-id-collision.test.ts:89`. `expect(imported[0].id).not.toBe(native.id)` is tautological: `tbd import` calls `updateConfigPrefixIfNeeded` (`import.ts:544`, `730-745`), which switches the display prefix from `coll` to `src`, so `list` renders every id as `src-…` and can never equal `coll-…`. Verified by hand: after import, `list --all --json` shows `src-agvk | Native issue…` and `src-e3gr | Imported bead…`. The `show(native.id).title` assertion at 82 is what pins the fix (under the old code the native file is overwritten and that fails). **Fix:** compare short-ID suffixes, or assert `all` has two issues with the two titles under distinct ids; optionally assert `extensions.beads.original_id === 'src-<short>'` on the imported bead via `show`. The `split('-').slice(1).join('-')` extraction is fine — prefixes are `[a-z]+` (`ids.ts:203`, `250`).

8. **Nit (Low)** — `git.ts:795`. `bothSidesAreLogs` is true when one side is a log and the other has no `comments` key at all; the name says both must be logs. Folding it into the shared predicate from finding 3 resolves this.

### Out of scope, worth tracking as beads

- `sync.ts:476-492`: `tbd sync --dry-run` still performs the docs sync for real. Verified: fresh repo, 0 files under `.tbd/docs`; after `tbd sync --dry-run --docs`, 96 files written and "✓ Synced docs: +96 doc(s)" printed. `syncDocs` only honours `dryRun` under `--status`. `.tbd/docs` is gitignored cache, so impact is low, but it is the same "dishonest dry run" class this PR names.
- `import.ts:630-639`: `imported++` is counted before `writeIssue`, and a write failure is swallowed unless `--verbose`, so "Import complete / New issues: N" and exit 0 can follow a failed write. Pre-existing, adjacent to the changed lines, same silent-failure class.

### Verified correct

- **`isCommentLog` / `bothSidesAreLogs` truth table.** For (log, log), (log, absent), (absent, log): union, as before. (absent, absent): LWW, as before. (log, non-log) and (non-log, non-log): LWW, which is the intended change. The only divergences from the old `Array.isArray` gate are the `null` case (finding 3) and the malformed-entry case (next bullet).
- **Malformed-entry regression (the trade-off).** Probe confirms the path: local `[A, BAD]` (BAD lacks `body`) newer than remote `[A, B]` → merged `[A, BAD]`; B, a valid concurrent comment on the same provider issue, leaves the live bead and lands only in the attic `lost_value`. Pre-PR the union produced `[A, B, BAD]`. I consider this acceptable, for reasons the PR does not state: the only tbd writer (`integrations/core/comment-store.ts:26-43`, `50-64`) parses every entry with `CommentEntry` on read and rewrites only valid ones, so it would drop BAD on its next write regardless; `isCommentLog` accepts every `CommentEntry`-valid entry (it is strictly looser: no `min(1)`, no datetime format); `CommentEntry`'s shape is unchanged since #206 (`43334c85`, shipped in v0.6.0), so no released version wrote a different shape; and the merge-time union (#279) is in no tag. A malformed entry can therefore only come from a hand edit or a foreign tool. Suggest one sentence in the `isCommentLog` doc comment stating that a single invalid entry disables the union for that merge and why that is safe.
- **`repairWorktree` early return.** Only `doctor.ts:1653` and `doctor.ts:2018` can reach the corrupted branch (`data-context.ts:141` repairs only `missing`/`prunable`); both already turn `success:false` into an error result. Nothing was removed, so skipping the post-`rm` `worktree prune` is correct. `getSharedPaths` is `resolveSharedTbdPaths` (`git.ts:1518-1520`), so the test occupies the right directory; `fs.cp` of a directory onto a regular file does fail. The test would fail on the old code (old path returns `success:true` and deletes `test-data.txt`).
- **Import re-import safety.** `extensions.beads.original_id` has been written by every released import: checked `import.ts` at v0.1.0, v0.2.0, and v0.4.1 through the GitHub API (`original_id` present at lines 167/188, alongside `existingByShortId`). `setup --from-beads` routes through `runImport` (`setup.ts:1765`). Verified by hand that a second import of the same file is matched by `original_id` (`skipped: 1`), and the imported bead carries `extensions.beads.original_id`. The deleted branch had no legacy-compat role.
- **Used-ID set completeness.** `addIdMapping` (`id-mapping.ts:486-489`) mutates `shortToUlid`, and the `Set` at `import.ts:595` is rebuilt from it on every iteration, so two colliding beads in one import cannot derive the same replacement, and short IDs preserved at 602 by earlier iterations are also excluded. `deriveShortIdFromUlid` widens to 5 chars before throwing.
- **No dead code.** `generateUniqueShortId` still has users (`create.ts:118`, `integration-runner.ts:585`); `extractUlidFromInternalId` and `options.verbose` still used in `import.ts`.
- **On-disk compatibility.** No format change; `CURRENT_FORMAT` untouched. YAML round-trip keeps `at` a string (`yaml` default 1.2 schema, `Timestamp = z.string().datetime()`), so `typeof entry.at === 'string'` holds for persisted logs. Mixed-version risk is nil for released versions: v0.8.1 has no comment union to disagree with.
- **Red-before-green.** By trace, the three `merge-namespaces` third-party cases fail on the old code (union empties the array to `[]`), and the import CLI test fails on the old code (native file overwritten, `show(native.id).title` changes).
- **Design.** Shape-based gating is the right call over key-name gating; no in-repo helper already does this. Accepted limitation: a third-party namespace that happens to use the exact `{id|local_id, at, body}` entry shape will still be unioned. The spec (`plan-2026-08-10-external-tracker-integrations.md` §Comments) describes union-by-ID within the provider namespace and does not need to change; a one-line mention of the shape gate there is optional.

---

jlevy:
Addressed the review findings in `d877f53c`:

- **R1: fixed** — the notice now passes `{ skippedSurfaces: ['integrations'], reason: 'dry-run' }`, so it survives `--json`. You were right and the original was worse than wrong: the comment claimed "same reasoning as the surface-narrowing notice above" while omitting the payload that makes the reasoning hold. Verified in a scratch repo (linear enabled, `on_tbd_sync: guarded`): text mode prints it; `--json` puts the JSONL row on stderr with stdout clean. Test gap closed — 5 new cases in `cli-sync-surface-honesty.tryscript.md`, verified red by removing the payload (20 passed / 1 failed).
- **R2: fixed** — the dry-run branch now computes `inert` from config before emitting, matching both neighbours. This repository is the case in point: `.tbd/config.yml` has `on_tbd_sync: off`, so `tbd sync` never touches the tracker here, yet it was printing the notice on every dry run.
- **R3: fixed** — this was the important one. Both sites now call one shared `commentLogsUnionable(left, right)`; the `?? []` idiom is gone. Reproduced your probe exactly before fixing: an untouched `{ myapp: { comments: null } }` came out `{ comments: [] }`. Your secondary point reproduced too — the second new test failed with `expected [ {...} ] to be null`, confirming the merge archived the losing comments and the postcondition unioned them back, so the attic recorded a loss that had not happened.
- **R4: fixed** — comment rewritten. The determinism claim was false; `generateInternalId` mints a fresh random ULID per run. It now states the real property: agreement with `resolveDuplicateShortIds` if two clones' `ids.yml` later collide. PR description corrected.
- **R5: fixed** — exact `toEqual({ comments: ['a','b','c'], note: 'remote' })`. Your determinism reasoning checks out, and the weak assertion could not see the re-sort its own comment claimed to test.
- **R6: fixed** (kept, not dropped) — both cases are now genuinely two-sided: `[A,C]` vs `[A,B]` from base `[A]`, and a sibling-field edit against a first-comment append. Kept because the gate's positive path deserves coverage in the file that tests the gate, with a comment recording why both sides must differ.
- **R7: fixed** — verified the prefix switch first (`import.ts:544` → `updateConfigPrefixIfNeeded`), then replaced the comparison with short-ID suffixes plus `extensions.beads.original_id`, and asserted the repository holds exactly two issues.
- **R8: fixed** — folded into R3's shared predicate, as you suggested.

Also took the suggestion from your "Verified correct" section: `isCommentLog` now documents that one malformed entry disables the union for that namespace, and why that is safe (the comment store validates every entry on read and rewrites only valid ones, and this predicate is strictly looser than `CommentEntry`).

**Out of scope, filed rather than fixed** — both reproduced here first:

- `tbd-3nzb` — `tbd sync --dry-run` performs the docs sync for real. Hit it by accident while verifying R2: a plain `--dry-run` in a scratch repo printed `✓ Synced docs: +96 doc(s)`.
- `tbd-gju9` — `import.ts` counts `imported++` before `writeIssue` and swallows write failures unless `--verbose`.

Full gate green on the result: 2604 tests, 172 files, format + lint + build clean.

Tracking: parent `tbd-8v20`, one child per finding (`tbd-y37d`, `tbd-bhiy`, `tbd-luwz`, `tbd-6aew`, `tbd-q3zs`, `tbd-c18m`, `tbd-sbhg`, `tbd-d2hp`), all closed.

---

jlevy:
CI on `d877f53c`: 6 of 7 green. **Test (windows-latest, Node 24)** failed, and it is not this PR's.

```
packages/tbd/tests/rust-lint-cost-script.test.ts:75
Error: EPERM, Permission denied: \\?\C:\Users\RUNNER~1\AppData\Local\Temp\fake-cargo-QL6NLq
Serialized Error: { errno: 1, code: 'EPERM', path: '...fake-cargo-QL6NLq', syscall: 'rm' }
```

That file is not in this PR's diff, which is `comment-union.ts`, `git.ts`, `import.ts`, `sync.ts` and their four test files.

The mechanism, rather than calling it flaky: the test copies `process.execPath` — `node.exe` — into a temp directory as a fake `cargo`, spawns it, and its `afterEach` does `rmSync(dir, { recursive: true, force: true })`. Windows keeps a running or recently-run executable's image locked, so the unlink returns `EPERM`, and `force: true` does not cover that — it suppresses `ENOENT` only. It is intermittent because it depends on whether the image is still mapped when `afterEach` runs: the same job passed on `0a31ef57` and failed on `d877f53c` with no change to that file or anything it imports.

Same class as `tbd-vugg` (lockfile EPERM on Windows), different file and different lock source. No fix for it exists to port, so per the repo's rules I am not widening this PR with one; filed as **`tbd-nkpl`** with a proposed patch — retry `rmSync` with `maxRetries`/`retryDelay` (Node supports both for `EPERM`/`EBUSY` on Windows) rather than broadening `force`, so a genuine leak still surfaces.

Re-running that one job now, which is the single re-run the rules allow, to confirm it is not deterministic.

I could not read the job log directly from this session — the egress policy blocks `productionresultssa5.blob.core.windows.net`, where Actions stores job logs. The diagnosis above comes from the check-run annotations API, which was reachable.

---

PR #288, opened 2026-09-14
Stability sprint: f08 contract tests and sync attic entries

jlevy:
Continues the stability sprint after #286, #287 and #268. Work is tracked in beads under
the sprint epic; this PR carries the f08 compatibility contract tests and the sync-path
fixes that depend on them.

### f08 contract T4 — cross-version coexistence (`tbd-9fpp`)

The packed upgrade gate proved four *sequential* claims: a repository created by a
published baseline, upgraded once, still works. Nothing covered the *concurrent* case,
which is the one a mixed-version team is actually in — nobody upgrades every machine at
the same moment.

`validateCrossVersionCoexistence` is a fifth scenario in
`packages/tbd/scripts/validate-upgrade-package.mjs`. It puts a clone on the oldest
published f08 release and a clone on the candidate against one bare remote, and asserts:

- the baseline reads candidate-written beads from a plain `git clone` with no setup step
  — its first `tbd sync` provisions the data worktree itself;
- the baseline lists and shows a candidate-written attic entry with its lost value
  intact, and **restores** a candidate-written entry back onto the issue;
- the baseline creates and pushes a bead, the candidate merges it, and neither side's
  bead is lost;
- an `extensions` namespace the older client does not understand comes back unchanged,
  and a link record and journaled intent under `bridge/` survive its merge and push byte
  for byte — checked both on the shared branch and in the candidate's checkout.

The attic assertions are the precondition for `tbd-ajq2` (making `tbd sync` write attic
entries): that change emits a format older clients must already read, so T4 gates it
rather than depending on it.

Two workarounds the scenario needs, both documented at the call site:

- `tbd attic restore` writes back only text fields, and no path on this build produces a
  conflict on one — the workspace importer passes the older side as the merge base, so
  one side always equals base and last-writer-wins has nothing to archive. The scenario
  seeds one entry by hand and has the *candidate* restore it; that restore produces the
  entry the baseline is actually tested against, through the same writer
  (`writeAtticEntryFile`) every other entry goes through.
- No CLI command sets arbitrary `extensions` or bridge state — an integration does — so
  those fixtures are written to disk directly.

Not asserted, deliberately: whether the older *reader* accepts candidate-written intents.
`listIntentFiles` is reachable only through the credentialed sync path, so that needs
Linear or the mock server and belongs with T3 (`tbd-s4kb`). Preservation is covered here;
interpretation is not.

## Verification

- `pnpm --filter get-tbd qa:upgrade-package` — all five scenarios green.
- Both new assertion groups were run against a deliberately wrong expectation and failed
  as intended before being restored (branch-file presence, and the extensions round
  trip).
- `pnpm run ci` — 172 files, 2604 tests, format, lint, typecheck, build all green.

## Also in this PR

- `docs/publishing.md` now documents all five scenarios and the upgrade-vs-coexistence
  distinction.
- The sprint spec's T4 row and checklist are updated.
- The data-sync worktree lookup is factored out of its two existing copies.

## Known gaps, tracked

- `tbd-z87t` (filed here): `tbd attic list` runs the FIELD and WINNER columns together
  for a field name 14 characters or longer, e.g. `extensions.demo`. Cosmetic; `--json` is
  unaffected.
- `tbd-80vz` and `tbd-u9eg` need a live Linear workspace. `api.linear.app` is outside
  this environment's egress allowlist, so they cannot be run from here.

https://claude.ai/code/session_014oXhUN2vYNgbpKYQF1r6r5

---
_Generated by [Claude Code](https://claude.ai/code/session_014oXhUN2vYNgbpKYQF1r6r5)_


jlevy:
## Independent senior review of PR #288 (head `bb86852b`)

**Scope.** `git diff origin/main...claude/eloquent-brown-l1ua1y` — the six commits `ded79a33..bb86852b`, 20 files (+1764/−122). Reviewed against `review-github-pr` / `review-code` / `code-review-rules`, `general-eng-agent-principles`, `error-handling-rules`, `general-testing-rules`, `backward-compatibility-rules`, `ci-and-gates-rules`. Every claim below was traced in the code; the ones marked *reproduced* were run against the built CLI at HEAD. Severity uses the repo's own vocabulary (Blocker/High/Medium/Low); nothing here is a Blocker.

**Verdict: approve with changes.** The design is right — carrying `winner_source` on the conflict from the branch that decided it, and routing every attic writer through `writeAtticEntryFile`, are the correct fixes, and both tryscripts and the coexistence gate genuinely fail if the corresponding change is reverted. Three Medium items should land before merge (R1–R3); R4 is a docs overstatement that the sprint plan cites as a contract, so it should be corrected in this PR too. The rest are Low.

### Lineage rule verdict (item 4, for `tbd-cskr`)

**The rule as implemented is correct for every path that reaches `mergeIssues`, and the two gate sites agree.** Traced: `resolveNamespace` unions only when `isPlainObject` on both sides **and** `commentLogsUnionable` **and** `commentsShareLinkLineage`; otherwise LWW keeps one namespace verbatim and archives the other. `preserveNamespaceComments` uses the identical `commentLogsUnionable` gate, unions only on shared lineage, and on differing lineage reports the side the resolved value does not match — which dedups against the LWW report through `recordNamespaceConflict`. With an approximate base (workspace import passes the older side as base, so one side reads "unchanged"), the postcondition is what catches the lineage split; I traced both orientations and the label is right in both. A queued comment against provider issue X cannot be delivered to Y through any tbd writer at HEAD: every writer that touches `extensions` also bumps `updated_at` (`integration.ts:611-612,729-730,811-812`; `sync-engine.ts:478-479,514-515,1290-1291,1425-1426,1563-1564,1726-1727`), so two clones always conflict on the `updated_at:` line and the structured merge runs. Third-party namespaces: a `comments` array of strings or of identity-less objects is not a log, so the namespace takes LWW and is kept verbatim; a namespace cannot be emptied. A third-party array whose entries happen to have `id`/`at`/`body` is unioned (dedup by identity, sorted by `at`) — that is the documented shape-based trade and I found no way to make it drop an entry that has a unique identity.

Two caveats, both filed below: **R2** — enforcement depends on git reporting a *textual* conflict, which is guaranteed only by that `updated_at` side effect, not by the merge; a hand edit or foreign tool (the very case `comment-union.ts` says the shape test exists for) bypasses it, and I reproduced a comment landing on the wrong issue that way. **R1** — for a namespace whose `id` is not a non-empty string, the postcondition's fallback writes a *wrong* attic label (not a wrong merge).

### Findings

**R1 — Medium — defensive fallback writes a spurious, inverted attic entry for namespaces with a non-string or empty `id`.**
`packages/tbd/src/file/git.ts:876-887`. `commentsShareLinkLineage` returns false for `id: 42`, `id: ''`, `id: null` even when both sides are identical, so with any comment-shaped `comments` array (`[]` qualifies) the postcondition reaches the fallback and reports `localValue` as lost with `winner_source: 'remote'`. *Reproduced* with `mergeIssues` directly:
- base/local/remote `foo: {id: 42, other: base|L|R, comments: []}`, local newer → merged keeps `L`; conflicts: `{lost: R, winner: L, winner_source: local}` (correct) **and** `{lost: L, winner: L, winner_source: remote}` (the value it names as lost is the value that was kept).
- `linear: {id: '', comments: []}` unchanged on both sides, only `title` differs → an `extensions.linear` entry is written with `winner_source: remote` for a namespace nobody edited.
Before this PR these were phantom `ConflictEntry`s; after `413d7e69` every `tbd sync` merge of such a bead writes them as files and counts them in "N conflicts resolved, archived in the attic". Reachable by a third-party namespace using `id` + `comments`, or a hand-edited/legacy provider namespace.
**Fix:** never report a side as lost when its value *is* the resolved value: in the else branch guard each `onConflict` with `!deepEqual(localValue, resolvedValue)` / `!deepEqual(remoteValue, resolvedValue)` (keep the existing `!deepEqual(localValue, remoteValue)` dedup). Add the `id: 42` and `id: ''` cases to `tests/merge-namespaces.test.ts`.

**R2 — Medium — the lineage rule is enforced only when git reports a text conflict; a two-sided `extensions` edit that leaves `updated_at`/`version` untouched is text-merged and transplants the comment.**
`packages/tbd/src/cli/commands/sync.ts:890-965` (clean-merge branch runs no structured merge; only `--diff-filter=U` beads reach `mergeBeadAcrossRefs`). *Reproduced*: two clones, base namespace `{id: issue-X, key, url, state, comments: []}`; clone B adds a queued comment under `comments:`; clone A rewrites `id`/`key` to issue-Y; neither touches `updated_at`. `tbd sync` reports `sent 1 updated, received 1 updated` with no conflict, the bead is `{id: issue-Y, …, comments: [queued against issue-X]}`, attic empty. Not reachable through tbd's own commands today (every writer bumps `updated_at`), but the invariant lives in the writers, not in the merge, and the module docs explicitly anticipate hand edits and foreign tools. `tests/cli-sync-relink-pending-comment.tryscript.md` sets `updated_at` explicitly on both sides, so it cannot see this.
**Fix (pick one):** (a) write `.tbd/data-sync/issues/.gitattributes` with `*.md merge=binary` the same way `mappings/.gitattributes` is written (`sync.ts` ~1240), so any bead changed differently on both sides is a conflict and always goes through `mergeBeadAcrossRefs`; `mergeBeadAcrossRefs` is present in v0.7.0 (checked `merge-base --is-ancestor`), and the coexistence gate would prove the older client handles it. (b) In the clean-merge branch, run the structured merge for beads changed on both sides relative to the merge base. Either way, state in `tbd-design.md` §3.4 what currently makes the field-level path fire.

**R3 — Medium — sync claims "archived in the attic" even when archiving failed; the failure is a warning with exit 0.**
`packages/tbd/src/cli/commands/sync.ts:872-887` catches every `saveConflictToAttic` error and continues; `packages/tbd/src/lib/sync-summary.ts:103-108` appends "archived in the attic" unconditionally; `pushChanges` (`sync.ts:829-831`) still says "preserved in attic". The per-entry "Archived …" line is `output.info`, which is hidden unless `--verbose` (`output.ts:419-420`), so at default verbosity the user sees only the summary and — if it fired — the warning, and the exit code says success. `error-handling-rules` principles 2 and 6. Note the merge is *not yet committed* at this point (archiving runs before `git add -A`), so throwing here leaves the same in-progress-merge state the conflict-marker check already leaves.
**Fix:** count archive failures; when any occurred, say `N conflicts resolved, M NOT archived (see warnings)` and fold it into the non-zero surface rollup the way `integrationFailure` is — or throw a `SyncError` like the marker check does. Either is defensible; silently reporting success is not.

**R4 — Medium (docs) — two new documentation claims are false as written.**
- `packages/tbd/docs/tbd-design.md` §3.5 "Archiving": "Every strategy above that discards a value records it in the attic." `immutable` retains base and drops both edits (`git.ts:1122-1124`), `union_by_key` drops remote's same-key entry (`unionArraysByKey`, local-first, no report), `max` and `min_timestamp` drop the other value — none record anything. The sprint plan cites §3.5 as the contract, so this matters.
- `packages/tbd/docs/tbd-docs.md` "attic": "Every path that resolves a conflict writes the same entry format to the same place: `tbd sync`, `tbd save --workspace`, and `tbd import --workspace` … They travel on the sync branch." `saveToWorkspace` writes to `<workspace>/attic` under `.tbd/workspaces/` (`workspace.ts:228`) — on `main`, not the sync branch — and `tbd attic list/show/restore` read only the data-sync attic (`resolveAtticDir`, `paths.ts:680-686`), so a workspace-save conflict is invisible to every `tbd attic` command. The same sentence appears in `tbd-design.md` §3.5.
**Fix:** "`lww` and `namespace_merge` archive the loser; `immutable`, `max`, `min_timestamp`, `union_by_key` do not" (or make them), and scope the "same place / every clone" sentence to `tbd sync` and `tbd import --workspace`.

**R5 — Low — attic entry fidelity (pre-existing, but this PR moves the writer and makes sync produce these entries on the main path).**
`packages/tbd/src/file/attic-entry.ts:64-69`: `context.local_updated_at` and `remote_updated_at` are both `now()`, so `tbd attic show` prints "Local updated: just now / Remote updated: just now" for every sync conflict; `mergeIssues` has both issues and could put the real timestamps on `ConflictEntry` (the integration-runner writer at `integration-runner.ts:643-655` already records real ones). `git.ts:1035-1046` (`whole_issue`, local-created-first branch) passes `remote.version, local.version` into the `localVersion, remoteVersion` slots — swapped relative to the sibling branch. `lost_value` encodes a nullish loser as `''` while the restore writer (`attic.ts:72`) and integration-runner encode `null` as `'null'`, so restoring a "description cleared" entry sets `description: ''` instead of clearing it. Two conflicts on the same issue+field in one merge (only the R1 fallback produces that) can collide on a same-millisecond filename and overwrite.
**Fix:** carry `local_updated_at`/`remote_updated_at` on `ConflictEntry`; fix the argument order; use `JSON.stringify(conflict.lost_value ?? null)`.

**R6 — Low — T1's baseline is `package.json`'s version today, so the config round trip currently never runs against 0.8.1, and the default must be hand-bumped after every release.**
`validate-upgrade-package.mjs:61,1441,1590-1596`: `candidateVersion` is `manifest.version` (= `0.8.1` on this branch), which equals `latestFormatBaseline`, so every CI run on this branch and on `main` falls back to 0.7.0 — the log line says so, which is good, but the sprint spec marks `tbd-3dti` done with "0.8.1 is a second baseline" while no run has yet exercised 0.8.1. After 0.8.2 ships, the constant still says 0.8.1 until someone edits it, and the "newest published" claim in `docs/publishing.md` decays silently. The script already packs published versions from npm.
**Fix:** resolve the default from `npm view get-tbd dist-tags.latest` (keep `TBD_UPGRADE_LATEST_FORMAT_FROM` as the override), and mark T1 done only after a run against 0.8.1. Related: T2's `BASELINE_DROPS_FROM_LINK_RECORD` is 0.7.0's schema, so overriding `TBD_UPGRADE_SAME_FORMAT_FROM` (which `publishing.md` invites) now fails T2 — note it next to the constant.

**R7 — Low — coexistence scenario: stale rationale for the hand-seeded entry, and weak bridge assertions.**
`validate-upgrade-package.mjs` ~L790: "no path on this build produces a conflict on one [text field]" is false at HEAD — `413d7e69` in this same PR makes `tbd sync` archive `title` LWW losers (reproduced: two clones closing with different `--reason` archive `closed_at` and `close_reason`). The seed itself is legitimate (the entry the baseline is tested against is written by the candidate's `attic restore` through `writeAtticEntryFile`), and the `extensions.demo` round trip is not vacuous because the baseline's `attic restore` rewrites that bead through its own serializer. But the link-record/intent "byte for byte" assertions only prove that 0.7.0's sync does not delete files it never opens.
**Fix:** replace the seed with a real two-clone `tbd update --title` conflict so T4 exercises the sync-written entry cross-version; update the comment; describe the bridge assertions as "not deleted", not "survives its merge".

**R8 — Low — `agentMapProblems` wording and scope.**
`registry.ts:70` joins with `' '`; the `team_key` message has no terminal period, so both problems together read "…when Linear is enabled integrations.linear.identity.agent_map.bot must be…". `provider-settings.ts:238-250` takes a `provider` parameter but hard-codes "Linear app user UUID". `GithubIntegrationSchema` spreads `IntegrationProviderBase` (`schemas.ts:979-985`), so `integrations.github.identity.agent_map` is accepted and neither validated nor consumed (no GitHub adapter exists) — inert today, so no check is needed, but the `provider` parameter suggests otherwise. `status.ts:184` remedy "Set the missing value" for a value that is present but malformed. `UUID_RE` is **not** dead (`adapter.ts:233,1534`).
**Fix:** join with `'; '` or end each message with a period; drop the `provider` parameter or make the message provider-neutral; reword the remedy.

### Tests (item 7)

- `tests/agent-map-validation.test.ts`: the `setAgentMap` regex cannot silently no-op — `tbd init` writes no `integrations:` block and ends with `\n` (checked), so the first call appends and later calls replace; a broken append would surface as a YAML parse error in `doctor` and fail the positive `toContain` assertions, not pass. Reverting `685098ae` fails the two positive tests. The negative test is a control only.
- `tests/attic-conflict-archive.test.ts`: the "deleted namespace" case is the one that distinguishes per-conflict from per-bead labeling (a per-bead `updated_at` guess says `local`, the test demands `remote`) — good. The first test's comment says the old link loses "because" lineage differs; it loses because local is newer — with the timestamps swapped the relink would be archived instead. The test is fine; the comment overstates.
- Both tryscripts fail on revert (attic length 0; `grep -c 'archived in the attic'` 0). `grep -c 'conflict'` = 1 holds only because `Merge conflict, attempting file-level resolution` is `info` and hidden by default.
- Ran locally: the 5 relevant vitest files (67 tests) and both tryscripts (14 steps) pass on the built dist.

- `archiveConflicts` prints one `info` line per entry; consider a single summary line listing filenames so `--verbose` output stays readable on a large merge.
- `docs/publishing.md` says the gate "packs the candidate with the release version" but T1's fallback depends on the working-tree version equalling the newest published one; one sentence tying the two together would save the next reader the trace.

### False positives / checked and benign

- `parseAtticFilename` (`attic.ts:31`) uses `([^_]+)` for the field and cannot parse `close_reason`, `parent_id`, … — harmless: entries are loaded from YAML content and the filename is used only for the entity-id filter. Reproduced with a `close_reason` sync conflict; `attic list`/`show` are correct.
- Abandoned merge: `archiveConflicts` runs before `git add -A`, so a later throw (marker check, commit, `assertNoCorruptBeads`) leaves the entries staged; `git merge --abort` (what doctor tells the user to run) deletes staged new files (verified), and a re-run of `tbd sync` commits them together with the resolutions already written — consistent either way.
- Push-retry double writes: each `mergeRemoteIntoSyncBranch` archives only its own merge's conflicts; `fullSync` adds pull-merge and retry-merge conflicts once each.
- `winner_source` in every other branch (`whole_issue` ×2, delete-on-one-side, same-lineage union, LWW, postcondition local/remote) is correct; only the fallback (R1) is wrong.
- Move of `saveConflictToAttic`: filename, key order and `lost_value` encoding are byte-identical to the old code; `AtticEntrySchema.parse` cannot reject anything the old code wrote (`IssueId`, `Timestamp`, non-negative `Version` all hold for merge output).

### Design assessment

Sound. Per-conflict `winner_source` is the only representation that survives the three decision rules, and one writer for every attic path is the right shape. The materially better alternative is R2(a): make the structured merge the *only* way two-sided bead edits are combined, rather than relying on every writer to touch `updated_at`.

### Documentation

R4 (design §3.5 and `tbd-docs.md` attic section), the T1 checkbox in the sprint spec (R6), and the coexistence comment (R7). `sync-failure-recovery.md` and `tbd-sync-troubleshooting.md` match the code (`restore` limited to `title`/`description`/`notes`, refuses structured fields, archives what it replaces).

### CI status

GitHub: all six checks green on `bb86852b` (Ubuntu 22.12/24, macOS, Windows, Coverage & Lint, Benchmark); the Ubuntu/Node 24 job ran `qa:upgrade-package`, so all seven scenarios passed there. Locally: `pnpm run ci` green (174 files, 2612 tests, format/lint/typecheck/build), plus both new tryscripts green on the built dist.

---

jlevy:
## Review addressed — all eight findings fixed in `a0f4d629`

Every finding was reproduced before it was fixed. Two were worse than filed; one report was
partly wrong about the mechanism and I've noted where.

| | Disposition |
|---|---|
| **R1** | **Fixed.** Reproduced both cases exactly. |
| **R2** | **Fixed** — worse than filed, see below. |
| **R3** | **Fixed.** Failures counted and named. |
| **R4** | **Fixed.** Both claims were false. |
| **R5** | **Fixed**, all four parts. |
| **R6** | **Fixed.** Baseline resolved from npm. |
| **R7** | **Fixed.** Hand seed removed entirely. |
| **R8** | **Fixed**, except `UUID_RE` — agreed, not dead, unchanged. |

### R1 — confirmed, including the inverted entry

Reproduced verbatim: `id: 42` on all three sides with local newer produced the correct
conflict **and** a second one whose `lost_value` was the value that had just been kept,
labelled `winner_source: remote`. And `id: ''` unchanged on both sides produced an
`extensions.linear` entry for a namespace nobody edited.

The root cause is slightly different from "a future strategy that synthesizes a third
lineage": these namespaces have no lineage on *any* side, so the fallback fires for
namespaces that are not in conflict at all. The fallback now reports only a side the result
did not keep, and says why in the comment. Three cases added to `merge-namespaces.test.ts`,
including one asserting a genuine same-absent-lineage conflict is still reported, so the
guard cannot swallow a real loss.

### R2 — confirmed, and it corrupts the record, not just the comment

My first attempt did **not** reproduce it: with `id`/`key`/`comments` on consecutive lines
git conflicts and the structured merge runs. Adding `url`, `state` and `synced_at` between
them — the shape a real link record has — reproduced it, and the result is worse than a
transplanted comment:

```
id: issue-Y                              ← A's relink
key: OS-2                                ← A's relink
url: .../issue/OS-1                      ← the OLD issue's url
comments: [queued against issue-X]       ← transplanted onto Y
```

`✓ Synced: sent 1 updated, received 1 updated`, attic empty. That is a namespace no writer
emits, and whether it happens at all depends on how many unchanged lines separate the two
edits. That also undercuts this PR's own claim that every value a sync merge discards is
archived — here one is discarded with no conflict and no entry — so it's in scope rather
than a follow-up.

Took option (a): `issues/.gitattributes` with `*.md merge=binary`, written with the rest of
the data scaffold. Verified before and after on the built CLI — after, the comment stays
off issue-Y and the abandoned link is archived. The attribute rides the sync branch, so an
older client applies it and reaches its own `mergeBeadAcrossRefs`; the coexistence gate
passes unchanged. Three cases added to `cli-sync-relink-pending-comment.tryscript.md`
covering the path with neither side touching `updated_at`, and `tbd-design.md` §3.4 now has
a "What makes the field-level merge run" section stating the mechanism instead of leaving
it to the writers' side effect.

### R3–R8

- **R3** — `archiveConflicts` counts failures; the summary reads `N conflicts resolved, M
  NOT archived — see warnings`, the push line likewise, and the per-entry `info` lines are
  now one summary line rather than one per entry (your non-blocking suggestion). Test added.
- **R4** — corrected. Only `lww` and `namespace_merge` discard anything; the others keep
  both sides or a derivable bound, and that is now stated as the reason rather than a list.
  The workspace-save attic is described as what it is: a record for whoever inspects the
  workspace, which the `tbd attic` commands do not read.
- **R5** — `ConflictEntry` carries both sides' real `updated_at`; the swapped `whole_issue`
  version arguments are fixed (it also passed `remote.id` where its sibling passes
  `local.id`); a nullish loser encodes as `null`, so restoring a cleared field clears it;
  and `writeAtticEntryFile` advances the timestamp by a millisecond rather than overwriting
  an existing entry — an archive that overwrites an archive isn't one.
- **R6** — resolved from `npm view get-tbd dist-tags.latest`, with the env override kept, so
  it stops decaying after each release. The fallback notice now explains that the candidate
  *is* what npm serves as latest. T2's coupling to the same-format baseline is documented
  next to `sameFormatBaseline`. The sprint spec's T1 row already recorded the fallback; I've
  left the checkbox ticked on that basis, say the word if you'd rather it wait for a run
  against a distinct published release.
- **R7** — the hand seed is gone. The restorable entry now comes from a real two-clone
  `tbd update --description` conflict, which the sync-archiving commit in this PR made
  possible; the scenario is strictly more real than the seed it replaces. Stale rationale
  rewritten, and the bridge assertions now say what they prove — the older client does not
  delete files it never opens — with the sweep-shaped failure they do catch named.
- **R8** — every message ends in a period so the join reads as sentences; the wording is
  provider-neutral (`must be an app-user UUID`) with the Linear-only scope explained at the
  function, including why GitHub's accepted-but-unread `agent_map` is deliberately not
  checked; the `status.ts` remedy now says "Correct the named key". `linear-adapter.test.ts`
  now pins the full shared message so the two call sites cannot drift apart. `UUID_RE`
  confirmed still used at `adapter.ts:233,1534` — no change.

### Verification

`pnpm run ci` green (174 files, 2616 tests), `test:tryscript` green (1128 steps),
`qa:upgrade-package` green (all seven scenarios). R1 and R2 each re-run against the built
CLI before and after their fix.

---

jlevy (commented):
## Follow-up Senior Review of PR #288

**Scope:** Audited the eight findings in the
[independent review](https://github.com/jlevy/tbd/pull/288#issuecomment-5671648186)
against `a0f4d629`, then reviewed the subsequent `0da306d5` slot change and compared it
with PR #290. The follow-up diff is `bb86852b..0da306d5` (17 files).
Inspected formal reviews, inline comments, PR comments, and GitHub issues referencing
#288; no additional review document was linked.
This follows the tbd `review-github-pr` and `review-code` shortcuts.
This is a coverage audit, not a replacement for the original review.

**Verdict: changes required.** Five original findings are addressed; R3, R4, and R6
remain partially addressed.
Retaining their original IDs below avoids duplicate tracking.

### Remaining Findings

**R3 — Medium — archive failure still returns success to scripts.**

`packages/tbd/src/cli/commands/sync.ts:896` increments `conflictsNotArchived`, and
`:1397` copies it into the summary.
Neither path records an issue-surface failure or sets a nonzero exit status.
The rollup at `:366` only fails for entries in `failures`. Therefore an attic write
failure followed by a successful merge/push still exits zero.
The new text accurately reports the missing archive to a human, but the machine-readable
half of the original finding remains open.

**Fix:** After completing the safe merge/push work, report the archive failure through
the issue-surface rollup and return a nonzero status for full sync, pull, and push-retry
paths.
Add a failure-injection check asserting the exit status as well as the warning and
summary. This is traced from the code; I did not inject a filesystem failure locally.

**R4 — Medium — the revised design still promises preservation that the merge does not
provide.**

`packages/tbd/docs/tbd-design.md:2724` says `union_by_key` keeps both sides and that the
bounds discarded by `max` and `min_timestamp` remain derivable from the merged issue.
It concludes that only last-writer-wins throws away a value.
In `packages/tbd/src/file/git.ts:669`, `unionArraysByKey` keeps only the first object
for a key, so a differing remote object with the same key is dropped; `immutable` at
`:1165` retains the base instead of either changed input; `max` and `min_timestamp`
retain one bound, not enough information to recover the other.
The workspace attic location portion of R4 is correctly fixed.

**Fix:** State directly that `lww` and `namespace_merge` archive reported losers, while
`immutable`, `union_by_key`, `max`, and `min_timestamp` do not archive their discarded
inputs. Remove the claim that those inputs are all retained or derivable.
No merge behavior change is needed.

**R6 — Low — the latest-release config test still substitutes the older baseline.**

`packages/tbd/scripts/validate-upgrade-package.mjs:1635` switches to
`sameFormatBaseline` whenever the candidate manifest version equals the newest npm
version. The default is now dynamically resolved, which fixes drift, but equality of
version strings does not make an unpublished candidate containing this PR’s changes the
same artifact as the published package.
Consequently the ordinary pre-release branch still does not exercise the latest
published client, which was the other half of R6.

**Fix:** Pack and run the latest published artifact even when its version string matches
the candidate, then record that run as the T1 evidence.
The package extraction already uses separate directories.
If retaining the fallback is intentional, explicitly defer this part of R6 instead of
marking it fixed, and leave the corresponding evidence requirement open.

### Other Original Findings

- **R1: addressed.** Fallback reports only a source value that differs from the resolved
  value, with absent/invalid lineage regression cases.
- **R2: addressed.** Shared `issues/.gitattributes` forces two-sided Markdown changes
  through the structured merge, with unchanged-`updated_at` relink coverage.
- **R5: addressed for the reported cases.** Real source timestamps and correctly ordered
  versions flow into entries; nullish losses serialize as `null`; sequential filename
  collisions advance the timestamp.
- **R7: addressed.** The restorable entry comes from a real two-clone description
  conflict, and bridge assertions describe the narrower preservation claim.
- **R8: addressed.** Config messages have sentence boundaries and provider-neutral
  wording; the remedy handles malformed values; the still-used UUID expression remains.

### Slot Change and Reconciliation

The final slot commit was outside the original review scope.
It correctly selects a slot before a status band for the blocked-bead convergence case,
and its test checks mutation traffic and the resulting Backlog column.
It is a partial version of PR #290’s adapter change.
Merge #288 before #290 and retain #290’s `slotToLinear(slot, { hold, status })`, its
`writesState` label-preservation condition, and the accompanying reconciliation changes.
Taking #288’s side wholesale at the adapter conflict would discard those corrections.
Validate the combined behavior for blocked, paused/deferred, custom-column, and
unchanged-repeat sync cases before merging the second PR.

### Design and Documentation

Per-conflict provenance, a shared attic writer, and a mandatory structured bead merge
fit the existing architecture.
No replacement design is needed for this audit’s findings.
The bounded changes above complete error reporting and make the preservation and
test-evidence claims match the implementation.
The PR description also predates most of the branch’s work and should be refreshed when
the fixes land.

### Suggestions and Checked Non-findings

- Keep both convergence tests until combined validation proves their public guarantees
  overlap; one checks CLI mutation traffic while #290 checks broader reconciliation
  semantics.
- This audit does not claim mixed-version Linear convergence is proved.
  The final commit explicitly leaves it open, and the generic f08 coexistence test does
  not exercise credentialed provider writes.

### CI Status

All seven checks pass on head `0da306d5`: four OS/Node test jobs, Coverage & Lint,
Benchmark, and DeepSource Secrets.
The prior review’s reported local test runs were inspected as evidence, not re-run here.
The isolated audit checkout has no installed dependency tree or built CLI, so this
follow-up used source/control-flow inspection and current GitHub CI results.

jlevy:
## Addressed the follow-up review findings in `3289adfa`

- R3: fixed — `tbd sync` now records an issue-surface failure and exits nonzero after a merge/push whose discarded conflict value could not be archived. The new CLI fixture injects the attic failure, checks the warning and summary, checks exit 1, restores the path, and proves the resolved issue was still pushed.
- R4: fixed — the design now distinguishes strategies that retain all inputs from `immutable`, same-key `union_by_key`, `max`, and `min_timestamp`, whose discarded inputs are not recoverable from the merged issue unless the caller explicitly reports and archives them.
- R6: fixed — the upgrade gate always uses an independently packed npm artifact for the newest published f08 client, even when its manifest version matches the checkout candidate. The candidate remains a separately packed working-tree artifact.

Tracking: `tbd-xx3l`, `tbd-drlh`, and `tbd-9jbc` under `tbd-bnj1`.

Validation on the remediated branch:

- `pnpm --filter get-tbd qa:upgrade-package`: all seven checks passed, including published 0.8.1 rewriting the separately packed 0.8.1 candidate config.
- `pnpm --filter get-tbd test:tryscript`: 1,134 passed, including the new failure-injection fixture.
- Full Vitest run: 2,616 passed, one unrelated setup hook exceeded its suite-load timeout, and that exact file passed 4/4 in isolation.
- Reconciled with merged PR #289 without code conflicts; pre-push format, lint, typecheck, and build passed. Three known load-sensitive timing tests missed their bounds under the hook and then passed 13/13 together in isolation; remote CI is the clean-environment gate.

The Linear slot overlap remains intentionally assigned to PR #290: merge #288 first, then keep #290's carrier-aware adapter implementation and strengthen #288's CLI convergence expectation on the combined branch.

PR #292, opened 2026-09-15
fix: write the bead merge attribute before every sync merge

jlevy:
Fixes tbd-yqq7 (Release 1 gate). Updated after the independent review; the disposition is in a PR comment.

## Problem

`issues/.gitattributes` (`*.md merge=binary`, added in #288) is what sends every two-sided bead change to the structured merge (`mergeBeadAcrossRefs`). Its only writer was `ensureDataSyncScaffold`, which runs only when a worktree is created. A repository whose `tbd-sync` branch was created by 0.8.1 or earlier (this repository's `origin/tbd-sync` included) never got the file, and git takes merge attributes from the merging worktree, so those clones kept line-merging beads. Two quiet edits a few lines apart (a relink on one clone, a queued comment on the other) combined into a namespace no writer emits, reported as a clean sync with nothing archived.

## Fix

- `DATA_SYNC_MERGE_ATTRIBUTES` in `file/git.ts` is now the one list of merge attribute files; the scaffold spreads it.
- New `ensureDataSyncMergeAttributes(worktreePath, publishedRef?)`: writes a missing attribute file (taking the fetched remote branch's committed copy when it already has one, so both sides add the same blob and the merge stays clean), stages it with `--force` (the scaffold's legacy-ignore reason), and commits only those paths once as `tbd sync: add merge attributes`. A file already on disk is left alone; nothing is committed when nothing changed, so there is no commit on later runs and no "Initialize tbd-sync data layout" commit.
- `tbd sync` calls it in two places:
  - in the full sync, replacing the inline `mappings/.gitattributes` block (same position, before the merge), so every full sync publishes the file even from a clone that never merges; an older client applies it only once it reaches that client's branch;
  - at the top of `mergeRemoteIntoSyncBranch`, because a rejected `tbd sync --push` merges there without passing through the full sync.
- Both calls go through `SyncHandler.ensureMergeAttributes`, which raises any failure as a `SyncError`. The full sync's `try` reads every other error as a first-sync fetch failure and carries on, so a plain failure there would skip the merge and still exit 0.
- `tbd-design.md`: the §3.4 claim that the attribute is written with the scaffold and is simply on the sync branch was false for upgraded repositories. The section now covers:
  - how the file gets there: the scaffold's own commit, and sync's commit after a successful fetch and before a push-retry merge;
  - that sync prefers the remote's copy, and that a failure stops the sync;
  - that git reads attributes from the merging worktree;
  - when an older client starts applying the file;
  - that 0.7.x and 0.8.x write no attic entries for sync merge conflicts, so under the attribute a quiet two-sided edit to one multi-line field loses a side into `tbd-sync` history there.

  The Resolution Flow lists the attribute step. The two directory trees and the "guaranteed scaffold" sentence now list `issues/.gitattributes`.
- The helper's docstring states the shared data-sync lock precondition.

One visible string changed: the rare commit that added `mappings/.gitattributes` on sync was `chore: add merge=union for ids.yml` and is now `tbd sync: add merge attributes`.

## f08 compatibility

No format bump, no bead keys, no bridge-record fields, parser unchanged. The only persisted change is the attribute file itself on `tbd-sync`, which fresh repositories have carried since #288.

## T4 (`validateCrossVersionCoexistence`) not extended

T4 does not have the 0.7.0 baseline merge a two-sided bead edit under `merge=binary`. I did not extend it, for two reasons:

1. 0.7.0, 0.7.1, 0.8.0, and 0.8.1 all resolve conflicts the same way: `git diff --name-only --diff-filter=U` over `issues/`, then `mergeBeadAcrossRefs(worktreePath, id, 'HEAD', 'MERGE_HEAD')` (checked with `git show v0.x.y:packages/tbd/src/cli/commands/sync.ts`). That reads blobs from refs, not the working file. A binary conflict leaves the same unmerged index entry as a textual one, with no markers in the working file, so the old client takes the path it already takes for every textual bead conflict.
2. A CLI-driven T4 step would not test the attribute. Every tbd writer bumps `updated_at`, so two-sided CLI edits already conflict textually in an old client with or without it. Only a quiet (hand or foreign-tool) edit tells the two apart, and T4's `patchIssueFrontMatter` reserializes the whole front matter, which makes its own textual conflicts. The candidate-side quiet-edit case is covered by the new tryscript.

## Tests (red, then green)

- `tests/cli-sync-merge-attributes-upgraded-repo.tryscript.md` (new): both clones' sync branches lack the file, as 0.8.1 left them.
  - **Publish without a merge:** Session B, which is not behind, publishes the file from the full sync's own check. The assertion reads `origin/tbd-sync` right after B's sync. Deleting the full-sync call turns it red.
  - **Full-sync merge:** Session A's quiet relink-vs-comment edit reaches the structured merge: the comment is not carried onto the new link, and the abandoned link is archived. A merges a branch that already carries the file, so this also shows git ignores the incoming side's attributes. Later syncs from both clones add no commit and leave the worktree clean.
  - **Rejected `tbd sync --push`:** same outcome through the push-retry merge. With only the merge-time call removed, those assertions failed.
  - **Failure surfaces:** with Session A's worktree `index.lock` held, `tbd sync` exits 1 with `Could not add the tbd-sync merge attributes`, and the next sync merges normally. Before the `SyncError` wrap it exited 0 without merging.
- `tests/data-sync-merge-attributes.test.ts` (new):
  - adds only the missing file, in one commit;
  - a second call writes and commits nothing;
  - a committed file deleted from disk is restored without an empty-commit error (fails without the staged check);
  - takes a published ref's differing copy, so the following merge is clean (an add/add conflict before);
  - falls back to the default when the ref lacks the file or does not exist;
  - does not commit unrelated staged work with it.

Run locally: the new tests; `tests/cli-sync*.tryscript.md`, `cli-format-compatibility`, and `cli-shared-common-dir-worktree`; the sync, scaffold, and rescue vitest files; typecheck; eslint; prettier; flowmark. **The full suite and `qa:upgrade-package` were not run locally; they ran in CI.**

## CHANGELOG

- Fixed: repositories whose sync branch was created by 0.8.1 or earlier now get the bead merge attribute (`issues/.gitattributes`) on their next `tbd sync`, so two-sided bead edits always go through the field-level merge instead of git's line merge.


jlevy (commented):
## Review of #292 at 2ce1620f (base main @ 1238038e)

**Verdict: Approve with nits.** The root cause in tbd-yqq7 is fixed. Every git content merge of `tbd-sync` now has `issues/.gitattributes` in the merging worktree first. The file is committed and published once, and older f08 clients handle the binary conflicts it causes through the same ref-based path they already use. Nothing below blocks merge. Finding 1 is a small fix worth making here.

Scope: read the full diff, `sync.ts` and `file/git.ts` at head, the new tests, `tbd-design.md` §3.4, `validate-upgrade-package.mjs` (T4), and the conflict paths in v0.5.0 through v0.8.1. Nothing was built or run. CI: all 7 checks pass (3 OS test jobs, Coverage & Lint including tryscripts, Benchmark, DeepSource). There are no earlier reviews.

### Findings

**1. Should-fix: in a full sync, a failure in `ensureDataSyncMergeAttributes` silently skips the merge and can still report success.**
`sync.ts:1278` sits inside the full-sync `try` (`sync.ts:1216-1284`). The `catch` at `sync.ts:1285-1294` rethrows only `SyncError` and logs everything else at debug level as "Fetch failed (may be first sync)". The new function throws `GitError` or fs errors (`git.ts:2061`, `:2070`, `:2073`, `:2075`), so any such failure skips STEP 3. By then `summary.received` is already filled (`sync.ts:1260`). With nothing to push, the run prints `Synced: N received` and exits 0 (`sync.ts:1477-1481`), but nothing was merged. The integration fold also runs on the unmerged state, which the comment at `sync.ts:1296-1300` says must not happen. With commits ahead, the push is rejected and the merge-time call at `sync.ts:928` throws again, so that case does fail loudly, just later.
The old inline block had the same position, but it wrapped its commit in `try/catch`, so a commit failure still let the merge run. Now a commit failure skips the merge. This also contradicts the comment at `git.ts:2059-2060` ("a failed write here would abandon the merge it protects"): in this path the merge is abandoned silently. Likely triggers are rare: `git commit --only` refuses a partial commit while `MERGE_HEAD` exists, index.lock contention, or EACCES/ENOSPC.
Confidence: control flow verified by reading. How often it triggers is inferred.
Suggested fix: rethrow as `SyncError` at the call site, e.g. `await ensureDataSyncMergeAttributes(worktreePath).catch((e) => { throw new SyncError(\`Could not write merge attributes: ${e.message}\`); })`, or wrap inside `mergeRemoteIntoSyncBranch` and drop the separate full-sync call's exposure.

**2. Nit (forward compatibility): if the remote already has the attribute file with different content, the next sync hits an add/add conflict and fails hard.**
`ensureDataSyncMergeAttributes` checks only the local worktree (`git.ts:2062`). It commits its default content (`git.ts:2075`) and then the merge runs. Scenario: clone X has a hand-edited `issues/.gitattributes` (for example an extra line) and pushes it. Clone Y, upgraded and never synced with this build, writes and commits the default content, then merges X's branch. The result is an add/add text conflict. The conflict path stages the marker-bearing file (`sync.ts:1130`), and the marker check (`sync.ts:1135-1151`) throws "Cannot commit: 1 file(s) still have merge conflict markers ... This is a bug in tbd sync", leaving the worktree mid-merge. The same trap applies if a later release changes the contents in `DATA_SYNC_MERGE_ATTRIBUTES`: a clone jumping from 0.8.1 straight to that release would conflict with a remote this release already published. Today it takes a hand edit to reach this, hence Nit. (Identical content from two clones merges cleanly, so the common case is fine.)
Confidence: code path verified by reading. The git add/add behavior is standard, not run.
Suggested fix: before writing, check whether `${remote}/${syncBranch}` already has the path (the fetch has already happened at both call sites). If it does, take that blob (`git checkout ${remote}/${syncBranch} -- <path>`) and commit it, so both sides add identical blobs. Or at minimum, record in the `DATA_SYNC_MERGE_ATTRIBUTES` comment that its contents cannot change without a migration.

**3. Nit (test precision): the tryscript never isolates the full-sync call site, and two prose lines are inaccurate.**
`cli-sync-merge-attributes-upgraded-repo.tryscript.md:112` says "the only clone that can put the attribute back is one running the sync-time check". But Session B runs this build, and its `tbd sync` in the next test (`:121`) runs the full-sync check: behind is 0, so the file is written, committed, and pushed before Session A merges. So the assertion at `:169` ("the attribute is now on the published sync branch") passes from B's push, whatever A's sync did. If `sync.ts:1278` were deleted and `sync.ts:928` kept, every assertion would still pass, so the PR's stated reason for the full-sync call ("publishes them, even from a clone that never has to merge") has no regression test. Also, `:131` says "Session A has never had the attribute in its worktree", but A's worktree got it from the scaffold in `before` and it was then `git rm`'d. The state at merge time is still correct. This setup actually tests a stronger case than the prose says: A merges a branch that already carries the file, which shows git ignores the incoming side's attributes.
Confidence: verified by reading.
Suggested fix: right after B's comment sync (`:121`), assert `git show origin/tbd-sync:.tbd/data-sync/issues/.gitattributes` prints `*.md merge=binary`. That pins the non-merging full-sync publish. Then correct the prose at `:112` and `:131`.

**4. Nit (docs): §3.4 has two small inaccuracies and leaves out one fact about older clients.**
- `tbd-design.md:2643-2644`: "Each writes the file only when it is missing and commits it once (`tbd sync: add merge attributes`)". The scaffold commits as `Initialize tbd-sync data layout` or `Initialize tbd-sync branch` (`git.ts:2214-2226`). Only sync uses the quoted message.
- `tbd-design.md:2642`: "on every full sync". The call runs after the fetch (`sync.ts:1278`), so an offline or first-push full sync does not write it. That is harmless, but the sentence overstates it.
- `tbd-design.md:2649-2653`: accurate that 0.7.x/0.8.x reach `mergeBeadAcrossRefs`. Missing: those releases do not write attic entries for values their merge discards (`archiveConflicts` is 413d7e69, which `git tag --contains` places in no release; v0.7.0 and v0.8.1 `sync.ts` never call `saveConflictToAttic`). Once the attribute reaches an old client's branch, a quiet same-field two-sided edit that git used to line-merge (both kept) goes to last-writer-wins there, and the loser survives only in `tbd-sync` history. CLI edits bump `updated_at` and already conflicted, so this affects only hand or foreign-tool edits, the same scope #288 accepted for fresh repositories. One sentence would keep the compatibility story complete.
- Optional: the Resolution Flow list (`tbd-design.md:2672-2689`) could name the attribute check before step 4.

**5. Nit: the lock precondition is not documented.** Its sibling `ensureDataSyncScaffold` states "MUST be called while holding `withSharedDataSyncLock`" (`git.ts:2098`). `ensureDataSyncMergeAttributes` (`git.ts:2041-2054`) writes and commits in the shared worktree with no such note. Both call sites hold the lock (`withDataSyncContext(..., { lock: true })`, `sync.ts:225-262`), so this matters only for future callers.

### Verified OK

- **Q1, every merge path.** The only git content merge of `tbd-sync` in `src` is `sync.ts:944` in `mergeRemoteIntoSyncBranch`. The attribute check runs first at `sync.ts:928`. That function is reached from the full sync (`sync.ts:1282`) and from every `pushWithRetry` merge callback (`sync.ts:1177`), including `--push` (`sync.ts:838`) and the outbox import push (`sync.ts:1615`). `pullChanges` (`sync.ts:718`) and `updateWorktree` (`git.ts:2416`) use `--ff-only`, so no content merge happens. `rescueUnrelatedHistory` (`git.ts:2901`) resets and reconciles at the file layer. Workspace import merges in TypeScript. `pushFreshOrphan` resets. Doctor repair goes through `initWorktree`/`ensureDataSyncScaffold`, which already writes the file, and does not merge. `pushWithRetry` has no other caller. No path is missing.
- **Committed vs working tree.** For a merge, git reads `.gitattributes` from the working tree first, falling back to the index, never from `MERGE_HEAD`. Writing the file is enough for the local merge. Committing it is what carries it to other clones, older ones included, so committing is the right choice. The one-time commit happens before the merge, so the merged result and the push include it.
- **Q2, commit behavior.** One commit, and only when a write changed the index. The `diff --cached` guard avoids an empty commit when a committed file was only deleted from disk. `--only` keeps unrelated staged work out. The full-sync call follows `commitWorktreeChanges`, and the merge-time call follows it too, so the worktree is committed at both points. No leftover dirt: the tryscript checks `status --porcelain` is empty after repeated syncs. A file already present with other content, or `issues/` with other attributes, is left alone by design (see finding 2 for the cross-clone case). A root `.gitattributes` from a legacy snapshot has lower precedence than `issues/.gitattributes`. `--force` handles a legacy ignore of `.tbd/data-sync`. The one visible string change (`chore: add merge=union for ids.yml`) has no remaining references in the repo.
- **Q3.** Both calls run inside the data-sync lock. Line endings: LF content, and git's attribute parser treats a trailing `\r` as whitespace. Git is given forward-slash relative paths and fs calls use `join`, which is correct on Windows. The unit test's Windows skip follows the pattern in 16 other test files. The read-only FS case is finding 1.
- **Q4, f08 contract.** No `CURRENT_FORMAT` change, no bead keys, no bridge-record fields, parser untouched, no generated-script change (rule 8 does not apply; `validate-upgrade-package.mjs` expectations are unaffected, since `.tbd/.gitattributes` in its list is the main-branch file). Old clients: v0.5.0, v0.6.0, v0.7.0, v0.7.1, v0.8.0, and v0.8.1 all resolve unmerged beads with `git diff --name-only --diff-filter=U -- issues` plus `mergeBeadAcrossRefs(worktreePath, id, 'HEAD', 'MERGE_HEAD')`. That reads ours, theirs, and the merge base from blobs, never the working file. A binary conflict leaves the same unmerged index entry as a text conflict, just without markers, so the old client resolves it structurally, stages it with `git add -A`, and passes its marker check. No f08 client lacks this path; pre-f08 releases refuse the newer format. Modify/delete on a bead fails the same way with or without the attribute. `.gitattributes` is filtered out by every old `issues/` reader (`listIssues` keeps only `.md`; doctor's temp-file scan matches only `.tmp`). The only behavior difference for old clients is the unarchived last-writer-wins case in finding 4.
- **Q5, red on main.** On main, B's scaffold adds the file, the test removes it, and neither clone re-adds it. A's worktree lacks it, and the relink (id/key/url) and comment edits sit two unchanged lines apart, so git line-merges them cleanly and `grep -c 'archived in the attic'` prints 0. The `--push` half fails the same way. The unit tests fail on main (missing export), and the restore case fails without the staged guard. The tryscript's steps are sequential with no timing dependence and use unique `../` fixture names, matching the other `cli-sync*` tryscripts. Simulating a 0.8.1 branch with `git rm` gives a different history than a branch that never had the file, but the merge-time property is the same, and in the realistic history the attribute addition merges cleanly.
- **Q7.** `DATA_SYNC_MERGE_ATTRIBUTES` removes the duplicated content between the scaffold and sync, and the writer reuses `writeScaffoldFileIfMissing`. No dead code: the `access` and `atomically` imports are removed from `sync.ts`.

jlevy:
## Review disposition (review of `2ce1620f`)

Every finding verified before acting. New head `a9d3bc71` (one commit on top of `2ce1620f`).

**1. Should-fix: a failure in `ensureDataSyncMergeAttributes` silently skips the merge and can still report success.** — **Fixed** (`a9d3bc71`).
Verified the control flow: both call sites sit inside the full sync's `try` (`sync.ts:1216-1284`), whose `catch` rethrows only `SyncError` and logs everything else at debug level as a first-sync fetch failure, after `summary.received` is already filled. Both calls now go through `SyncHandler.ensureMergeAttributes`, which wraps any failure as `SyncError` (`Could not add the tbd-sync merge attributes before merging: …`).
Red-green: the tryscript's new section removes the attribute again, has Session B publish a bead so Session A is behind with nothing of its own to push, then holds the worktree's `index.lock` while `tbd sync` runs. Before the wrap the run exited 0 without merging; now it exits 1 with that message, and the next sync (lock released) merges Session B's bead.

**2. Nit (forward compatibility): a remote copy with different content causes an add/add conflict.** — **Fixed** (`a9d3bc71`).
`ensureDataSyncMergeAttributes(worktreePath, publishedRef?)` now checks `cat-file -e <ref>:<path>` on the fetched remote branch and, when it is there, `checkout`s that blob instead of writing the default, so both sides add the same content. Both call sites pass `<remote>/<syncBranch>`, which the fetch has already updated (`pushWithRetry` re-fetches before its merge callback). This also covers a future release changing `DATA_SYNC_MERGE_ATTRIBUTES`.
Red-green: `data-sync-merge-attributes.test.ts` gains a case where a published ref carries `*.md merge=binary` plus an extra line. It now takes that copy and the following merge is clean; with the default written it was an add/add conflict. A second case covers a ref that lacks the file and a ref that does not exist, which both fall back to the default.

**3. Nit (test precision): the tryscript never isolates the full-sync call site, and two prose lines are inaccurate.** — **Fixed** (`a9d3bc71`).
Confirmed: Session B (running this build, not behind) publishes the file from the full-sync check, so the old assertion at the end could not fail if that call were deleted. The assertion is now taken right after B's sync, which is exactly the non-merging publish; deleting the full-sync call turns it red. The prose at `:112` and `:131` is corrected: B undoes the scaffold's copy and publishes, and A's worktree lost the attribute in setup while the branch it merges now carries it (which also shows git ignores the incoming side's attributes).

**4. Nit (docs): §3.4 inaccuracies and a missing fact about older clients.** — **Fixed** (`a9d3bc71`).
The section now attributes `tbd sync: add merge attributes` to sync only (the scaffold commits under its own initialization or repair message), says "every full sync whose fetch succeeds", records the remote-copy preference and the fail-loud rule, and adds the older-client consequence: 0.7.x and 0.8.x write no attic entries for sync merge conflicts (checked: no `saveConflictToAttic` in v0.7.0 or v0.8.1 `sync.ts`), so under the attribute a quiet two-sided edit to separate lines of one field reaches their last-writer-wins merge and the losing side is kept only in `tbd-sync` history. The Resolution Flow list names the attribute step before the merge.

**5. Nit: the lock precondition is not documented.** — **Fixed** (`a9d3bc71`). The docstring now states that callers must hold `withSharedDataSyncLock`.

Verified-OK items needed no change. The T4 decision is unchanged and is argued in the PR description.

PR #290, opened 2026-09-15
fix: write the outbound Linear slot so a not-ready pair stops alternating (#265)

jlevy:
## Summary

A linked open bead that `tbd ready` omits computes slot `backlog`, but the old outbound
path sent only the decomposed `status: open`, which Linear maps to Todo.
The adapter now writes the canonical slot, the create path sends it immediately, and an
inbound projection that changes no persisted bead field no longer rewrites the bead.
This fixes the push/pull alternation in #265 without broadening the claim to the other
open defects in that issue.

## Changes

- Write `patch.slot` through the carrier-aware `slotToLinear`, preserving human labels
  and paused, blocked, or deferred carrier labels.
- Send slot, resolution, and hold on create and journal replay.
- Suppress no-op bead writes while still advancing the successful bridge base.
- Make the Linear mock honor `stateId` on create.
- Reconcile #288 first, retain its real-CLI convergence test, and strengthen the
  combined assertion: creation lands in Backlog and runs 2–9 issue zero mutations.
- Extend the maintained Linear QA playbook and shared scenario contract with
  `blocked-slot-create-settle`, including release/PR maintenance rules.
- Reconcile the active stability spec with the behavior proved here and leave the
  mixed-version and reporter-mirror gates explicitly open.

## Review and Reconciliation

The original #290 head received an independent senior engineering review with no
standalone findings.
The review required #288 to merge first, retention of the carrier-aware adapter path and
label-preservation check, and a zero-write assertion in #288’s overlapping CLI test.
This head implements that exact reconciliation.
A follow-up review of the combined head found one Medium live-QA cleanup gap; `664bf89a`
fixes it by rediscovering owned fixtures independently of candidate-local state and
failing cleanup if discovery or archival is incomplete.

Merge order: #289, #288, then #290.

## Validation

Candidate head: `664bf89a`.

- Focused combined suite: 5 files, 147 tests passed, including two cleanup failure-path
  tests.
- Full local CI: 175 files, 2,625 tests passed, 1 intentional skip; formatting,
  Flowmark, type checking, lint contracts, action-pin checks, and build passed.
- The pre-push gate independently repeated the earlier 175-file / 2,623-test pass before
  the cleanup fix; the full local CI result above covers the final fix.
- Live Linear QA on 2026-09-15 against team `OS`, project `tbd`: all 12 scenarios
  passed, independent token-scoped discovery succeeded against the live API, and all
  disposable fixtures were archived.
[...]

PR #291, opened 2026-09-15
docs: correct update-specs-status shortcut guidance

zerone0x:
## Summary
- update the spec-move guidance to include bare filename and `./` sibling links
- recommend `tbd list --all --parent/--spec --count` instead of reading the data-sync worktree directly

Refs #274

## Verification
- `pnpm format:md:check`
- `git diff --check`

PR #293, opened 2026-09-15
fix: push-only mirror writes the reconciler's slot

jlevy:
Fixes tbd-evn3 (Release 1 gate for Linear users; a regression from #290). Updated after the independent review; the disposition is in a PR comment.

## Problem

Since #290 the reconciler writes the computed slot, so open work that is not ready (an open blocker, a future `deferred_until`) goes to Backlog. The push-only mirror (`tbd integration sync --push`, `tbd sync --push --integrations`) still built its patch with `status`, `resolution`, and `hold` and no `slot` (`mirror.ts:264-266`). The Linear adapter falls back to `statusToLinear` when `patch.slot` is absent, and that maps open work to Todo.

The new test reproduced the sequence against the Linear mock:
- `--push` moved a blocked epic and a future-deferred epic from Backlog to Todo.
- The full sync after it left both in Todo. It did not move them back, as the bead had inferred.
- The second `--push` kept them in Todo.

## Fix

- `outboundPosition(bead, ready)` in `integrations/core/slots.ts` returns `{ status, resolution, hold, slot }`, with the slot from `computeSlot`. This is the body of the engine's `positionOf`, moved out of the engine.
- `runSync`'s create path calls it through `positionOf`, so its behavior is unchanged.
- The reconciler's `localViewOf` takes its slot from the same helper, replacing its own `computeSlot` call. The inputs were already the same four, so there is no behavior change; the point is that the three outbound paths cannot drift apart.
- `planMirror` calls it with the ready set it already computes (`readyIssueIds(allIssues, readyAt)`). The mirror and the reconciler now send the same slot for the same bead.

Three behaviors change for `--push`:

- **Open and not ready, on a team with a Backlog state:** these items now go to Backlog instead of Todo. This is the fix.
- **Open and not ready, on a team with no backlog-type state:** these items now get no state write at all. The slot `backlog` names no state there, so the update carries no `stateId` and the item stays in whatever column it is in, even Done. The title, priority, labels, attachments, and description are still written, so the run reports the item as updated. The reconciler has written the same since #290. Reporting a state push that did not resolve belongs to tbd-od0z (1a) and is not changed here.
- **Legacy five-value `status: blocked`:** these items now go to the team's named Blocked state when one exists, instead of the default started state plus the `blocked` label. The reconciler already writes this for such beads.

On a team with a Backlog state, `slotToLinear` and `statusToLinear` resolve to the same state and carrier labels for every other position:
- ready open work;
- held open work;
- legacy `deferred`;
- `in_progress` with or without a hold;
- closed work with each resolution.

What `--push` still does differently from a full sync:

- It matches a full sync's slot, not its column. The mirror passes no `stateId`, while the reconciler replays `refinement_state_id` and writes nothing when the slots agree. So `--push` still moves In Review and custom started columns (In QA, say) to In Progress, and Draft to Backlog or Todo, where a full sync leaves them. The mirror reads no remote refinement; tbd-6md1 and tbd-9tj0 (one planner) address that.
- Pre-existing and unchanged here: the carriers differ for legacy `status: deferred`. The create path and the mirror send `status: deferred` plus the deferred carrier label. The reconciler's pair path rebuilds status and hold from the slot (`decomposeSlot`), so when it writes, it sends `status: open` without the label.

No docs describe which column a push-only run chooses, so there is no doc change.

## f08 compatibility

No format bump, no bead keys, no bridge-record fields, and no intent change: the mirror does not journal, and its patch still carries `status`, `resolution`, and `hold` beside `slot` (rule 4). This changes what tbd writes to Linear (rule 7), in the direction the reconciler has written since #290. A 0.8.1 `--push` still sends Todo for these items; that is the existing T3 (`tbd-s4kb`) scenario and is not changed here.

## Tests (red, then green)

- `integration-slot-convergence.test.ts` › `keeps not-ready linked epics in Backlog when push-only and full syncs alternate`: built CLI against the Linear mock. A blocked epic, a future-deferred epic, and a ready blocker go through full sync, then `--push`, full sync, `--push`. Before the fix, both epics were in Todo after each step. Now they stay in Backlog (the blocker in Todo), and the middle full sync issues no mutations.
- `integrations-mirror.test.ts` › `sends the slot the reconciler computes, so not-ready open work stays in Backlog`: the planned `{ slot, status, resolution, hold }` for every position:
  - ready, blocked, future-deferred, and held open work;
  - legacy `blocked` and `deferred`;
  - `in_progress`, and `in_progress` held blocked or paused;
  - closed work that is completed, canceled, or duplicate.

  Before the fix, `slot` was undefined for all of them.
- `integrations-mirror.test.ts` › `writes no state for not-ready open work on a team without a Backlog state`: the mock team has its backlog state removed and holds a blocked epic in Todo with a human `Bug` label. After `--push`, the update carries no `stateId`, keeps the label, the item stays in Todo, and it is reported as updated. Before the fix, the update sent Todo's `stateId`.

Run locally: those files plus `integrations-sync-engine`, `linear-adapter`, and the other integration suites; typecheck, eslint, and prettier. **The full suite ran in CI, not locally.**

## CHANGELOG

- Fixed: `tbd integration sync --push` writes the same board slot as a full sync, so open work that is blocked or has a future `deferred_until` stays in Backlog instead of moving to Todo. It still passes no workflow state id, so In Review and custom started columns still move to In Progress. On a team with no Backlog state, such work now gets no state write: the item stays where it is but is still reported as updated, which is also what a full sync does in this release.


jlevy (commented):
## Review of #293 at head `e42f0f70`

**Verdict: Approve with nits.** The mirror and the reconciler now compute the same slot from the same helper, and the fix does what tbd-evn3 asks. There is one Should-fix: a test plus a note in the PR description. It needs no code change. CI is green on all 7 checks.

Scope: I read the diff and the surrounding code at the head: `mirror.ts`, `slots.ts`, `sync-engine.ts`, `reconcile.ts`, `linear/adapter.ts`, `linear/mapping.ts`, `issue-selection.ts` and `integration-runner.ts`. I also checked the f08 contract and Phase 1A/1B in the stability-sprint plan and the state-model spec. I ran nothing locally.

### Findings

**S1 (Should-fix). A team without a Backlog state now gets no state write for open work that is not ready, and there is no test for it.**
- Where: `mirror.ts:269` sends `slot: 'backlog'`. `slotToLinear('backlog')` returns `stateType: 'backlog'` with no `stateName` (`mapping.ts:421-429`). In `toInput`, `patch.stateId ?? named?.id ?? meta.stateIdsByType['backlog']` is undefined when the team has no backlog-type state (`adapter.ts:1044`). The update is still sent with title, priority and preserved labels, and `applyMirror` reports the item as updated.
- Before this PR, `statusToLinear('open')` resolved to Todo. So on such a team, `--push` no longer sets the column for these items. For example, if a person moves the item to Done in Linear while the bead is still open, the item stays in Done and the push reports `updated`.
- The reconciler has done the same since #290, so the two paths still agree. But this is a third behavior change that the PR description does not list, and its sentence "for every other position, `slotToLinear` and `statusToLinear` resolve to the same state" only holds for teams that have Backlog. T3 (tbd-s4kb) lists this team shape.
- Suggested fix: add a test that pins the fallback: mock team without Backlog, `--push` on a blocked bead, assert no `stateId` in the input and labels preserved. Name the behavior in the PR description and CHANGELOG. Reporting a push that did not happen belongs to 1a/1b, not here.
- Confidence: verified by reading.

**N1 (Nit). Slot computation still has two call sites, and the new doc comment claims more than the code guarantees.**
- `slots.ts:141-145` says every writer that projects without comparing takes its position from `outboundPosition`, "so no two of them can disagree". The pair path still calls `computeSlot` separately in `localViewOf` (`sync-engine.ts:224-233`). The inputs are the same four, so the slot cannot differ today, but nothing enforces that.
- The carrier fields also differ. The pair path derives `status`/`hold` from `decomposeSlot` (`sync-engine.ts:839-848`). For a legacy `status: deferred` bead it therefore sends `status: open` and no `tbd:deferred` label, while create and the mirror send `deferred` plus the label. This predates the PR.
- Suggested fix: have `localViewOf` use `outboundPosition(bead, ready.has(bead.id)).slot`, or narrow the comment to "slot".

**N2 (Nit). The unit table skips the other behavior change the PR names.**
- `integrations-mirror.test.ts:266` covers ready, blocked, future-deferred, held-open and canceled.
- It does not cover legacy `status: blocked`, which the PR says now goes to the named Blocked state. It also skips `in_progress`, `in_progress` with a hold, closed `completed` and `duplicate`.
- The mock team has no Blocked state, so an end-to-end test would not show the change. A table row asserting `slot: 'blocked'` for `status: blocked` is cheap.

**N3 (Nit). The CHANGELOG line claims more than the change does.**
- "it writes the same board position as a full sync" is true for the slot, not the column. The mirror passes no `stateId`. The reconciler replays `refinement_state_id` (`sync-engine.ts:858-864`) and writes nothing when the slots agree.
- So `--push` still moves In Review and custom started columns (for example In QA) to In Progress, and Draft to Backlog or Todo, where a full sync leaves them. The PR description says this correctly.
- "blocked or deferred" can also be read as the legacy `status: deferred`, which already went to Backlog.
- Suggested wording: "writes the same slot as a full sync, so open work that is blocked or has a future `deferred_until` stays in Backlog".

### Answers to the review questions

- **Same slot for every state:** yes.
  - Both paths call `computeSlot` with `{status, hold, resolution, ready}` and no refinement.
  - Readiness comes from `readyIssueIds` over all issues in both paths: `mirror.ts:213` over `context.allIssues`, which is `listIssues(dataSyncDir)`, and `sync-engine.ts:606`.
  - Parent state is not a readiness input in either path (`issue-selection.ts:82-110`), so "parent-blocked" does not apply.
  - The mirror reads the clock once per provider (`integration-runner.ts:405`) and uses the same `readyIds` for slots and child counts.
  - Note: an open bead with a `delegate` is not ready (`issue-selection.ts:101`), so delegated open work goes to Backlog on both paths.
- **New column drags:** none. `statusToLinear('in_progress')` already resolved `stateIdsByType.started`. The only change for a person's column is that Draft on not-ready work now goes to Backlog instead of Todo.
- **Out of scope, inferred by reading, not run:** the reconciler looks likely to drag a refinement on the second run after pulling it.
  - A pull sets `merged.slot` to `in_review` or `draft` (`reconcile.ts:320-322`), and the base is written from it (`sync-engine.ts:1515`).
  - `localViewOf` never passes a refinement, so the next run sees local `in_progress` ≠ base `in_review` with the remote unchanged, and pushes `in_progress`.
  - The `stateId` replay needs `refinement_slot === pushed slot`, so it does not apply.
  - The engine test at `integrations-sync-engine.test.ts:938` runs only one sync after the move. Phase 1A already asks for that test to run four times. It is worth filing a bead if none exists.
- **f08 rule 4 (journaled patches):** does not apply to the mirror.
  - `applyMirror` calls the adapter directly. Journaling exists only in `runEnabledIntegrations` (`integration-runner.ts:633+`).
  - The engine's journaled `create_issue` still spreads `positionOf` (`sync-engine.ts:1250`), with the same four keys in the same order as before, so it carries `status`/`resolution`/`hold` beside `slot`. 0.8.1 keeps unknown patch keys, and `slot` has been there since #290.
  - Rule 7: this changes what tbd writes to Linear while T3 is still open, so it remains a Release 1 gate, as the PR says.
- **Tests:**
  - By reading, both new tests fail on main. The unit test gets `slot: undefined`. The CLI test gets Todo after the first `--push`.
  - Covered: push, full sync, push with no flip, a blocked epic and a future-deferred epic, and zero writes in the middle full sync.
  - Not covered: a team without Backlog (S1). Bulk-guard thresholds are not covered either, but they need no test here: the mirror's guard counts `plan.creates`/`plan.updates` (`integration-runner.ts:419-421`), which this PR does not change.

### Verified OK
- `outboundPosition` has the same body as the old engine `positionOf`, so the engine's create behavior is unchanged. `computeSlot` is still imported and used in the engine.
- For ready, held-open, legacy `deferred`, `in_progress` with or without a hold, and closed work with each resolution, the mirror's Linear write is unchanged on a team that has Backlog (compared `slotToLinear` against `statusToLinear`).
- No format, bead-key, bridge-record or intent change.

### Joint note with #294
- `git merge-tree --write-tree review/pr293 review/pr294` merges cleanly (tree `e7a8b938`), auto-merging `mirror.ts` and `integrations-mirror.test.ts`. The merged `planMirror` reads coherently: the `outboundPosition` spread and the `publishesDelegate` gate are in separate lines.
- Both PRs also merge cleanly onto current main `adb4ebbd` (#291, docs only).
- One semantic overlap: the bead in #294's mirror test has a delegate, so after both merge its update also writes Backlog. The delegate assertions count only `delegateId`, so they are unaffected.
- No CI run has covered the combined tree.
- Recommended order: merge **#293 first** (P1 Release 1 gate; touches the engine and slots), then #294. #294 needs no textual rebase, but update its branch from main so CI runs on the combined tree before it merges.

jlevy:
## Review disposition (review of `e42f0f70`)

Every finding verified before acting. New head `25b537fd` (one commit on top of `e42f0f70`).

**S1 (Should-fix). A team without a Backlog state gets no state write for not-ready open work, untested and unnamed.** — **Fixed** (`25b537fd`).
Verified: `slotToLinear('backlog')` returns `stateType: 'backlog'` with no `stateName`, and `patch.stateId ?? named?.id ?? meta.stateIdsByType['backlog']` is undefined on such a team, so the update goes out without `stateId` and `applyMirror` still reports the item as updated.
Test: `integrations-mirror.test.ts` › `writes no state for not-ready open work on a team without a Backlog state` removes the mock team's backlog state, holds a blocked epic in Todo with a human `Bug` label, and asserts the update carries no `stateId`, keeps the label, leaves the column, and is reported as updated. Before this PR it sent Todo's `stateId`.
The PR description now lists this as a third behavior change (including the Done example), and the CHANGELOG-ready line names it. Reporting a state push that did not resolve stays with tbd-od0z (1a), as you suggested.

**N1 (Nit). Slot computation still has two call sites, and the comment claims more than the code guarantees.** — **Fixed** (`25b537fd`), both halves.
`localViewOf` now takes its slot from `outboundPosition(bead, ready.has(bead.id)).slot`, so the pair path, the create path and the mirror share one computation. The inputs were already identical, so there is no behavior change. The `slots.ts` comment no longer claims writers "cannot disagree": it says they agree on the slot and that the pair path rebuilds status and hold from the slot, so a legacy `status: deferred` bead goes out as `open` there without the deferred carrier label. That pre-existing difference is also named in the PR description.

**N2 (Nit). The unit table skips positions.** — **Fixed** (`25b537fd`).
The table now covers ready, blocked, future-deferred and held open work; legacy `blocked` and `deferred`; `in_progress` plain, held blocked and held paused; and closed completed, canceled and duplicate. Legacy `blocked` asserts `slot: 'blocked'`, which is the second behavior change the PR names.

**N3 (Nit). The CHANGELOG line claims more than the change does.** — **Fixed** (PR description).
Verified the mirror passes no `stateId` while the reconciler replays `refinement_state_id`. The line now says `--push` writes the same board *slot* as a full sync, that it still passes no workflow state id so In Review and custom started columns still move to In Progress, and what happens on a team without a Backlog state.

**Out-of-scope inference (the reconciler dragging a pulled refinement on the second run).** — **Not changed here**, recorded instead: the reasoning is now in tbd-od0z's notes, including that `integrations-sync-engine.test.ts` runs only one sync after the move and that a multi-run test would confirm or clear it. It is the same "In Review drag" that bead already tracks for Phase 1A, and fixing it would change the reconciler beyond this PR's scope.

**Joint note (merge order).** — **Adopted.** #293 first, then #294 updated from main so CI covers the combined tree; `git merge-tree` of the two heads still merges cleanly. The recommendation is in #294's description.

PR #283, opened 2026-09-09
feat: inventory native comments and reconcile coordination contracts

jlevy:
## Current status (2026-09-14)

- **Merge, release, and format upgrade map** for the whole native comment effort, including safe release on existing repositories and the later f09 activation: [#277](https://github.com/jlevy/tbd/pull/277) (stage A).
- **Code ready; waits on #282** (stack order). Base #282 (`ecd2a682`); head `4a42a6ac`. All checks pass and GitHub reports it mergeable. The earlier CI and review evidence below applies to the unchanged code:
  - The branch was restacked onto #282's new head after #280 merged: `a48a4416` became `175eac50`, patch-identical with an identical tree ([verification](https://github.com/jlevy/tbd/pull/283#issuecomment-5657920440)), so Astra's approval of `a48a4416` covers the code.
  - `4a42a6ac` adds one docs-only commit to the coordination plan: stack landing (`tbd-m88s`), the #279 merge gates, the attic decision, and release notes.
- **Layer 4 of 4**, tracked as `tbd-w3tv` under `tbd-m88s`.
- **Release-compatibility findings for this layer**, each a bead:
  - `tbd-tia7`: with `identity.agent_map` now reaching the adapter, an invalid map (empty name, non-UUID) fails `integration sync`, `setup`, and `link`, where `main` ignored the setting. Validate at config load.
  - `tbd-80vz`: a valid map makes the push-only projection send `delegateId` for every selected linked bead on every run, closed beads included; Linear may start an Agent Session each time.
  - `tbd-ajq2`: the provider-comment passages in `tbd-docs.md` and `tbd-design.md` say a losing namespace is archived, but `tbd sync` writes no attic entry today. Decided: sync will always save to the attic, so these passages become true once that lands.
  - `tbd-xzyh` (`tbd status` still names a nonexistent `tbd setup beads --disable`; pre-existing) and `tbd-af8w` (comment delivery docs omit the `tbd sync` fold).
  - Release notes: `tbd-lz1q`. They cover the `agent_map` behavior, upgrading every clone, AGENTS.md and skill churn between versions, and changed CLI strings.

---

The native-comment phase needs a source-independent view of filesystem, Git-tree, and
staged-index state before sync and recovery can enforce immutable records.
This PR adds that internal inventory and transition foundation, fixes grouped Linear
agent-map resolution, and reconciles the project’s coordination documentation with the
behavior that actually ships.

## Behavior

[...]

PR #294, opened 2026-09-15
fix: push-only mirror writes a delegate only when it changes

jlevy:
Fixes tbd-80vz (Release 1: "fixed or named"). Updated after the independent review; the disposition is in a PR comment.

**Merge order:** merge #293 first, then this PR. The two merge cleanly both ways (checked with `git merge-tree` at both heads), but no CI run has covered the combined tree. After #293 merges, update this branch from main and let CI run before merging this one.

## Problem

Since #283 honors `identity.agent_map`, the push-only projection (`tbd integration sync --push`, `tbd sync --push --integrations`) sent `delegateId` on every run for every selected linked bead whose delegate maps (`mirror.ts:292-294`, `linear/adapter.ts:1009-1016`). The mirror does not compare against Linear. Closed beads were included because `tbd close` keeps `delegate`. A delegate write is what makes Linear start an Agent Session, so every push re-asserted the delegate, including on finished work. The reconcile path never sends or compares delegate.

## Design choice

I implemented the interim fix the bead specifies. tbd-9tj0 ("Port delegate with change detection and skip closed beads") and the actor-axis spec agree with it: only mapped agents are published, and an absent delegate is no opinion, never a clear.

1. **Never for closed beads.** `planMirror` leaves `delegate` off the patch for a closed bead, on creates and updates. It also no longer reports a `delegate` field skip for a closed bead with an unmapped delegate, so the `skippedFields` report shrinks; its shape is unchanged. Not publishing finished work's delegate is policy, not a failure to surface.
2. **Only when it differs from the issue's current delegate.** `LinearAdapter.applyChanges` drops `delegateId` when the issue already has that app user, comparing by app-user id because `agent_map` allows two names for one app user. An issue with no delegate, or with an app user not in `agent_map`, still gets the write.

   The comparison needs the issue's current delegate. `applyChanges` already reads the issue before a state or label write when the patch carries no full label set, and the change reuses that read. With the default `mirror_labels: none`, mirror updates already made that read, so there is no extra request. With `mirror_labels: prefixed` or `all`, the patch carries labels and skips that read, so each linked bead with a mapped delegate costs **one extra issue read per push** (plus label pages past 50).

What this rule does not stop:

- **A delegate removed in Linear is restored.** While the bead still names a mapped agent, the next `--push` restores a delegate that a person removed in Linear (for example to stop the agent) or replaced with an app user not in `agent_map`, and it may start a new Agent Session. It is sent once after each removal, where before it was sent every run.
- **Stopping it:** clear the bead's delegate (`tbd update <id> --delegate ""`).
- **Why no base comparison:** honoring a remote removal would need a stored base, and a new bridge-record field conflicts with f08 rule 3. The question is recorded on tbd-9tj0, which owns the engine port.

Unchanged:
- An unmapped delegate on open work is still reported as skipped.
- A bead with no delegate still never clears Linear's.
- Creates of open beads still send the delegate: a new item has none to compare.

## f08 compatibility

No format, bead-key, bridge-record, intent, or parser change. What tbd writes to Linear only narrows: fewer `delegateId` writes, none for closed beads. A 0.8.1 clone's `--push` still re-sends delegates, closed beads included; release notes should name the minimum version.

## Docs

`tbd-docs.md` (agent_map paragraph):
- It says when the push-only projection writes a delegate, naming both `tbd integration sync --push` and `tbd sync --push --integrations`.
- It says that the projection restores a delegate removed in Linear, and how to stop it.
- It says bare `tbd integration sync` does not apply delegates: the provider reads Linear's delegate, but the reconciler neither compares nor writes it.
- It drops the claim that a mapped delegate is "read back as a bead delegate".

`plan-2026-08-18-actor-axis-and-identity.md`: the Phase 3 checkbox and the Mapping section said inbound delegate mapping shipped. Only the adapter half did. The checkbox is split into a done adapter item and an open sync item, and a dated status note sits under Mapping. The note also records that an unknown app user reads as no delegate without the planned warning.

## Tests (red, then green)

- `integrations-mirror.test.ts` › `publishes a delegate for open work but never for a closed bead`. Before: the closed bead's patch carried `delegate`.
- `integrations-mirror.test.ts` › `sends a delegate once, not on every push (tbd-80vz)`: `applyMirror` three times against the Linear mock with an agent-mapped adapter. Before: 3 `IssueUpdate`s with `delegateId`. After: 1.
- `linear-adapter.test.ts` › `does not rewrite the delegate an item already has (tbd-80vz)`: with and without a state write in the patch. Before: `delegateId` sent both times.
- `linear-adapter.test.ts` › `writes a delegate that differs from the item's current one`: a foreign app user, and no delegate. This is the control; it passed before and after, and it pins the restore-after-removal behavior described above.

Run locally: the adapter, mirror, and integration suites plus `docmap` and `integration-files`; typecheck, eslint, prettier, flowmark. **The full suite ran in CI, not locally.**

**Not verified:** whether live Linear starts a new Agent Session when the same `delegateId` is written again, or when a delegate is restored. That needs a workspace with an installed agent, and neither the QA playbook nor `validate-linear-integration-live.ts` has a delegate or `agent_map` scenario (recorded on tbd-80vz).

## CHANGELOG

- Fixed: `tbd integration sync --push` (and `tbd sync --push --integrations`) no longer re-sends a Linear delegate, which can start an Agent Session, on every run. It writes one only when the issue's delegate differs, and never for closed beads. While a bead names a mapped agent, a delegate removed in Linear is still restored by the next push, and may start a new session; clear the bead's delegate to stop that. Closed beads with an unmapped delegate no longer appear in the skipped-fields report.


jlevy (commented):
## Review of #294 at head `c3370c48`

**Verdict: Approve with nits.** The two changes are correct and scoped to what tbd-80vz asks for: no delegate for closed beads, and no rewrite of a delegate the item already has. There is one Should-fix, which is documentation of a consequence of the chosen rule. CI is green on all 7 checks at `c3370c48`.

Scope: I read the diff and the surrounding code at the head: `mirror.ts`, `linear/adapter.ts` (`applyChanges`, `toInput`, `toCanonical`, `canPushDelegate`), `linear/queries.ts`, `integration-runner.ts`, `sync-engine.ts`, `reconcile.ts`, and every `delegate` reference under `src/`. I also checked the actor-axis spec, the f08 contract in the stability-sprint plan, and tbd-80vz and tbd-9tj0. I ran nothing locally.

### What the rule is
1. `planMirror` leaves `delegate` off the patch when `issue.status === 'closed'`, for creates and updates (`mirror.ts:241`, `:296`).
2. `LinearAdapter.applyChanges` reads the item whenever a mapped delegate is on the patch (`adapter.ts:521-529`). It drops `delegate` when the item's current delegate, mapped back through `agent_map`, has the same lowercased app-user id (`:535-543`).

The comparison is against the live remote value, not a stored base. There is no base to compare against: `BridgeBase` has no delegate, and adding one would be a new bridge-record field (f08 rule 3). So this is the only f08-safe form of change detection available now.

### Findings

**S1 (Should-fix). A delegate removed in Linear is restored by the next `--push`, and the docs and CHANGELOG do not say so.**
- Failure scenario: an open or in-progress bead has `delegate: cyrus`, mapped. A person un-delegates the agent in Linear to stop it, or the app clears its own delegate. `current.delegate` is now null, so `unchangedDelegate` is false and the next `--push` from any clone writes `delegateId` again.
- The adapter comment at `adapter.ts:1008` notes that this can start a new Agent Session. The same happens when someone replaces the delegate with an app user that is not in `agent_map`: `toCanonical` maps it to null (`adapter.ts:950`), and `linear-adapter.test.ts:596` pins the overwrite.
- Before this PR the delegate was sent on every push. After it, the delegate is sent once after each removal. That is a large improvement, but it is still tbd overriding a person's decision in Linear. It matters more than a title overwrite because the write can start an agent.
- The new sentence in `tbd-docs.md:1778` implies this without stating it.
- Suggested fix:
  - Add one sentence to `tbd-docs.md` and the CHANGELOG: while the bead still names the agent, `--push` restores a delegate removed in Linear. To stop publishing it, clear the bead's delegate (`tbd update <id> --delegate ""`).
  - Note on tbd-9tj0 that the engine port must decide how to treat a delegate removed on the remote, since a base field would conflict with rule 3.
- Confidence: the tbd side is verified by reading. Whether Linear starts a new session on re-delegation is not verified, as the PR says.

**N1 (Nit). The PR description overstates that the read comes for free.**
- The description says `applyChanges` "already reads the item before a state or label write, and every mirror update writes state". That read is gated on `patch.labels === undefined` (`adapter.ts:521`).
- With `mirror_labels: prefixed` or `all`, the mirror sends `labels`. The delegate check then adds one issues read per delegated linked bead per push, plus label pages past 50.
- The default `none` is unaffected. The cost is acceptable; the description should say it accurately.

**N2 (Nit). The spec now disagrees with the docs.**
- The corrected `tbd-docs.md` text is right: no path writes a Linear delegate onto a bead. `delegate` does not appear in `sync-engine.ts` or `reconcile.ts`, and `ExternalIssue.delegate` is populated but unused.
- The actor-axis spec still reads as if this shipped end to end. Its Mapping section says "Inbound, a Linear delegate ... maps through `agent_map` in reverse", and Phase 3 is checked `[x] Inbound delegate maps through agent_map in reverse; unknown app users warn and leave the bead unchanged`.
- Suggested fix: add a note to the spec that only the adapter half shipped, or file a bead.

**N3 (Nit). The docs wording is slightly incomplete.**
- `tbd-docs.md:1778` names only `tbd integration sync --push`. `tbd sync --push --integrations` uses the same projection (`sync.ts:340`, `:475` → `runEnabledIntegrationPushes`).
- "neither publishes nor reads delegates" at `:1780`: the adapter does read the delegate; the engine ignores it. "does not apply" would be exact.

### Answers to the review questions
- **Correct per the actor-axis spec:** yes. Only mapped agents are published. `canPushDelegate(null)` returns false (`adapter.ts:336-343`), so removing the bead's delegate never clears Linear's; the spec defines an absent actor as no opinion.
- **New or changed delegates:** still written. An item with no delegate or a foreign one gets the write, covered by the tests at `linear-adapter.test.ts:596` and `integrations-mirror.test.ts:486`. Two agent names for one app user compare equal.
- **First push after upgrade:** needs no base. Items already delegated to the mapped agent are skipped. Items without that delegate get one write, and S1 covers what happens afterwards.
- **Repeated Agent Session starts:** avoided in steady state. `integrations-mirror.test.ts:486` shows 3 pushes producing 1 `delegateId` write.
- **f08:**
  - No bridge-record field is added or reinterpreted. There is no intent, schema, or parser change, and no data a 0.8.1 rewrite strips.
  - Mixed versions: 0.8.1 `--push` still re-sends delegates, closed beads included. The PR notes the minimum version.
  - Rule 9: closed beads with an unmapped delegate no longer add a `skippedFields` entry (`mirror.ts:316-321`). That shrinks the report without changing its shape. Worth one CHANGELOG word.
- **Docs:** accurate apart from N3 and the missing consequence in S1.

### Verified OK
- `ISSUES_BY_ID_QUERY` selects `delegate { id name displayName }` with `includeArchived: true` (`queries.ts:84`), so the comparison sees the real value.
- The reconcile path is unchanged: the engine never sets `patch.delegate`. `spliceDescription` calls `applyChanges({ description })`, which triggers no extra read.
- The closed-bead gate sits in the one patch builder, so creates of closed beads also carry no delegate.
- By reading, the new tests fail on main: 3 writes instead of 1, the closed-bead patch carries `delegate`, and the adapter sends `delegateId` twice.

### Joint note with #293
- `git merge-tree --write-tree review/pr293 review/pr294` merges cleanly (tree `e7a8b938`), auto-merging `mirror.ts` and `integrations-mirror.test.ts`. The merged `planMirror` is coherent.
- Both PRs also merge cleanly onto current main `adb4ebbd` (#291, docs only).
- Semantic overlap: after #293, an open bead with a delegate is not ready (`issue-selection.ts:101`), so its slot is `backlog`. The bead in this PR's `sends a delegate once` test therefore also gets a Backlog state write. The assertions count only `delegateId`, so they are unaffected.
- No CI run has covered the combined tree.
- Recommended order: **#293 first** (P1 Release 1 gate), then this PR. No textual rebase is needed, but update the branch from main after #293 merges so CI runs on the combined tree.

jlevy:
## Review disposition (review of `c3370c48`)

Every finding verified before acting. New head `94c6ff84` (docs and spec only; no code change, so the `c3370c48` test results still stand).

**S1 (Should-fix). A delegate removed in Linear is restored by the next `--push`, undocumented.** — **Fixed** (`94c6ff84`).
Verified: `canPushDelegate` gates on the bead's delegate only, and `unchangedDelegate` compares against the live remote value, so a null or unmapped remote delegate makes the write happen again. `tbd-docs.md` now states it, names the consequence (the write may start a new Agent Session), and gives the way out (`tbd update <id> --delegate ""`). The CHANGELOG-ready line in the PR description says the same. The design question for the engine port, including why a `base.delegate` field would conflict with f08 rule 3, is recorded in tbd-9tj0's notes.

**N1 (Nit). The PR description overstates that the read comes for free.** — **Fixed** (PR description).
Verified: the read is gated on `patch.labels === undefined` (`adapter.ts:521`), and the mirror sends `labels` under `mirror_labels: prefixed` or `all`. The description now says that with label mirroring on, each linked bead with a mapped delegate costs one extra issue read per push, plus label pages past 50, and that the default `none` is unaffected.

**N2 (Nit). The actor-axis spec disagrees with the corrected docs.** — **Fixed** (`94c6ff84`).
Verified: no `delegate` reference in `sync-engine.ts` or `reconcile.ts`, and `toCanonical` maps an unknown app user to null with no warning (the spec claimed a warning). Phase 3's checkbox is split into a done adapter item and an open sync item, and a dated status note under Mapping records exactly what shipped in each direction.

**N3 (Nit). Docs wording incomplete.** — **Fixed** (`94c6ff84`).
Verified `tbd sync --push --integrations` reaches the same projection through `runEnabledIntegrationPushes`. Both commands are named now, and "neither publishes nor reads" is replaced with what is true: the provider reads Linear's delegate, the reconciler ignores it, so bare `tbd integration sync` does not apply delegates in either direction.

**Joint note (merge order).** — **Adopted.** Recommended order is in the PR description: #293 first, then update this branch from main so CI covers the combined tree. `git merge-tree` of the two heads still reports a clean merge.

The rule 9 note (closed beads with an unmapped delegate no longer produce a `skippedFields` entry) is now in the CHANGELOG-ready line.

PR #306, opened 2026-09-16
fix: detect dependency cycles in doctor

jlevy:
## Summary

- detect directed blocker cycles during `tbd doctor` without recursion
- report one deterministic depends-on path per cyclic component using public bead IDs
- preserve the existing clean and orphan-only dependency diagnostics
- document reproduction, diagnosis, and manual repair in the packaged CLI manual

Tracks `tbd-scf0`. A separate pre-existing orphan-repair defect discovered during this work is tracked as `tbd-vv9t`.

## Reproduction

On the base commit, create three beads A, B, and C, then run:

```bash
tbd dep add <A> <B>
tbd dep add <B> <C>
tbd dep add <C> <A>
tbd doctor
```

Before this change, all three beads are blocked but doctor prints `✓ Dependencies`. The dependency check only verifies that each stored target exists; it never traverses the directed graph.

With this branch, doctor exits nonzero and reports a public-ID path such as:

```text
✗ Dependencies - 1 directed cycle(s)
    depends-on cycle: cyc-ban1 -> cyc-5e34 -> cyc-y7aq -> cyc-ban1
    Break each cycle with: tbd dep remove <issue> <depends-on>.
```

## Fix

`findDependencyCycles` builds the user-facing depends-on graph from the inverse stored `blocks` edges, computes strongly connected components with iterative Kosaraju traversals, and returns one deterministic closed path per cyclic component. Missing targets remain the existing orphan-reference concern and are excluded from cycle traversal. `doctor` formats paths through the public ID mapping and treats cycles as hard errors. It does not auto-delete an edge because tbd cannot infer which dependency was unintended.

## Validation

- `pnpm --filter get-tbd exec vitest run tests/issue-dependency-graph.test.ts tests/doctor-sync.test.ts` — 13 passed
- manual built-CLI reproduction in a disposable Git repository — base printed a green Dependencies check; branch prints the public-ID cycle path and exits 1
- `pnpm lint:check` — passed
- `pnpm publint` — passed
- first complete `pnpm run ci` during implementation — 2,683 passed, 1 skipped
- final complete run passed format, type, lint, and build; 2,676 tests passed and 7 unrelated 5-second process/performance tests timed out under concurrent host load. Every timed-out selection passed on isolated rerun, including the remaining import collision test on its single-test rerun. GitHub CI is the clean-host final gate.

Base: `1262b9741d0fe7e75b477339b9a7ea5bda9c7588`
Head: `076e1ade`

---

> [!NOTE]
> **Low Risk**
> Adds read-only diagnostics on the doctor path with no change to dependency write behavior or auto-fix for cycles.
> 
> **Overview**
> **`tbd doctor`** now flags **directed depends-on cycles** in blocker dependencies as **error-level** findings (nonzero exit), not only missing dependency targets. A new graph helper finds strongly connected cyclic components and emits one deterministic closed path per component in CLI order (`A -> B` means A depends on B), with **public short IDs** when `ids.yml` loads and **internal IDs** as a fallback so cycles stay visible even when the mapping is conflicted.
> 
> Orphan references remain **warnings** with **`tbd doctor --fix`**; cycles are **manual repair** via **`tbd dep remove`** (no auto-fix). Docs describe reproduction, semantics, and repair. Unit and integration tests cover multi-node and self-loops, combined cycle+orphan output, and ID formatting edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88a4ddce5df96a612693e208d5bd3b03968b3a85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jlevy (commented):
**Review A** (senior engineering review, round 1) · head `076e1ad` · base `main` at `1262b97`
Reviewer: strong tier, requested `fable` at `xhigh` · Channel: formal review
Coverage: 4 of 4 changed files; skipped none
Tests run: `pnpm --filter get-tbd exec vitest run tests/issue-dependency-graph.test.ts tests/doctor-sync.test.ts` (13 passed); scratch probe of `findDependencyCycles`/`dependencyFinding` over self-loop, two disjoint cycles, cycle plus orphan, missing target closing a would-be cycle, SCC with a self-loop on its smallest node, duplicate edges, and a 30,000-node chain and ring (all as designed; ring in 367 ms, no recursion); built-CLI reproduction in a disposable repo (3-bead cycle reported with public IDs, exit 1; `dep remove` with internal IDs breaks it and doctor returns to green; unmapped-bead and conflicted-`ids.yml` runs reproduce A1)

## Summary and verdict

The change is sound. `findDependencyCycles` builds the depends-on graph from the inverse `blocks` edges, runs an iterative Kosaraju pass, and reports one shortest closed path per cyclic component, sorted, so output is deterministic and safe on large graphs. Missing targets stay with the orphan diagnostic and are excluded from traversal, which is the right split. Doctor treats a cycle as an error-level finding (exit 1) and does not guess which edge to remove; I agree with both choices. Design assessment: no materially better approach found; the module mirrors the existing `issue-hierarchy.ts` split (pure graph check in `lib`, formatting in `doctor`), and reuse from `dep add` is possible later (A5).

**Verdict: approve with nits.** No Blocker or High findings. A1 and A2 are worth fixing before merge because PR #307 edits the same `Dependencies` check (see A2), and both need one small change each.

Not sensitive in security, performance, or correctness beyond what this review covered: read-only diagnostic, no persisted data or file mutation, no untrusted input beyond the repo's own issue files. No dedicated review recommended.

## Findings

**A1 · Medium · `packages/tbd/src/cli/commands/doctor.ts:1041-1051`** · The public-ID formatter is not resilient, in two ways with one root cause. (a) An issue whose ULID is missing from `ids.yml` renders as the constant `<unmapped bead>`, so every unmapped member of a path becomes the same token: reproduced as `depends-on cycle: <unmapped bead> -> cyc-np1u -> cyc-bz6z -> <unmapped bead>`, and a fully unmapped cycle prints four identical placeholders with a suggestion (`tbd dep remove <issue> <depends-on>`) the user cannot follow. This is a state doctor itself recognizes: "ID mapping coverage" reports it and `--fix` repairs it later in the same run (check order), so the `--fix` run still prints placeholders. (b) `loadIdMapping` throws on a conflicted `ids.yml` (`MergeConflictError`), and because the mapping is loaded before anything is formatted, the whole check degrades to `✗ Dependencies - check could not complete: ...` and the cycle or orphan finding is hidden until the mapping is fixed. The old orphan check never touched the mapping, so this is a small regression in resilience.
**Fix:** make formatting best-effort: wrap `loadIdMapping` in try/catch and fall back to an identity formatter; for an unmapped ULID return the internal ID (`tbd dep remove` accepts `is-<ulid>`, verified), optionally suffixed ` (no short ID)`, instead of a constant placeholder. Add a unit test through `dependencyFinding` with a formatter that maps only some IDs.

**A2 · Medium · `packages/tbd/src/cli/commands/doctor.ts:164-174`** · When a cycle and orphans coexist, the combined `error` result omits `fixable: true` and the `Run: tbd doctor --fix` guidance, so the summary flips from "Run with --fix to repair" to "may require manual intervention" and the `[fixable]` marker disappears for the orphans (probe output: `1 directed cycle(s) and 1 orphaned reference(s)`, no `fixable`). On this branch alone that is harmless because orphan repair does not exist yet (`tbd-vv9t`), but PR #307 implements it in the same method (`checkOrphanedDependencies(issues, fix)`), and the two PRs conflict textually on both hunks. Whichever lands second must reconcile them, and the merged check must still run and advertise the orphan repair when a cycle is also present.
**Fix:** set `fixable: orphans.length > 0` on the combined result and keep the orphan suggestion alongside the cycle one (for example `details` as now, `suggestion: 'Break each cycle with: tbd dep remove <issue> <depends-on>; run: tbd doctor --fix for the orphaned reference(s)'`), and cover the combined case in the unit test. Coordinator note: rebase the later of #306/#307 onto the other and re-run its tests; the orphan-repair path from #307 should run before `dependencyFinding` reports.

**A3 · Medium · `packages/tbd/tests/issue-dependency-graph.test.ts:1-73`, `packages/tbd/tests/doctor-sync.test.ts:58-91`** · Coverage stops at the two cases named in the bead. Untested: the combined cycle-plus-orphan branch (the only branch with new message logic), a self-loop, two components (sort order of the returned paths), the orphan-only path through `dependencyFinding`, and the CLI contract (error-level finding, exit 1), which `error-handling-rules` asks for on every failure path. The existing `detects orphaned dependencies` test in `doctor-sync.test.ts` re-implements the orphan loop inline and asserts its own copy, so it protects nothing in `doctor.ts`.
**Fix:** add table-driven cases to `issue-dependency-graph.test.ts` for self-loop, two components, and cycle plus orphan (assert message, `details` order, and `fixable` per A2); replace the inline loop in `doctor-sync.test.ts:79-90` with a call to `dependencyFinding`; and add one e2e case in an existing doctor e2e suite (for example `common-dir-layout-doctor.test.ts`) that writes a 2-bead cycle and asserts the `Dependencies` entry in `doctor --json` has `status: 'error'` and the process exits 1.

**A4 · Low · `packages/tbd/docs/tbd-design.md:4076-4084`, `packages/tbd/docs/tbd-docs.md:1015-1019`** · Documentation gaps. The design doc's doctor table ("5. Schema and Reference Checks") lists orphaned dependencies but not the new cycle check (error, not auto-fixable). In the manual, the paragraph "The dependency record stores the inverse `blocks` direction ... Earlier versions of doctor checked only that every stored target existed ..." is change history and storage internals rather than user guidance.
**Fix:** add a `Dependency cycle | error | no | Directed depends-on cycle` row to the design table; in the manual keep the arrow semantics sentence and drop the "earlier versions" history, and state that a cycle is an error-level finding (doctor exits nonzero).

## Suggestions

**A5 · Low** · `tbd dep add` still writes cycle-closing edges; the hierarchy code has the precedent of refusing at write time (`checkParentAssignment` in `update.ts:178-190`) and detecting in doctor. `findDependencyCycles` could be reused after the validate-all-then-apply step in `dep.ts` to reject an edge that closes a cycle. Out of the bead's scope; suggest a follow-up bead rather than widening this PR.

- `await import('../../file/id-mapping.js')` in `checkDependencies` conflicts with the static-import rule in `typescript-rules`, but it copies the file-wide pattern (ten dynamic imports in `doctor.ts`); changing one instance would not help. Fix file-wide separately if wanted.
- Non-null assertions in `issue-dependency-graph.ts` (`get(...)!`, `component[0]!`, `left[0]!`) are all bounds-established locally (map pre-seeded with every issue id; SCCs are non-empty; cycles have length ≥ 2), and the lint config allows `!`.
- `for (const node of queue)` while pushing to `queue` is a valid BFS in JS (the array iterator re-reads length); the 3-node path test exercises it.
- Mixed `sort()` and `localeCompare()` comparators: IDs are lowercase ASCII, so the orderings agree and output is deterministic.
- One representative path per component: a component with several simple cycles (or a self-loop on its smallest node, which wins because it sorts first) needs a second doctor run after one edge is removed. The manual says to re-run doctor; acceptable.
- `fixable: true` on the orphan-only result is a pre-existing false promise, already tracked and fixed by #307.
- Performance: O(V+E) with per-node sorts; 30,000-node ring in 367 ms, iterative traversals, no recursion depth risk.

## CI status

All 7 checks pass at `076e1ad` (Benchmark, Coverage & Lint, DeepSource Secrets, Test on macOS Node 24, Ubuntu Node 22.12.0 and 24, Windows Node 24). Not repeated locally beyond the targeted tests above.

jlevy:
**Dispositions for review A** at `88a4ddc`

- **A1: fixed** in `8608253`: dependency diagnostics no longer depend on a usable `ids.yml`. `dependencyIdFormatter` (`doctor.ts`) returns the internal ID for a ULID missing from the mapping instead of the constant `<unmapped bead>`, so every path member stays distinct and `tbd dep remove is-<ulid>` works; `checkDependencies` now loads the mapping with `.catch(() => null)`, so a conflicted `ids.yml` falls back to internal IDs for every bead rather than collapsing the check into "check could not complete". The ID mapping checks still report the mapping problem itself. Confirmed by `names beads without a short ID, or with no usable mapping, by internal ID` in `tests/issue-dependency-graph.test.ts` (partial mapping and `null` mapping) and by the conflicted-`ids.yml` half of the new CLI test in `tests/common-dir-layout-doctor.test.ts`.
- **A2: fixed** in `ccc38de`: the combined cycle-plus-orphan branch of `dependencyFinding` now sets `fixable: true` and carries both repairs in the suggestion (`Break each cycle with: tbd dep remove <issue> <depends-on>. Run: tbd doctor --fix to repair the orphaned reference(s).`), so the `[fixable]` marker and the "Run with --fix to repair" summary survive when a cycle is present. The cycle-only branch was split out and keeps the manual-repair suggestion. `fixable: true` rather than `fixable: orphans.length > 0` because that branch is only reached when `orphans.length > 0`. Confirmed by the `keeps orphan repair guidance when a cycle is also present` case in `tests/issue-dependency-graph.test.ts`.
- **A3: fixed** in `ccc38de` and `8608253`: `tests/issue-dependency-graph.test.ts` gains a table-driven `dependencyFinding` suite covering a self-loop, two cyclic components (asserting the sorted `details` order), the orphan-only path, and cycle-plus-orphan (asserting message, `details` order, and `fixable`), plus the formatter case above. `tests/doctor-sync.test.ts` now calls `dependencyFinding` and asserts the whole result instead of re-implementing the orphan loop inline. `tests/common-dir-layout-doctor.test.ts` gains an e2e case that creates a 2-bead cycle through the built CLI and asserts `doctor --json` reports the `Dependencies` entry with `status: 'error'` and exits 1. All three files pass (37 tests).
- **A4: fixed** in `88a4ddc`: `tbd-design.md` gains the `Dependency cycle | error | no | Directed depends-on cycle` row in the "Schema and Reference Checks" table. In `tbd-docs.md` the "earlier versions of doctor" history and the storage-direction sentence are dropped, the arrow-semantics sentence is kept, the paragraph now says a cycle is an error-level finding that makes doctor exit nonzero, and it notes that a bead without a short ID appears by its internal ID. Confirmed by `flowmark --check` on both files and by review of the rendered sections.
- **A5: deferred**: tracked as `tbd-08gn` (open), waiting on this PR merging to `main`, since the write-time check would reuse `findDependencyCycles` from `src/lib/issue-dependency-graph.ts`, which lands here. The bead records the `checkParentAssignment` precedent in `update.ts` and the insertion point after the validate-all-then-apply step in `dep.ts`.

Beads: parent `tbd-sdm9`; children `tbd-vc1f` (A1), `tbd-okhw` (A2), `tbd-m386` (A3), `tbd-cr9x` (A4) closed; `tbd-08gn` (A5) left open.

Verification at this head: `tests/issue-dependency-graph.test.ts`, `tests/doctor-sync.test.ts` and `tests/common-dir-layout-doctor.test.ts` (37 passed), the full package suite via the pre-push hook (2689 passed, 1 skipped), `pnpm --filter get-tbd typecheck`, prettier and eslint on the changed TypeScript, and flowmark on the changed Markdown.

Nothing in the "False positives / do not fix" section was changed.

PR #309, opened 2026-09-17
plan: PR review lifecycle, policy grants, and sub-agent delegation

jlevy:
## Summary

Adds project policy grants, a consistent PR review lifecycle, and tier-based sub-agent delegation. Bottom layer of stack 312, with #310 above it. The implementation is tracked in epic `tbd-kegh` and the [plan](https://github.com/jlevy/tbd/blob/claude/pr-review-lifecycle-and-delegation/docs/project/specs/active/plan-2026-09-16-pr-review-lifecycle-and-agent-delegation.md).

- `tbd policy show|grant|revoke|set|refresh` records policy in the tbd integration block of `AGENTS.md`. Effective grants come from a verified, fetched remote-default ref; branch proposals remain separate. Repositories without remotes retain local policy behavior.
- Setup preserves project instructions and policy blocks, installs Claude Code and Codex tier definitions, and reports failed installations. Policy parsing and writes reject malformed boundaries and unsafe file targets. Doctor and prime report policy state.
- Review shortcuts share pinned heads, finding IDs, dispositions, review coverage, and an explicit merge gate. Atomic stack merges gate every included layer and use the selected merge method.
- Generated integrations use format `f100`, distinct from repository format `f08`, so older clients refuse to overwrite policy-bearing surfaces. Issue commands and sync remain compatible.
- Updates skill routing, README content and generated reference tables, engineering docs, changelog, and the delegation research.

## Status

**Not ready to merge.** Rebased onto current `main` (`a92ecab9`) and conflict-free; #316's reviewable-unit and stack-routing rules were reconciled and verified to survive the rebase.

Round 10 published two reviews at this head, both of which required changes:

- [Review N](https://github.com/jlevy/tbd/pull/309#issuecomment-5751213188), senior: 5 High, 16 Medium, 8 Low.
- [Review O](https://github.com/jlevy/tbd/pull/309#issuecomment-5751213060), security: **2 Blockers**, 1 High, 3 Medium/Low.

The Blockers are policy-visibility bypasses: two inputs produce a block that renders with no grant visible on GitHub while `tbd policy show` reports `github-merge: autonomous` as effective. This was the fourth round to find an input of that class, so the read path is being replaced with an allow-list (locate the markers as top-level lexer tokens, read grants only from the expected list token, fail closed on any raw HTML between the markers) rather than patched again.

Fix work for all findings is in progress on this branch. Every finding has a bead under `tbd-uci2`; disposition replies for reviews J, K, L, M, N and O follow the fixes.

[...]

PR #310, opened 2026-09-17
Tier agent definitions: platform-specific bodies, and sub-agent caching research

jlevy:
## Summary

Improves generated sub-agent instructions and documents how delegation affects context and prompt-caching cost. Top layer of stack 312, above #309 (base: `claude/pr-review-lifecycle-and-delegation`), tracked by `tbd-49pp` under `tbd-kegh`.

- Generated Claude Code and Codex tier-agent bodies omit unreliable model/effort self-descriptions, retain task boundaries, and require evidence in reports. All eight checked-in definitions match the generator byte for byte.
- The research explains platform system prompts, exact-prefix matching, cache writes and reads, model-specific prices and minimums, TTL behavior, and report costs. It separately compares inline work, forks, and fresh agents.
- Caching examples distinguish effort-only reuse from different definition bodies and account for the extra reads incurred by more frequent polling.
- `agent-model-tiers` and `delegate-to-subagents` describe what definitions supply on each platform and how to manage waits. The brief remains the portable task contract.

## Stack status (both layers)

**The stack is not ready to merge.** Both layers are rebased onto current `main` (`a92ecab9`) and linked as stack 312. It merges bottom to top, atomically, so #310 cannot land before #309 is ready.

| Layer | Head | Round 10 verdict | Blocking work |
| --- | --- | --- | --- |
| #309 (bottom) | `5f29c4dd` | Changes required | 2 Blockers (policy-visibility bypass), 5 High, 16 Medium |
| #310 (top, this PR) | `dabcaa41` | Changes required | 3 Medium, 4 Low, no code defect |

The stack's blocking work is on #309: two confirmed inputs make a policy block that renders with no grant visible on GitHub while the CLI reports `github-merge: autonomous` as effective. Its read path is being replaced with an allow-list. See [Review N](https://github.com/jlevy/tbd/pull/309#issuecomment-5751213188) and [Review O](https://github.com/jlevy/tbd/pull/309#issuecomment-5751213060).

Every finding across both layers has a bead under `tbd-uci2` (43 findings). Fix work is in progress; disposition replies for J, K, L, M, N, O on #309 and D, E, F on #310 follow the fixes. The merge itself is gated by `tbd-jivd`, and the by-name confirmation of #309's seven proposed policy values by `tbd-px37`. Both are open, and this task does not authorize a merge.

## This layer's status

[Review F](https://github.com/jlevy/tbd/pull/310#issuecomment-5751213300) found no code defect. The generator change is correct: a fresh `setup --auto` reproduces all eight definitions byte for byte, and the `--list` golden matches the file size. Every cost example in the research was recomputed independently and reproduces, including $0.4125 for 16 polls with five-minute writes against $0.3900 for seven polls with a one-hour write.

Open on this layer:

- F1 and F8: the Windows-timeout commit also raises Linux and macOS budgets (30 s to 90 s, 45 s to 180 s) against the stated rule, and leaves an intermittent stall undiagnosed. It also belongs on #309 or its own PR, since it fixes exposure that #309's setup code creates and that neither layer otherwise touches.
- F2: "3 to 15 times" is quoted as Anthropic's figure; the same document gives 3 to 10 times against a single agent, with 15 against chat. The matching fix on #309 is N26.
- F3: the claim that a fork cannot change model or level is true on Claude Code but not on Codex.
- F4 to F7: a Windows watch deadline equal to its test timeout, three arithmetic slips in the fork example, wording fixes, and tier-body comments that overstate what an override changes.

Reviews A through E are dispositioned or verified; D1 to D3 and E1 are confirmed fixed at this head.

## Validation

Generator snapshots, byte-for-byte drift checks for tier definitions, integration and review-contract tests, and CLI documentation goldens cover this layer. CI runs on every push to both layers.


jlevy (commented):
**Review B** (senior engineering review, round 2) · head `4139d024` · base `claude/pr-review-lifecycle-and-delegation` at `751e0f69`
Reviewer: strong tier, requested `fable` at `xhigh` · Channel: formal review
Performed at `78665299` (merge base `34db3783`) and re-pinned to `4139d024` after a rebase onto #309 `751e0f69` whose PR diff is byte-identical to the reviewed diff (compared via `gh pr diff` against `git diff 34db3783 78665299`, index lines ignored); the file:line references below are unchanged by that rebase.
Coverage: 16 of 16 changed files. Read in full: `setup.ts` (`tierAgentBody`, `renderTierAgentDefinition`, callers), `setup-tier-agents.test.ts`, `integration-files.test.ts`, `lefthook.yml` and the `format:md` scripts, all eight generated `.claude/agents/tbd-*.md` and `.codex/agents/tbd-*.toml`, `agent-model-tiers.md`, `delegate-to-subagents.md`, and the research brief's new sections (every cost figure recomputed and every new reference V39 to V46 fetched); the brief's unchanged sections were not re-read. Round 2 is a full independent re-review of the whole diff after review A was addressed, not a follow-up limited to A's fix commits.
Tests run: `pnpm install --frozen-lockfile --prefer-offline`; `pnpm --filter get-tbd build`; `vitest run tests/setup-tier-agents.test.ts tests/review-lifecycle-contract.test.ts` (38 passed); `vitest run tests/integration-files.test.ts -t "formatter boundary"` (1 passed); `tryscript run tests/cli-doc-output.tryscript.md` (9 passed); `node packages/tbd/dist/bin.mjs doctor --json` from this head's build (Claude Code tier agents and Codex tier agents both `ok`, `current`) and from the base branch's dist (both `warn`, `stale managed file`, expected: that build carries the old generator); appended one byte to `.claude/agents/tbd-fast.md` and re-ran `vitest -t "committed tier agent files"` (fails as intended, then restored). Sources re-read: Claude Code sub-agents and prompt-caching pages (V1, V39), Claude API prompt caching and pricing (V40, V43), OpenAI prompt caching and pricing (V42, V46), `openai/codex` `child_config.rs`, `role.rs`, `awaiter.toml` at `main` (V41), and the bundled `claude-api` skill file (V45).

## Summary and verdict

The generator change is right and is now proven: the body names no model or level, the checked-in files are byte-identical to `renderTierAgentDefinition` (drift test passes at this head and fails on a one-byte hand edit), `tbd doctor` from this head's build reports both surfaces current, the lefthook `format-md` exclusion is probed by the integration test, and the CI `find` already skips `*/.claude/*`. The contract test confirms no shipped doc names a model outside the dated suggestions.

The research additions are well sourced: every figure I checked in the new sections matches the pages cited (Claude multipliers, per-model prices and minimum prefixes, the two Claude Code TTL buckets and their override order, the sub-agent system-prompt sentence, the Codex source assignments, and the OpenAI GPT-5.6+ regime and list prices). The A5 re-verification is reflected in the text, not only in the dispositions.

Three Medium findings are documentation errors in the parts agents will act on: a wrong Fable 5.1 cost comparison that contradicts the brief's own price table (B1), a break-even sentence that recommends the 1-hour TTL where the brief itself says the 5-minute default suffices (B2), and a guideline sentence that says a definition "matters only" for the level, contradicting the next paragraph and the reason this PR added the evidence rule to the body (B3). The rest are Low.

**Verdict: mergeable after #309 once B1 to B3 land (three doc corrections, no code change). No Blocker or High findings. Nothing in this PR needs to move into #309; the one #309-owned line this PR leaves stale is the plan's "Still open" item (B7), which either layer can fix.**

## Review A dispositions checked

- A1 (Blocker): holds. Golden is `7.96 kB, ~2.3k tok`; tryscript passes locally; Coverage & Lint green.
- A2 (Medium): holds. Drift test passes; `doctor` from this head's build reports `current`; the Bugbot inline is the same finding and is resolved.
- A3 (Medium): holds. `committed tier agent files match the generator` exists and fails on a hand edit; `.claude/agents/**` is in the lefthook exclude list and the integration test probes it with a real lefthook run.
- A4 (Low): holds. `effort: medium` in `tbd-fast.md` and the guideline table; the decision bead is closed.
- A5 (Low): holds. The brief's OpenAI paragraph, header note, Next Steps item, and references V42 and V46 were updated; the figures match the OpenAI pages as of this review; bead `tbd-2f9j` is closed.

## Findings

**B1 · Medium · `docs/project/research/current/research-2026-09-16-subagent-guidance-anthropic-openai.md:1149-1150`** · "On Fable 5.1 the fork's reads fall to a quarter (about $0.60 for the same 30 turns, because reads are $0.25 per million) and the fresh sub-agent's to about $0.15". A Fable 5.1 cache read is $0.25/MTok against Opus 5's $0.50/MTok: half, not a quarter (the multiplier is 0.025x of a base price that is twice as high). The brief's own table paragraph says "a cached read is half the price of an Opus 5 read". At 30 turns of about 160k tokens the fork figure is about $1.20, and the fresh sub-agent's about $0.30.
**Fix:** "fall by half (about $1.20 ... ) and the fresh sub-agent's to about $0.30"; the ranking sentence that follows stays true.

**B2 · Medium · same file, `:1021-1026`** · "so an agent that takes more than three turns inside the hour always comes out ahead with the longer entry". The break-even figures in that sentence are against uncached input (as the pricing page states them), not against the 5-minute entry. Requests that share a prefix and start less than five minutes apart refresh the 5-minute entry for free, so the 1-hour entry only wins when a start-to-start gap exceeds five minutes; the brief says exactly that later ("A sub-agent that works continuously never expires ... The 5-minute default is fine"), so this sentence contradicts it and would send a reader to the 2x write for nothing.
**Fix:** "so either entry beats no caching after a few turns; the 1-hour entry beats the 5-minute one only when a gap between requests exceeds five minutes, and an agent that takes one turn and stops pays double for nothing."

**B3 · Medium · `packages/tbd/docs/guidelines/agent-model-tiers.md:77-79`; same claim in the research brief `:834-837`** · "A definition matters only when its level differs from the session's: a session already at `xhigh` gets the same result from naming the tier's model on the Agent tool as from `tbd-strong` or `tbd-moderate`." Without a definition the spawn runs under the built-in general-purpose agent's prompt, not the tbd body, so it does not get the `tbd shortcut <name>` bootstrap, the no-commit/push rule, or the evidence rule that lines 81-84 of the same guideline (and this PR's `tierAgentBody` change) say are the body's purpose. Same model and level, not the same result. The brief's own point 2 in "What System Prompt a Sub-Agent Runs Under" makes the same argument against this sentence.
**Fix:** "A definition changes the model and level only when its level differs from the session's; it still supplies the body, so prefer a `tbd-*` definition whenever the brief does not restate those rules." Apply the same correction to the brief's paragraph.

**B4 · Low · research brief `:775-778`; `packages/tbd/src/cli/commands/setup.ts:916-918`** · "Everything the main session's system prompt supplies is absent: the harness rules, the git conventions ('commit only when asked'), the reporting rules, the tone and format guidance" and the code comment's "this is the sub-agent's entire system prompt". The docs sentence the brief quotes (V1) is accurate, but the elaboration overstates it. Evidence from this review itself, a `tbd-strong` spawn on Claude Code v2.1.274: the sub-agent's system prompt is a one-line framing ("You are Claude Code ... running within the Claude Agent SDK"), the definition body, harness notes that are reporting and tone rules (absolute paths, no emojis, no report files, cwd reset), the environment block, and `CLAUDE.md` via a system reminder; and the Bash tool description carries "Commit or push only when the user asks". The conclusion (state the rules in the body or brief) still holds, because none of that is guaranteed by the docs.
**Fix:** qualify to "the Claude Code system prompt's instructions are absent; the harness still adds a short framing, environment details, a few reporting notes, and the tool descriptions, which include a commit-only-when-asked line, but none of that is documented, so the body and brief must carry the rules"; in `setup.ts` say "the sub-agent's behavioral system prompt (the docs promise only the body plus environment details)".

**B5 · Low · research brief `:936-938` versus `:1143-1151`** · The same 30-turn, 150k-token scenario is worked twice with different totals: "$2.25 ... against roughly $0.30 as a fresh sub-agent with a 20k to 40k prefix" in the first pass and "$2.40 ... $0.60" in the second (30 turns at 40k is $0.60; $0.30 is the 20k end).
**Fix:** use one set of figures, or have the first pass point at the worked example.

**B6 · Low · research brief `:1273-1277` and `:898-900`** · Two inferences are written as facts. (a) "on GPT-5.6+ implicit mode a fresh child with a different task message will miss the parent's prefix unless the client places an explicit breakpoint" follows from the OpenAI docs, but the same paragraph says `client.rs` was not re-read for `prompt_cache_options`, so whether Codex places one is unknown. (b) "the Agent tool does not" hold same-prefix spawns: the Claude Code page documents the 5-second hold for workflow fan-outs only and says nothing about the Agent tool.
**Fix:** mark both as inferences ("if Codex sends no explicit breakpoint, ..."; "the page documents the hold for workflow fan-outs only"), or re-read `client.rs` and cite it.

**B7 · Low · `docs/project/specs/active/plan-2026-09-16-pr-review-lifecycle-and-agent-delegation.md:1361-1363` (base branch, #309) and the PR body** · The plan's "Still open" list carries "Re-verify OpenAI prompt-caching figures (`tbd-2f9j`, #310 A5). Lands on #310"; this PR did that work and closed the bead but leaves the item open. The PR body still says the OpenAI figures are "marked not re-verified" and that `gh stack link` was not run; both are stale (stack 312 exists, and `78665299` re-verified the figures).
**Fix:** tick or remove the plan item in whichever layer the coordinator prefers (a one-line change; fine on #310 since it is stacked), and refresh the PR body's Research and Stack note paragraphs.

**B8 · Low · research brief `:852`, `:855`, `:2600-2606`** · Wording and citation nits. (a) "512 tokens on the Claude 5 family (1,024 on Sonnet 5 ...)": Sonnet 5 is in that family, so say "Fable 5, Fable 5.1, and Opus 5". (b) "Caches are per model ... [V40]": the API page states workspace isolation; the per-model statement is on the Claude Code page (V39, "each model has its own cache"). (c) V45 cites a file bundled inside Claude Code, which no reader can open; the `max_tokens: 0` keep-alive and the every-platform minimum are on the public prompt-caching page according to that file's own text, so cite V40 for them and keep V45 as the secondary source for the preserved-thinking rule.
**Fix:** as listed.

## Suggestions

**B9 · Low · `agent-model-tiers.md:117-118`, `delegate-to-subagents.md:149-151`** · `subagentPromptCacheTtl` and `CLAUDE_CODE_SUBAGENT_PROMPT_CACHE_TTL` require Claude Code v2.1.242 or later (V39). A reader on an older build gets silent no-op. Optional: add the version floor, or a pointer to the brief's TTL section which could carry it.

- The base branch's dist (old generator) reports all eight files stale. Expected: `doctor` compares against its own generator, so any tbd older than this change says stale until it is upgraded, as for every generated surface; this head's build says current.
- The `description` frontmatter and TOML `description` still name the model ("on fable"): that is the coordinator-facing default for picking an agent, not the sub-agent's prompt, and it changes only with `TIER_MODELS`.
- `tierAgentBody` dropping its `platform` argument is correct; the body is identical on both platforms by design.
- The drift test's CRLF normalization is appropriate (Windows CI checks out with autocrlf; the generator emits LF).
- No shipped doc names a model outside the dated suggestions: the contract test passes at this head.
- Codex claims: a fresh child gets `base_instructions` from the session and a role's `developer_instructions` replace the config value (`build_next_config`); the `awaiter` role is commented out and runs at `low`. All verified in `openai/codex` at `main`.
- OpenAI figures (1.25x write, 0.1x read, 1,024-token minimum, `30m` as the only TTL, `gpt-5.6-sol` $4/$0.40/$5/$20, `gpt-6-astra` $10/$1/$12.50/$50): verified on 2026-09-17 (Pacific) against V42 and V46.
- Claude Code claims (sub-agents get 5 minutes even on a subscription; override order `FORCE_PROMPT_CACHING_5M`, env var, setting, `experimental.cacheTtl` v2.1.248+ with `1h` ignored on usage credits, `ENABLE_PROMPT_CACHING_1H`, bucket default; Fable 5.1 effort exception v2.1.260+; `/usage` line v2.1.251+; workflow fan-out 5-second hold; per-machine-and-directory scope): verified against V39.

## Design assessment

Removing the model and level from the body is the right call for both stated reasons (per-spawn `model` and `CLAUDE_CODE_SUBAGENT_MODEL_FORCE` make a body claim unreliable; a varying body would split the cache prefix). Keeping the body identical across platforms and letting the brief name the shortcut keeps the definitions portable. The drift test plus the hook exclusion is the standard pattern this repo uses for generated surfaces (`ci-and-gates-rules`), now applied consistently to `.claude/agents/**`. No better alternative found. One thing this locks in: the guideline now documents Claude Code internals (TTL buckets, setting names) that will age; the brief is the right home for the detail, and the guideline's pointer to it is good, so keep the guideline's version-specific text minimal (B9 is the only addition I would make).

## Documentation

- Plan "Still open" item and PR body (B7).
- `docs/development.md` and `docs-overview.md` need no change; the contract section already describes `agent-model-tiers` and its tests.

## Dedicated reviews

None apply. The lefthook change narrows a formatter's scope (no new execution, no untrusted input); there is no hot path or intricate logic. Security, performance, and correctness reviews are not warranted for this layer.

## CI status

At `78665299` (the reviewed head): all eight checks pass (Coverage & Lint, four Test jobs, Benchmark, Cursor Bugbot, DeepSource Secrets). Checks at `4139d024` were not re-read before publishing; the rebase added only #309's plan commit `751e0f69` beneath the same seven commits.

## Stack

Stack 312: #309 then #310. This review is of the layer's diff against its #309 merge base (`34db3783` when performed, `751e0f69` after the rebase; identical diff). Merge after #309; B1 to B3 first.

jlevy:
**Review D** (senior engineering review, round 4) · head `c99a141a` · base `claude/pr-review-lifecycle-and-delegation` at merge base `9ad8d197`
Reviewer: strong tier, requested `gpt-6-astra` at `max` · Channel: PR comment, published by coordinator
Coverage: 14 of 14 changed files. Full independent review of the layer, including the research additions and audit of reviews A, B, C and the unmarked Bugbot finding.
Tests run: `pnpm --filter get-tbd exec vitest run tests/setup-tier-agents.test.ts tests/review-lifecycle-contract.test.ts tests/integration-files.test.ts` — 63 passed in 3 files, 51.36 seconds. `pnpm --filter get-tbd exec tryscript run tests/cli-doc-output.tryscript.md` — 9 passed, 8.08 seconds. Coordinator supplied the build and dependency links; I did not rebuild. Read-only body comparison and Decimal arithmetic reproduced D2/D3 below.

## Summary and verdict

The generator change is sound. It removes an unreliable self-description of model/effort, retains the task boundaries, and asks for evidence. Both platform wrappers and all eight committed files agree with the generator; the snapshots, drift guard, formatter exclusion with positive control, review contract, and current `8.39 kB, ~2.4k tok` golden pass.

Three Medium findings remain in the research, all introduced by this layer: the comparison conflates forks with inline work, the same-cache claim ignores the distinct generated bodies, and the polling recommendation does not account for its additional reads. No Blocker or High finding and no remaining executable-code defect were found. Correct these bounded documentation errors before treating the cost comparison as reliable. The layer must still be refreshed after #309's final changes; review scope here deliberately uses the merge base rather than current #309 head `cba312b2`.

## Findings

**D1 · Medium · `docs/project/research/current/research-2026-09-16-subagent-guidance-anthropic-openai.md:1143` and `:1151-1152` (related `:1145`) — A fork does not return its tool transcript to the coordinator.**

The table groups “Single session (or fork)” and says all work remains in the coordinator's context. The worked fork example then explicitly says its 30 turns of tool output remain in the coordinator. A fork inherits history on input, but its new tool calls remain isolated; only its final result comes back. This gives fresh agents a nonexistent post-completion cost advantage over forks. The same conflation assigns a fork the main session's one-hour default in the “Prefix lifetime” row, although fork requests are in the five-minute sub-agent bucket, as this brief correctly states at lines 1181–1182.

Evidence: the official [Claude Code fork documentation](https://code.claude.com/docs/en/sub-agents#fork-the-current-conversation) explicitly states that fork tool calls stay outside the main conversation and only the final result returns. The [cache TTL buckets](https://code.claude.com/docs/en/prompt-caching#which-ttl-each-request-gets) put forks outside the main conversation. Existing parent-prefix cache entries may still be reused, but a fork's new writes do not inherit the main-conversation one-hour default.

**Fix:** separate inline work from forked work in this comparison. Both a fork and a fresh sub-agent return a report; inline work retains the tool transcript. Keep the valid fork cost distinction—each fork turn carries the inherited parent history—and give fork writes the sub-agent TTL. Update the worked example's final clause accordingly.

**D2 · Medium · same research file `:1116-1118` (related `:1086-1092`) — The two strong definitions do not have the same full cache prefix.**

The paragraph says `tbd-strong` at `xhigh` and `tbd-strong-max` at `max` are “one” cache on Fable 5.1. Their generated system bodies already differ in the first sentence: `.claude/agents/tbd-strong.md:11` says `You are tbd-strong, a tbd strong-tier sub-agent.`, whereas `.claude/agents/tbd-strong-max.md:11` says `You are tbd-strong-max, a tbd strong-tier sub-agent.` The Fable effort-change exception removes one invalidation cause; it cannot make these different system prompts equal. At most an earlier separately cached tool prefix could be shared, not the body/project/conversation prefix described here.

Evidence: direct read-only comparison of both committed bodies, which the passing drift test proves equal the generator; [Claude's exact-prefix rule](https://code.claude.com/docs/en/prompt-caching#how-the-cache-is-organized). The [effort exception](https://code.claude.com/docs/en/prompt-caching#changing-effort-level) concerns changing effort without changing the prefix.

The related paragraph also gives static model text a cache argument it does not have: the old generator embedded a fixed model string per definition, and a spawn override did not substitute that text. Removing it fixes a false self-description, but by itself does not make repeated spawns of one definition more cacheable.

**Fix:** say that effort alone need not invalidate the cache on supported Fable connections, while these named definitions still have different bodies and therefore cannot share the full prefix. Remove the claim that removing the old static model name itself improves within-definition cache reuse. Keep the generator change for its valid accuracy reason; no code change is needed.

**D3 · Medium · same research file `:1198-1200` — Polling every four minutes is not always the cheapest strategy.**

The paragraph compares a ten-minute poll interval using a one-hour TTL with four-minute polls using the five-minute TTL, then declares the latter cheapest because it uses reads. Extra polls are extra paid reads, even before counting their new messages or model output. Using the paragraph's own 30k-token Opus prefix and current $5/MTok input, $0.50/MTok cached rate, an hour from minute 0 through minute 60 costs:

- Five-minute TTL, polls at 0, 4, …, 60: one $0.1875 write plus 15 × $0.015 reads = **$0.4125**.
- One-hour TTL, polls at 0, 10, …, 60: one $0.30 write plus 6 × $0.015 reads = **$0.3900**.

This counterexample holds with a constant prefix and zero output cost; real added polling work only adds cost. The two intervals have different completion-detection latency, so the choice also involves that tradeoff. This is separate from B2: B2 correctly compared both TTLs at a fixed workload; this sentence changes the workload.

Evidence: Decimal calculation reproduced locally; the [Claude prompt-caching price table](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#pricing) confirms the rates used by the brief.

**Fix:** replace “cheapest of all (reads only)” with a conditional statement: polling within the TTL avoids expiry, but total cost depends on the extra polls, prefix size, output, and required detection latency; a longer TTL can be cheaper with less frequent polls. Keep the fixed-workload TTL break-even explanation.

## Previous dispositions audited

- **A1: fixed.** The guideline is 8,387 bytes; the current size/token golden and all 9 tryscript tests pass.
- **A2 / unmarked Bugbot: fixed.** All eight committed definitions match the generator in the passing drift test. Generated output and snapshots agree. The historical Bugbot issue is the same finding, not a new one.
- **A3: fixed in the base.** The drift test and formatter exclusion live below #310; the formatter test also has its README positive control and passes. Their relocation/provenance correction is present.
- **A4: resolved.** Fast remains `medium` in the source table, generated frontmatter, and guideline. The disposition explicitly records that decision; there is no need to invent another model decision.
- **A5: fixed.** The dated OpenAI re-read, reference records, and completed Next Steps entry exist. I independently fetched official OpenAI prompt-caching and pricing docs: the two caching regimes, GPT-5.6+ 1.25× write / 0.1× read / 1,024-token minimum / 30-minute TTL, and both named models' prices match.
- **B1: fixed.** Half-price Fable cached reads and the $1.20 / $0.30 examples match the price table; no quarter-price claim remains there.
- **B2: fixed.** Break-even against uncached input is distinguished from the extra TTL premium. D3 is the separate changed-poll-count mistake.
- **B3: fixed as requested.** Both guideline and research now acknowledge that a definition still supplies the body even when the requested reasoning level matches the session. The exact replacement requested in B was applied.
- **B4: fixed as requested.** The research qualifies the harness framing/tool descriptions; the code comment now describes the behavioral prompt and the documented guarantee rather than the entire raw request.
- **B5: fixed.** The earlier scenario points to the one $2.40 / $0.60 worked example rather than presenting conflicting totals. D1 concerns its separate return-context assumption.
- **B6: resolved.** The five-second hold is scoped to workflow fan-outs. The Codex client read is recorded rather than left hypothetical; current `client.rs` contains `prompt_cache_key` and no cache-options/breakpoint/retention field.
- **B7: addressed by comments / base-owned plan updates.** The PR description remains stale, but the existing correction comments explicitly cover the facts. The plan is in #309; current #309 metadata is the coordinator's scope.
- **B8: fixed.** Named per-model prefix minimums replace the overbroad family label; the per-model isolation and keep-alive citations point to their appropriate public sources.
- **B9: fixed.** The v2.1.242 floor is stated in both shipped documents and supported by the official Claude Code page.
- **C1: the rebuttal holds.** I independently fetched the entire [OpenAI prompt-caching guide](https://developers.openai.com/api/docs/guides/prompt-caching), including the shared-prefix gotcha. An eligible lookup boundary does not imply that a prior request wrote an entry there. The guide's static-developer/dynamic-user example requires an explicit shared-prefix breakpoint; the current clarification reflects it. Do not reopen C1 on the old argument.
- **C2: addressed by comment, description still stale.** The current description still has “not re-verified”, “pending a decision”, the old definition-benefit summary, and the branch-only stack note. The supplied history's C disposition explicitly corrects them and documents the original edit-channel restriction. A coordinator with current edit permission should refresh the description, but this is already tracked/resolved through the accepted comment fallback, not a new code finding.
- **C3: fixed.** Earlier-model minimums no longer claim 1,024–2,048; the 2,048 interval is correctly separated. Matches official OpenAI docs.
- **C4: fixed.** The medium-session exception is explicit.
- **C5: fixed.** The research filename, owning tbd repository, and non-packaged status are explicit in the shipped guideline.
- **C6: fixed.** `integration-files.test.ts` is absent from #310's merge-base diff.
- **C7: fixed in #309/base.** The non-excluded README positive control exists and passes here.
- **C8: fixed.** The TTL setting names and floor are explicitly dated 2026-09-18.

## Design assessment and false positives

One small platform-independent body rendered into two existing wrappers is appropriate; no new abstraction or dependency is needed. The brief continues to own task-specific authorization and boundaries. Dropping the now-unused `platform` parameter is correct. Coordinator-facing descriptions may name the default model/effort because that metadata is not the sub-agent's behavioral prompt. Keeping CRLF normalization in the drift guard is appropriate for Windows.

The key validity reason for removing model text is runtime overrides, not cache savings. D2 corrects only the research explanation. There is no reason to change the generated files again.

The inspected external claims are not being treated as guessed future features: current primary pages do document the named caching regimes, settings, versions, and models. I also read `openai/codex` `child_config.rs`, `role.rs`, and `client.rs` from `main`: base-instruction inheritance and role replacement match the report. Source blobs returned were respectively `dbde75b867d717b95ed396f1818ccdb3dc7c12a3`, `f6633e7590a903e7a0265f43b5fb23ccaeee5c5d`, and `1c5b49353a0fa1a5f41c1a3e1a613ca2cb813ebd`. These are source-level checks, not live billable API or Claude harness experiments. The bundled Claude API skill's V45-only behavior was not re-executed or independently inspected in this pass; no new finding is based on it.

No dedicated security or performance review is warranted by the changed string-building function; no new input boundary, permission control, I/O path, concurrency, or persisted format is introduced. The research's arithmetic was checked directly rather than spawning another reviewer.

## CI and stack status

All current CI checks pass at `c99a141a`; Bugbot is neutral after usage-limit skips and is not fresh automated-review evidence. The layer is behind #309 current head `cba312b2` and must be refreshed after the #309 fixes, with CI run again. The stale PR description will also be updated, completing the existing B7/C2 correction directly now that edit access is available.

jlevy:
**Review E** (senior engineering review, round 5) · head `c4aff495` · base `claude/pr-review-lifecycle-and-delegation` at `b4923d76`
Reviewer: delegated senior reviewer; coordinator's inherited configuration, no model/effort override requested · Channel: PR comment, published by coordinator
Coverage: 16 of 16 changed files. Read every changed hunk, generator and relevant setup callers, generated definitions, changed tests and surrounding fixtures, both shipped guidance documents, and all new research sections. The research's unchanged historical sections were not re-audited. This is a full review of this layer, not only the prior fixes.
Tests run: isolated `pnpm --filter get-tbd build` passed; CLI documentation goldens passed (9 tests). Focused Vitest run passed (88 tests, 5 files): `setup-tier-agents`, `bead-watch`, `doctor-managed-surfaces`, `review-lifecycle-contract`, and `integration-files`. CI: all seven reported checks pass at this head.

## Summary and verdict

**Verdict: approve with one Low documentation correction.** No new Blocker, High, Medium, or executable-code defect found. The single shared body remains a suitable design: platform wrappers retain model/effort selection, while the body describes task boundaries and evidence without claiming an overridden model identity. The added timeout budgets retain the watch-isolation and doctor assertions; neither test is skipped or weakened.

## Findings

**E1 · Low · `packages/tbd/docs/guidelines/agent-model-tiers.md:77-79` and `docs/project/research/current/research-2026-09-16-subagent-guidance-anthropic-openai.md:835-837` — Separate model binding from reasoning-level inheritance.**

The new sentence says a definition changes “the model and level only when its level differs from the session's.” Model selection is independent of the effort comparison. With an Opus/xhigh parent and no higher-priority model override, the generated `tbd-strong` definition selects Fable/xhigh: the model changes although the reasoning level is unchanged. The generated `model` and `effort` fields and the documented [Claude Code model precedence](https://code.claude.com/docs/en/sub-agents#choose-a-model) establish the counterexample. The next sentence correctly compares a definition with an explicitly supplied matching model, but the preceding blanket sentence does not state that premise.

**Fix:** say that a definition is needed to set a reasoning level different from the session's; independently it supplies a model default and the body. Preserve the next sentence's explicit-model comparison. Apply the same clarification to both documents, and refresh the guideline-size golden if it changes. This preserves B3's valid body-instruction correction; it does not reopen B3's disposition.

## Earlier reviews and remaining addressing work

The discovery sweep covered formal reviews, PR comments, inline comments, and GitHub issues referencing 310 (none). Reviews A–C have disposition replies. The old Bugbot generator mismatch is A2 and remains corrected.

Review D has no `tbd:dispositions v=1 review=D` reply in the retrieved history. Its three corrections are present in `18a77e9e` and the current head:

- **D1:** the table distinguishes inline work, forks, and fresh agents; forks return only a final report and new writes use the sub-agent TTL.
- **D2:** the research distinguishes effort-only cache reuse from different generated bodies, and no longer attributes within-definition cache savings to deleting static model text.
- **D3:** the polling example counts the additional reads: `$0.1875 + 15 × $0.015 = $0.4125` versus `$0.30 + 6 × $0.015 = $0.3900`.

Publish the missing D disposition with these existing fixes and validation; do not duplicate D1–D3 as new findings. The PR description currently claims that reply already records the evidence, so closing that loop also makes the description accurate.

## Design assessment, documentation, and false positives

No materially better implementation is needed. Removing the unused platform argument is appropriate. Keeping model names in coordinator-facing metadata is correct; removing them only from the behavioral body addresses runtime overrides. Existing byte-for-byte generation and CRLF-normalized drift coverage are the right regression boundary, and the base layer's formatter exclusion keeps those files generated.

The bounded source audit confirms the earlier fixes: official Claude documentation distinguishes fork output isolation and sub-agent TTLs, while the current [OpenAI prompt-caching guide](https://developers.openai.com/api/docs/guides/prompt-caching#gotchas) still distinguishes lookup boundaries from prefixes actually written. C1's rebuttal therefore holds; do not reinstate it. The research's external claims remain dated observations, not a promise that every future client version behaves identically. No live billable API experiment was performed.

## CI and stack

Formal stack 312 contains #309 then #310. At this review's head, Coverage & Lint, Benchmark, DeepSource Secrets, and all four OS/Node test jobs pass. Local results above are independent focused evidence; the full OS matrix was inspected on CI. No dedicated security, performance, or intricate-correctness pass is warranted by this layer's string-body change and test-budget adjustments. Any later #309 changes must be carried upward before final CI is checked again.

jlevy:
**Review F** (senior engineering review, round 6) · head `dabcaa41` · base `claude/pr-review-lifecycle-and-delegation` at `5f29c4dd`

Reviewer: Claude Code coordinator (Claude Fable 5.1) with an independent strong-tier sub-agent · Channel: PR comment
Coverage: all 16 changed files, then the stack as a whole. External pages (Anthropic, OpenAI, Claude Code) were not re-fetched; external figures were checked for internal consistency, attribution and dates only.
Tests run: a fresh `setup --auto` compared byte for byte with all eight tier definitions (identical); the `--list` golden checked against the file size (8,405 bytes, shown as 8.4 kB); every cost example in the research recomputed. Both layers were rebased onto current `main`; `git range-diff` applies to older heads.

## Summary and verdict

**Changes required: three Medium findings, no code defect in the generator.** D1, D2, D3 and E1 are fixed at this head. The generator change is correct and the definitions match it. The remaining problems are a timeout commit that does more than its message says and belongs to another layer, one misattributed figure, and one claim in the shipped guideline that is true on only one platform.

## Findings

**F1 · Medium · `packages/tbd/tests/doctor-managed-surfaces.test.ts:25, 44`; `tests/test-helpers.ts:19-20`.** The timeout commit (`3e7e19c3` at this head) raises non-Windows budgets and leaves a stall undiagnosed. The helper's argument is the non-Windows value, so Linux and macOS go from 30 s to 90 s and from 45 s to 180 s, against `vitest.config.ts`: "Raising it only on Windows keeps the tight budget everywhere it is meaningful". The commit message records the same hook taking under 11 s and over 60 s in one run, which indicates an intermittent stall rather than uniform load; "Windows CI under parallel load" is an explanation the repository's own principles ask to be confirmed, not assumed.
**Fix:** keep 30 s and 45 s off Windows and raise Windows only. On a setup failure print `setup.stderr` and the elapsed time so the next timeout names the phase that stalled.

**F2 · Medium · research `:948, 1172`.** "Anthropic's 3 to 15 times figure [V10], [V11]" is a synthesized number quoted as a source's. The same document at `:730-732` gives 3 to 10 times against a single agent, with 15 times against chat.
**Fix:** "the 3 to 10 times single-agent multiplier [V11] (about 15 times chat [V10])". #309's `delegate-to-subagents.md:59` carries the same error (Review N, N26) and should be fixed on that layer.

**F3 · Medium · `packages/tbd/docs/guidelines/agent-model-tiers.md:109-110`.** "and cannot change model or level" is not scoped to a platform, though the paragraph covers both. The research at `:535-538` says Codex applies `model` and `reasoning_effort` on every spawn, full-history forks included.
**Fix:** "on Claude Code a fork cannot change model or level; on Codex the platform tells its model not to".

**F4 · Low · `packages/tbd/tests/bead-watch.test.ts:603`.** On Windows the watch deadline becomes 60 s, equal to the describe timeout at `:507`, so a real timeout surfaces as a vitest timeout rather than the `changed` assertion, and six polls fit in the window, so a first-poll regression passes.
**Fix:** `process.platform === 'win32' ? 9_000 : 5_000`, below both the poll interval and the test timeout.

**F5 · Low · research `:1157-1158, 1169-1170, 1204`.** The fork example uses a flat $0.08 per turn while the fresh example assumes 40k of growth; with matching growth the fork costs $2.55, not $2.40. "Either way and dominates" does not hold for the fork on the document's own figures ($2.40 against $0.50 to $1.50). "About $1.10 per hour" is $1.125. All other figures reproduce, including $0.4125 for 16 polls with five-minute writes against $0.3900 for seven polls with a one-hour write.
**Fix:** correct the three numbers and the sentence.

**F6 · Low · `delegate-to-subagents.md:155-159, 258`; `agent-model-tiers.md:78-79`; research `:836-837, 906-908`.** The v2.1.242 floor is attached to the five-minute default but belongs to the TTL settings. "Only if the brief or the shortcut it names states them" omits the definition body, which the previous sentence names, and the tool descriptions also carry a commit-only-when-asked line. "Its warmed cache" holds only within the TTL. "Whenever the brief does not restate those rules" refers to rules that first appear four lines later. The non-sharing explanation between tiers names model and effort only, though the bodies differ too.
**Fix:** reword each.

**F7 · Low · `packages/tbd/src/cli/commands/setup.ts:929-931`; `tests/setup-tier-agents.test.ts:317, 321, 336`; `packages/tbd/docs/tbd-design.md:5785-5787`.** The code and test comments say a per-spawn override can change model or level; the guideline and research say only the model. The tests assert only that the word "reasoning" is absent. Design section 6.4.7 omits the evidence rule and the no-model rationale.
**Fix:** align the comments, add `not.toContain` for each tier's model name in the loop for both platforms, and update the design section.

## Stack as a whole

Each layer stands alone as a review unit: #309 has no forward references into #310, its golden matches its own file, and its committed definitions match its own generator. **F8 · Medium:** the timeout commit is misplaced. It touches `bead-watch.test.ts`, which neither layer otherwise edits, and `doctor-managed-surfaces.test.ts`, whose cost comes from #309's setup code, while #310's only code change is a string. If #309 lands alone, `main` has the same Windows exposure.
**Fix:** move it to #309 or to its own PR, with the F1 correction.

The decision to keep the fast tier at `medium` is recorded and closed. C1's lookup-versus-write rebuttal holds. Anthropic and OpenAI caching rules are kept separate; the only omission is the read discount for earlier OpenAI models.

## CI status

All seven checks are pending at this head as this is published; the previous head conflicted with `main`, so Actions had not run since `b4923d76`. Local at this head: `ci:quality` and the build check pass; the full suite ran once alone with 2,981 passed, 1 skipped and 1 failed. The failure is `tests/git-remote.test.ts` "end-to-end: outbox save after merge", a real-Git test on vitest's default 5 s timeout that takes 1.4 s alone and timed out under full-suite load; it is unchanged from `main` and tracked as `tbd-t7s7`. The push skipped only the local `test` hook for that reason; CI is the full-suite gate.

This review does not authorize a merge.

jlevy:
**Dispositions for review F** at `85e2d023`

This layer was rebased onto #309's new head, so the SHAs below are post-rebase. The fixes are in `85e2d023`.

- F1: fixed, on the other layer. The timeout work now lives on #309 as `c122b241`, which raises the Windows subprocess floor to 90 s and leaves the Linux and macOS budgets at 30 s and 45 s, which is what this finding asked for. That commit also records what it ruled out (not ENOTDIR, not the watch and doctor pair) rather than attributing the failures to load without evidence. This layer no longer changes any test budget.
- F2: fixed in `85e2d023`: the research no longer attributes "3 to 15 times" to Anthropic. It is the 3 to 10 times single-agent multiplier, about 15 times chat, in all four places, and the shortcut and spec were corrected on #309 as N26.
- F3: fixed in `85e2d023`: the claim is scoped per platform. On Claude Code a fork cannot change model or level; on Codex the platform instructs its model not to while the runtime applies both on every spawn.
- F4: fixed, on the other layer, in `c122b241`: the watch deadline is the shared subprocess floor rather than a fixed 1 s, so a real timeout no longer surfaces as a vitest timeout on Windows.
- F5: fixed in `85e2d023`: the fork example now grows its context on the same assumption the fresh example uses, giving $2.55 rather than $2.40 and about $1.27 on Fable 5.1; the claim that the review work dominates "either way" is corrected, since it dominates the fresh sub-agent's context reads but not the fork's; and six polls per hour at $0.1875 is $1.13, not $1.10. The polling comparison was recomputed and stands: $0.4125 against $0.3900.
- F6: partly fixed in `85e2d023`, partly deferred. The delegation shortcut's boundary bullet was rewritten, which covers the sync point below. The remaining wording items (the v2.1.242 floor attached to the wrong sentence, the definition body omitted from the list of rule carriers, "warmed cache" without its TTL qualifier, the forward reference in `agent-model-tiers`) are deferred as `tbd-w1km`: they are wording, not claims that mislead about behavior.
- F7: fixed in `85e2d023`: the test now asserts that a body states neither its model nor its reasoning level, excluding the agent's own name, which is what the comment claimed and the old `not.toContain('reasoning')` did not check. The comments are corrected to say an override changes the model, and design section 6.4.7 records the evidence rule and why a body names neither fact.
- F8: fixed, as described under F1.

**One change beyond the review.** The generated tier bodies told every sub-agent "Do not commit, push, run tbd sync, or start sub-agents unless the brief says so". Syncing publishes bead updates on tbd's own data branch, carries no code and opens no PR, and needs no grant, so that blanket prohibition left a sub-agent's own bead updates unsynced for no benefit anyone stated. The bodies now forbid committing, pushing and nesting, and say syncing is ordinary bookkeeping; a coordinator that wants a single writer on the bead data still reserves it in the brief, which the delegation shortcut now gives as the only reason to withhold it. All eight definitions and their eight goldens match the generator.

**Validation.** `setup --auto` reproduces all eight definitions byte for byte; the tier tests are 25 passed; the `--list` size golden was refreshed for the guideline edit; `ci:quality` exits 0. CI results at this head follow on the PR.

jlevy:
**Dispositions for review D** at `85e2d023`

Verified at this head by reading the corrected text, not by commit message. Post-rebase SHAs.

- D1: fixed in `ae26c00a`: the comparison separates inline work, a fork, and a fresh sub-agent, gives the fork its own end state (only the final report enters the coordinator's context) and the sub-agent TTL for its new writes, and the worked example was corrected to match. Review F then found the fork's cost figure still flattered by a flat context assumption; that is fixed in `85e2d023` as F5.
- D2: fixed in `ae26c00a`: the claim that removing the static model name improved within-definition cache reuse is gone, replaced by the accuracy reason, and the text says the two strong definitions cannot share a full prefix because their bodies differ. The premise holds at this head: the two bodies still differ, which the generator drift test pins.
- D3: fixed in `ae26c00a`: the polling comparison counts the extra reads and states the latency tradeoff. Recomputed independently at this head: $0.1875 + 15 × $0.015 = $0.4125 against $0.30 + 6 × $0.015 = $0.3900.

No test reads the research document, so all three corrections are unguarded against regression. That is a real gap, not a claim of coverage.

jlevy:
**Dispositions for review E** at `85e2d023`

- E1: fixed in `82d1ba66`: the guideline and the research separate the two facts. A definition is needed to set a reasoning level different from the session's, and independently it supplies a model default and the body; the blanket claim that a definition changes the model only when the level differs is gone, and the explicit-model comparison that follows is preserved.

The only automated guard is the `--list` size golden, which notices a revert through the byte count rather than the wording. Review F's F3 found a related claim in the same paragraph that was true on Claude Code but not on Codex; that is fixed in `85e2d023`.

jlevy:
**F6 follow-up.** The wording items deferred as `tbd-w1km` in the review F dispositions are now fixed in `ef794435`, so no finding from review F remains deferred:

- The Claude Code v2.1.242 floor moved to the two TTL settings it governs, away from the 5-minute default, which every build has.
- The definition body joined the brief and the shortcut as a carrier of the rules a sub-agent would otherwise never see, with a note not to rely on a platform's own tool descriptions for them.
- "Its warmed cache" is now bounded by the TTL: after a longer gap the context survives but the prefix is rewritten, so continuing costs about what a fresh start costs.
- "Whenever the brief does not restate those rules" no longer refers forward to rules that appear four lines later.
- The explanation of why tiers do not share a cache now says their definition bodies differ too, not only their model and effort.

This commit was rebased onto `38519193`, the .gitignore hardening that keeps a `node_modules` symlink from being committed again. That symlink was my error in `85e2d023`, and it is what broke this PR's CI at `pnpm install` until `493e418f` removed it.

`ci:quality` exits 0 and the documentation size golden passes at this head.

